PR #2, opened 2020-12-18
Update links and CONTRIBUTING

hagenw:
Updates links and CONTRIBUTING to the new links of the package and documentation on Github.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -77,6 +77,9 @@ html_theme_options = {
     'footer_links': False,
     'wide_pages': ['data-example'],
 }
+html_context = {
+    'display_github': True,
+}
 html_title = title
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,6 +74,7 @@ html_theme = 'sphinx_audeering_theme'
 html_theme_options = {
     'display_version': True,
     'logo_only': False,
+    'footer_links': False,
     'wide_pages': ['data-example'],
 }
 html_title = title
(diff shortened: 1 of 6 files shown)


frankenjoe:
In the documentation we sometimes set the `source` of a `Database` to `https://gitlab.audeering.com/...'. I think we should change that.

docs/conf.py
```diff
@@ -71,10 +68,6 @@
     'python': ('https://docs.python.org/3/', None),
 }
 
-# Disable Gitlab as we need to sign in
-linkcheck_ignore = [
-    'https://gitlab.audeering.com',
-]
 
 # HTML --------------------------------------------------------------------
```
frankenjoe, line 72: `'footer_links': False,` is missing in `html_theme_options`

frankenjoe: and we should set:

```
html_context = {
    'display_github': True,
}
```

hagenw:
I updated one Gitlab link in the docs.

We have another one:

```python
    # Create Database
    db = audformat.Database(
        name='databasename',
        source='https://gitlab.audeering.com/data/databasename',
        usage=audformat.define.Usage.COMMERCIAL,
    )
    db
```
But I think there it would make sense to stick with it, as it is still the link that we should use internally for databases.

docs/conf.py
hagenw: Good point, added it.

hagenw: OK, added `html_context` as well

Issue #12, opened 2021-01-06
Input type for audformat.segmented_index()


hagenw:
In the documentation the following input types are allowed for `starts` and `ends`:

![image](https://user-images.githubusercontent.com/173624/103761861-a523f880-5017-11eb-9124-679322e06840.png)

But it also works for integers:

```python
>>> audformat.segmented_index(['a.wav', 'a.wav'], [0, 1], [1, 2])
MultiIndex([('a.wav',           '0 days 00:00:00', ...),
            ('a.wav', '0 days 00:00:00.000000001', ...)],
           names=['file', 'start', 'end'])
```

and floats as well (besides that the position after the comma is no longer considered):

```python
>>> audformat.segmented_index(['a.wav', 'a.wav'], [0., 1.1], [1., 2.1])
MultiIndex([('a.wav',           '0 days 00:00:00', ...),
            ('a.wav', '0 days 00:00:00.000000001', ...)],
           names=['file', 'start', 'end'])
```

I would propose that we add `int` as type as well and add a word to the documentation that this is then handled as `ns`.

hagenw:
Another more user friendly solution might be to add integer and float as supported inputs and tread them as time in seconds and convert to timedelta under the hood.

frankenjoe:
I would not include `float` as a supported type.

frankenjoe:
Ah, overlooked your last comment. Yes, if we treat `float` as seconds it makes sense.

PR #13, opened 2021-01-06
DOC: add examples for index related functions

hagenw:
This adds the following examples to the docstring of index related functions:

### filewise_index

image

### segmented_index

image

### index_type

image

### DIFF
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -66,6 +66,12 @@ def index_type(
         ValueError: if not conform to
             :ref:`table specifications <data-tables:Tables>`
 
+    Example:
+        >>> index_type(filewise_index())
+        'filewise'
+        >>> index_type(segmented_index())
+        'segmented'
+
     """
     if isinstance(obj, (pd.Series, pd.DataFrame)):
         obj = obj.index
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -39,6 +39,10 @@ def filewise_index(
     Returns:
         filewise index
 
+    Example:
+        >>> filewise_index(['a.wav', 'b.wav'])
+        Index(['a.wav', 'b.wav'], dtype='object', name='file')
+
     """
     if files is None:
         files = []
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -113,6 +113,27 @@ def segmented_index(
     Raises:
         ValueError: if ``files``, ``start`` and ``ends`` differ in size
 
+    Example:
+        >>> segmented_index('a.wav', '0s', '1s')
+        MultiIndex([('a.wav', '0 days', '0 days 00:00:01')],
+                   names=['file', 'start', 'end'])
+        >>> segmented_index(['a.wav', 'b.wav'])
+        MultiIndex([('a.wav', '0 days', NaT),
+                    ('b.wav', '0 days', NaT)],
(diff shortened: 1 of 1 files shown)


audformat/core/index.py
```diff
@@ -103,6 +113,23 @@ def segmented_index(
     Raises:
         ValueError: if ``files``, ``start`` and ``ends`` differ in size
 
+    Example:
+        >>> segmented_index(['a.wav', 'b.wav'])
+        MultiIndex([('a.wav', '0 days', NaT),
+                    ('b.wav', '0 days', NaT)],
+                   names=['file', 'start', 'end'])
+        >>> segmented_index(['a.wav'], [0], [1])
```
frankenjoe, line 121:
```suggestion
        >>> segmented_index('a.wav', '0s', '1s')
```

audformat/core/index.py
```diff
@@ -103,6 +113,23 @@ def segmented_index(
...
+        MultiIndex([('a.wav', '0 days', '0 days 00:00:00.000000001')],
+                   names=['file', 'start', 'end'])
+        >>> segmented_index(
+        ...     files=['a.wav', 'a.wav'],
+        ...     starts=[t * 10 ** 9 for t in (0, 1)],
```
frankenjoe, line 126: I find this syntax a little bit too complicated for a usage example.

audformat/core/index.py
```diff
@@ -103,6 +113,23 @@ def segmented_index(
...
+        ...     ends=[pd.Timedelta(t, unit='s') for t in (1, 2)],
```
frankenjoe, line 127:
```suggestion
        ...     ends=pd.to_timedelta([1, 2], unit='s'),
```

frankenjoe:
We should may also also include an example for `segmented_index` where `starts` and `ends' is not set, and maybe one where only one of it is set.

audformat/core/index.py
hagenw: Would the following be better:
```python
starts=[f'{t}s' for t in (0, 1)],
```
For me the above line is still easier then using `pd.Timedelta` as I can never remember how to use that one.
The idea behind this line is also to make it easier when https://github.com/audeering/audformat/issues/12 is addressed.

frankenjoe: What about just `['0s', '1s']`?

frankenjoe: or `['0s', '1000ms']` to show that other units are supported, too?

hagenw: Yes `['0s', '1s']` it's fine for an example, but in reality you most probably have already a long list with integers or floats and not strings.

frankenjoe: Ok, let's use `[f'{t}s' for t in (0, 1)]` here and for `ends` use `pd.to_timedelta([1, 2], unit='s')`. Then we show two scalable options.

audformat/core/index.py
hagenw: I updated this.

audformat/core/index.py
hagenw: Changed it accordingly.

hagenw:
> We should may also also include an example for `segmented_index` where `starts` and `ends` are not set, and maybe one where only one of it is set.

The first was already there, for the other two I added one additional example.

PR #14, opened 2021-01-07
Parallel processing for Database methods drop_, map_, pick_files() and save()

### LINKED ISSUE #11: Progress bar in pick/drop_files() and map_files()
On a large `Database` object, using the methods `pick_files()`, `drop_files()` and `map_files()` may take quite some time to complete. We should consider adding a `verbose` argument to display a progress bar.

hagenw:
Closes #11.

This uses `audeer.run_tasks()` for `Database.drop_files()`, `Database.map_files()`, `Database.pick_files()`.

Haven't tested yet if you really see a progress bar, but I don't see a reason why it shouldn't work.

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -320,6 +320,8 @@ class Database(HeaderBase):
                 define.TableStorageFormat.CSV
             ),
             header_only: bool = False,
+            num_workers: typing.Optional[int] = 1,
+            verbose: bool = False,
     ):
         r"""Save database to disk.
 
@@ -334,6 +336,10 @@ class Database(HeaderBase):
                 See :class:`audformat.define.TableStorageFormat`
                 for available formats
             header_only: store header only
+            num_workers: number of parallel jobs.
+                If ``None`` will be set to the number of processors
+                on the machine multiplied by 5
+            verbose: show progress bar
 
         """
         root = audeer.mkdir(root)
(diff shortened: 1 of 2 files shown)


frankenjoe:
Great thank you, will test it.

I think we should also use `run_tasks()` in `Database.save()`.

audformat/core/database.py
```diff
@@ -169,7 +169,9 @@ def drop_files(
                 str,
                 typing.Sequence[str],
                 typing.Callable[[str], bool],
-            ]
+            ],
+            num_workers: int = 1,
```
frankenjoe, line 173:
```suggestion
            num_workers: typing.Optional[int] = 1,
```

audformat/core/database.py
```diff
@@ -204,7 +215,9 @@ def pick_files(
...
```
frankenjoe, line 219:
```suggestion
            num_workers: typing.Optional[int] = 1,
```

audformat/core/database.py
```diff
@@ -239,6 +261,8 @@ def pick_tables(
     def map_files(
             self,
             func: typing.Callable[[str], str],
+            num_workers: int = 1,
```
frankenjoe, line 264:
```suggestion
            num_workers: typing.Optional[int] = 1,
```

frankenjoe:
I suggest to add tests with `num_workers` > 1 for the according tests in `test_database`.

frankenjoe:
> Haven't tested yet if you really see a progress bar, but I don't see a reason why it shouldn't work.

Can confirm it works.

hagenw:
> I suggest to add tests with `num_workers` > 1 for the according tests in `test_database`.

I added those tests.

hagenw:
> Great thank you, will test it.
> 
> I think we should also use `run_tasks()` in `Database.save()` when saving the tables. This may also be a lengthy operation for large databases.

I added it together with a test in another commit, so please choose "Rebase and merge" in the end.

PR #16, opened 2021-01-07
Allow for int, str, float in segmented_index()

### LINKED ISSUE #12: Input type for audformat.segmented_index()

hagenw:
Closes #12.

This adds `float`, `int`, `str` to the allowed types in `audformat.segmented_index()`:

image

Looks a little bit messy, but we are quite flexible with seems to me a good thing for this function.

Both are now interpreted as seconds which you expect as the default behavior anyway.

### DIFF
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -26,6 +26,14 @@ def to_array(value: typing.Any) -> typing.Union[list, np.ndarray]:
     return value
 
 
+def to_timedelta(times):
+    r"""Convert time value to pd.TimeDelta."""
+    try:
+        return pd.to_timedelta(times, unit='s')
+    except ValueError:  # catches values like '1s'
+        return pd.to_timedelta(times)
+
+
 def filewise_index(
         files: Files = None,
 ) -> pd.Index:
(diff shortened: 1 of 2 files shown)


frankenjoe:
For completness we should add `str` as another type.

hagenw:
> For completness we should add `str` as another type.

Added it and updated the description.

frankenjoe:
I think we should leave a note in the docstring that plain `int` and `float` values are interpreted as seconds.

audformat/core/index.py
```diff
@@ -156,8 +159,14 @@ def segmented_index(
             "'starts', and 'ends' differ in size",
         )
 
+    def convert_to_timedelta(times):
```
frankenjoe, line 162: I would move this outside of the function, as we do for the `to_array()` helper.

hagenw:
> I think we should leave a note in the docstring that plain `int` and `float` values are interpreted as seconds.

I added it to the docstring:

audformat/core/index.py
hagenw: Renamed it to `to_timedelta()` and made it a helper function.

audformat/core/index.py
```diff
@@ -26,6 +26,14 @@ def to_array(value: typing.Any) -> typing.Union[list, np.ndarray]:
     return value
 
 
+def to_timedelta(times):
+    r"""Convert time value to pd.TimeDelta."""
```
frankenjoe, line 30:
```suggestion
    r"""Convert time value to pd.Timedelta."""
```

Issue #17, opened 2021-01-07
to_segmented_index() returns wrong index type for end


hagenw:
Compare the output of:

```python
>>> idx = audformat.segmented_index('a')
>>> idx
MultiIndex([('a', '0 days', NaT)],
           names=['file', 'start', 'end'])
>>> idx.get_level_values('end')
TimedeltaIndex([NaT], dtype='timedelta64[ns]', name='end', freq=None)
```
with
```python
>>> idx = audformat.utils.to_segmented_index(audformat.filewise_index('a'))
>>> idx
MultiIndex([('a', '0 days', 'NaT')],
           names=['file', 'start', 'end'])
>>> idx.get_level_values('end')
DatetimeIndex(['NaT'], dtype='datetime64[ns]', name='end', freq=None)
```

This is very unfortunate as it makes it much harder to work with the indices in other applications, e.g. for example how to easily calculate a duration if the type of end can be different.

frankenjoe:
I know this is nasty. What I usually do to avoid that 'NaT'  is interpreted as `datatime` is `pd.to_timedelta(pd.NaT)`.

hagenw:
> I know this is nasty. What I usually do to avoid that `NaT` is interpreted as `datetime` is `pd.to_timedelta(pd.NaT)`.

I tried this at https://github.com/audeering/audformat/blob/master/audformat/core/utils.py#L348 but it did not help, so I created the solution presented in #18

frankenjoe:
OK! Good to know this fix does not always help.

PR #18, opened 2021-01-07
Fix index type for to_segmented_index()

hagenw:
Closes https://github.com/audeering/audformat/issues/17

I have to admit the whole pandas index universe is kind of a mess and you need to be very careful to not mess it up ;)

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -343,17 +343,10 @@ def to_segmented_index(
     else:
         index = obj
 
-    df = index.to_frame(index=False)
-    df[define.IndexField.START] = pd.to_timedelta(0)
-    df[define.IndexField.END] = pd.NaT
-
-    index = pd.MultiIndex.from_frame(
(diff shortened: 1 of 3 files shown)


audformat/core/utils.py
```diff
@@ -343,17 +343,10 @@ def to_segmented_index(
...
-        df,
-        names=[
-            define.IndexField.FILE,
-            define.IndexField.START,
-            define.IndexField.END,
-        ]
+    index = segmented_index(
+        files=list(index),
+        starts=[0] * len(index),
+        ends=[pd.NaT] * len(index),
```
hagenw, line 349: I think it is in general better to use our own `segmented_index` function here so we don't have to update this function if we update `segmented_index()`. Maybe it is not as efficient as the previous solution, but hopefully this will not a big deal.

frankenjoe: Agree

PR #22, opened 2021-01-14
Fix bool scheme

### LINKED ISSUE #21: Error saving NaN for a boolean scheme
This is a minimal example with one table that has a column linked to boolean scheme:

```python
import audformat


db.schemes['scheme'] = audformat.Scheme(audformat.define.DataType.BOOL)
db['t'] = audformat.Table(audformat.filewise_index(['f1']))
db['t']['c'] = audformat.Column(scheme_id='scheme')
```

By default, the labels in the column are initialized with `NaN` and we get the following expected output:

```python
db['t'].get()
```
```
        c
file     
f1    NaN
```

However, saving and loading the database as CSV fails (using pickle works):

```python
db.save('db')
audformat.Database.load('db')
```
```
ValueError: Bool column has NA values in column 1
```

There are two solutions:

1. we do not support `NaN` for boolean schemes and initalize with `False`
2. we try to find a fix for loading empty values to a boolean scheme from CSV

frankenjoe:
Closes #21 

* Fix #21 by using Pandas' [boolean](https://pandas.pydata.org/pandas-docs/stable/user_guide/boolean.html), which is nullable
* Fix `Scheme.draw()` to draw `bool` values for a boolean scheme
* Extend default test database by a column holding `bool` values

### DIFF
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -202,12 +202,14 @@ class Scheme(HeaderBase):
             return pd.api.types.CategoricalDtype(
                 categories=labels, ordered=False,
             )
+        elif self.dtype == define.DataType.BOOL:
+            return 'boolean'
+        elif self.dtype == define.DataType.DATE:
+            return 'datetime64[ns]'
         elif self.dtype == define.DataType.INTEGER:
             return 'Int64'
         elif self.dtype == define.DataType.TIME:
             return 'timedelta64[ns]'
-        elif self.dtype == define.DataType.DATE:
-            return 'datetime64[ns]'
         return self.dtype
 
     def __contains__(self, item: typing.Any) -> bool:
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -144,7 +144,12 @@ class Scheme(HeaderBase):
         """
         x = None
         if self.labels is None:
-            if self.dtype == define.DataType.INTEGER:
+            if self.dtype == define.DataType.BOOL:
+                x = [random.choice([False, True]) for _ in range(n)]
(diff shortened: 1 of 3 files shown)


audformat/core/scheme.py
```diff
@@ -200,12 +202,14 @@ def to_pandas_dtype(self) -> typing.Union[
             return pd.api.types.CategoricalDtype(
                 categories=labels, ordered=False,
             )
+        elif self.dtype == define.DataType.BOOL:
+            return 'boolean'
```
frankenjoe, line 206: Here's the important bit where we set `dtype='boolean'`, which is nullable. Other changes here are just cosmetic.

audformat/core/testing.py
```diff
@@ -246,57 +246,66 @@ def create_db(minimal: bool = False) -> Database:
     # Schemes #
     ###########
 
-    db.schemes['string'] = Scheme()
-    db.schemes['int'] = Scheme(
-        dtype=define.DataType.INTEGER, minimum=0, maximum=100,
-    )
+    db.schemes['bool'] = Scheme(dtype=define.DataType.BOOL)
```
frankenjoe, line 249: Here we add a scheme with boolean values to cover this type. Other changes are just cosmetic.

audformat/core/scheme.py
```diff
@@ -144,7 +144,12 @@ def draw(
         """
         x = None
         if self.labels is None:
-            if self.dtype == define.DataType.INTEGER:
+            if self.dtype == define.DataType.BOOL:
```
frankenjoe, line 147: Here we add the missing case for `bool`. Other changes are cosmetic.

PR #28, opened 2021-01-27
Implement __eq__

### LINKED ISSUE #26: Implement __eq__ for Database
Try:

```python
>>> import audformat
>>> import copy
>>> db1 = audformat.testing.create_db(minimal=True)
>>> db2 = copy.deepcopy(db1)
>>> db1 == db2
False
```

But it should return `True`. This is not the case as we have not implement `__eq__` for database objects.

frankenjoe:
Closes #26 

### Example

```python
import audformat


db2 = audformat.testing.create_db()

assert db != db2

db.save('test')
db3 = audformat.Database.load('test')

assert db == db3
```

### Note

In Python 3 a class that overrides `__eq__` does not have a `__hash__` method by default:

https://docs.python.org/3/reference/datamodel.html#object.__hash__

The implementation I propose here, implements `__hash__` except for `Database`, `Table` and `Column` as their state usually depends on one or more `pd.DataFrame` objects, which are mutable (actually there are ways to get the hash for a `pd.DataFrame`, but I decided to stay with the default behavior of `pandas`, which does not implement `__hash__`). This means, we cannot use  `Database`, `Table` and `Column` in a `set()` or as key in a `dict`.

### DIFF
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -222,3 +222,13 @@ class Column(HeaderBase):
                     'Cannot set values of a filewise column '
                     'using a segmented index.'
                 )
+
+    def __eq__(
+            self,
+            other: 'Column',
+    ) -> bool:
+        if self.dump() != other.dump():
+            return False
+        if self._table is not None and other._table is not None:
+            return self.get().equals(other.get())
+        return self._table is None and other._table is None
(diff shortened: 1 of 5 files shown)


hagenw:
Great, now the following also works as expected:

```python
>>> import audformat.testing
>>> db1 = audformat.testing.create_db(minimal=True)
>>> db2 = audformat.testing.create_db(minimal=True)
>>> db1 == db2
True
>>> import copy
>>> db3 = copy.deepcopy(db1)
>>> db1 == db3
True
```

PR #29, opened 2021-01-28
Add update_other_formats option to Table.save()

hagenw:
Adds the possibility to store a table to a given format,
but update other existing formats as well.
This is also the default behavior, as we have `update_other_formats=True`.

`Database.save()`

image

`Table.save()`

image

### DIFF
--- a/audformat/core/common.py
+++ b/audformat/core/common.py
@@ -227,12 +227,21 @@ class DefineBase:
 
     @classmethod
     def assert_has_value(cls, value):
-        attributes = inspect.getmembers(
-            cls, lambda x: not inspect.isroutine(x))
-        valid_values = [a[1] for a in attributes if
-                        not(a[0].startswith('__') and a[0].endswith('__'))]
+        valid_values = cls.values()
         if value not in valid_values:
             raise BadValueError(value, valid_values)
+
+    @classmethod
+    def values(cls):
+        attributes = inspect.getmembers(
+            cls, lambda x: not inspect.isroutine(x)
+        )
+        return sorted(
+            [
+                a[1] for a in attributes
+                if not(a[0].startswith('__') and a[0].endswith('__'))
+            ]
+        )
 
 
 def format_series_as_html():  # pragma: no cover (only used in documentation)
--- a/audformat/core/common.py
+++ b/audformat/core/common.py
@@ -226,13 +226,13 @@ class HeaderBase:
 class DefineBase:
 
     @classmethod
-    def assert_has_value(cls, value):
-        valid_values = cls.values()
+    def assert_has_attribute_value(cls, value):
+        valid_values = cls.attribute_values()
         if value not in valid_values:
             raise BadValueError(value, valid_values)
 
     @classmethod
-    def values(cls):
+    def attribute_values(cls):
         attributes = inspect.getmembers(
             cls, lambda x: not inspect.isroutine(x)
         )
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -655,9 +655,10 @@ class Table(HeaderBase):
             self,
             path: str,
             *,
-            storage_format: typing.Union[str, define.TableStorageFormat] = (
+            storage_format: define.TableStorageFormat = (
                 define.TableStorageFormat.CSV
             ),
+            update_other_formats: bool = True,
     ):
         r"""Save table data to disk.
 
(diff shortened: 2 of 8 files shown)


audformat/core/define.py
```diff
@@ -69,6 +69,7 @@ class TableStorageFormat(DefineBase):
     r"""Storage format of tables."""
     CSV = 'csv'
     PICKLE = 'pkl'
+    ALL = 'all'
```
hagenw, line 72: This is not very elegant.
Another solution would be to not add it here, but then we have to allow something like `typing.Union[str, TableStorageFormat]` as input for the `storage_format` argument.

frankenjoe: We could also support `storage_format=None` and in that case store in all formats?

hagenw: Yes, this owuld be another solution. The downside is that you would expect `storage_format=None` to be the default then, wouldn't you?

frankenjoe: Mhh, but it's also a case that will be only used in special occasion. Or you think this will be often used?

frankenjoe: Or maybe there is yet another solution: we could add a flag that - when turned on - checks if there is already a table in another format and in that case also stores the table in that format.

frankenjoe: Something like `update_other_formats=True`.

hagenw: I guess you are right. I changed it now to support `None` and removed `all`.

frankenjoe: What about the other solution? Maybe we should add it as well and set it to `True` by default. Then you will not end up with different versions of the same table even if you forget to set `storage_format=None`.

frankenjoe: + it will only write out the other format if really needed

hagenw: Good point, so I reverted again.

We have now:
* 'all', stores to all available formats
* 'update' updates only existing database files. This means if a database contains only PKL files for some tables, also only those are updated. Or if you delete a CSV file manually it will not written again.

hagenw: Please squash when merging

audformat/core/common.py
```diff
@@ -227,13 +227,22 @@ class DefineBase:
...
-        attributes = inspect.getmembers(
-            cls, lambda x: not inspect.isroutine(x))
-        valid_values = [a[1] for a in attributes if
-                        not(a[0].startswith('__') and a[0].endswith('__'))]
+        valid_values = cls.values()
         if value not in valid_values:
             raise BadValueError(value, valid_values)
 
+    @classmethod
+    def values(cls):
```
frankenjoe, line 235: Maybe we should choose a more descriptive name like `class_attributes`?

audformat/core/common.py
```diff
@@ -227,13 +227,22 @@ class DefineBase:
 
     @classmethod
     def assert_has_value(cls, value):
```
frankenjoe, line 229: This one we could rename to `assert_class_has_attribute`?

audformat/core/common.py
hagenw: But would `class_attributes` not refer to the name of the attribute, or would it refer to its content?

audformat/core/define.py
frankenjoe: You implemented it in a slightly different way what I had in mind. 

As far as I see, with the current implementation `Table.save(path, 'csv')` will not update if a pickled version exists (and vice-versa). Which is exactly the case we'd like to avoid. On the other hand, `Table.save(path, 'update')` will not save the table if it does not exist, which might also be dangerous.

What I had in mind is adding `update` as an additional boolean flag. If it is set to `True` it will - in addition to the format that is selected - check if other formats of the table exist already save those as well. E.g.: `Table.save(path, 'csv', update=True)` will always save the table as CSV, but if there is already a pickled table, it will also save as PKL. I think we can then even remove `'all'` again, cause as long as the flag is turned on, we can never have the case that we have a CSV and PKL file that do not match.

hagenw: If I read something like `Table.save(path, 'csv', update=True)` I would assume it would look for a cSV file, and updates its content if needed. But not that it would touch files with a different format.

I think our current solution with several output format is still not optimal, but I also don't have a better idea at the moment.

hagenw: One solution would be that we always write all formats.

audformat/core/common.py
frankenjoe: Good point. Maybe `class_attribute_values`  then?

audformat/core/define.py
frankenjoe: Then let's call it `update_other_formats` as I proposed earlier. I really think that this ensures the behavior we want: save to a specific format, but ensure to update other formats if they already exists.

frankenjoe: I neither see the use-case for saving to all formats, nor to only update formats that exist (which in the worst case means not saving the table at all and the user is not aware of it).

hagenw: OK, `update_other_formats` is fine with me, I will update accordingly.

audformat/core/common.py
hagenw: I renamed it to `attribute_values`

audformat/core/common.py
hagenw: I renamed it to `assert_has_attribute_value`

audformat/core/define.py
hagenw: I added `update_other_formats=True` and removed `'all'` and `'update'`.

I also added test that create PKL files that are different from CSV and updates them afterwards.
The test is still failing and I don't know why at the moment.

hagenw: I updated the description accordingly.

audformat/core/table.py
```diff
@@ -658,26 +658,37 @@ def save(
...
             ),
+            update_other_formats: bool = True,
     ):
         r"""Save table data to disk.
 
+        Existing files will be overwritten.
+
         Args:
             path: file path without extension
             storage_format: storage format of tables.
```
frankenjoe, line 669:
```suggestion
            storage_format: storage format of table.
```

tests/test_database.py
```diff
@@ -136,11 +136,60 @@ def test_map_files(num_workers):
...
+            else:
+                assert not os.path.exists(table_file)
+
+    # Test update other formats
+    if (
+            storage_format == audformat.define.TableStorageFormat.CSV
+            and db.tables
+    ):
+        db2 = audformat.testing.create_db()
+        db2.meta = {}
```
frankenjoe, line 160: Why do we need this?

hagenw: I removed them, we don't need them.

They are leftovers from me trying to get the test working for `create_db(minimal=True)` as well. But in the end I decided to just skip the test then as we don't have tables there anyway.

Issue #27, opened 2021-01-27
Where to store license of a database?


hagenw:
We don't have any `db.license` entry at the moment. Before it was handled inside the Gradle settings, but I think we should add an option to store it also in the database.
Of course, it can be done already by using the `meta` field or adding it to the description, but I think we should add an extra entry for it.

Some licenses might come with an URL, so maybe we should add such an option as well.

frankenjoe:
Actually, in an early version of `audata` we had a license field, but then decided to remove it and handle it with `Gradle` :) But yes, I would say it's an important piece of information that should be part of the header.

What about the list of maintainers?

hagenw:
> What about the list of maintainers?

Good point, I think it makes sense to put this into the database as well.

hagenw:
>> What about the list of maintainers?
>
> Good point, I think it makes sense to put this into the database as well.

I think I changed my mind and would not include `maintainers` in the database, but we might want to add `author` instead?

frankenjoe:
Fine for me, too.

frankenjoe:
Closing as we now have `license` (see https://github.com/audeering/audformat/pull/34) and `author` (see https://github.com/audeering/audformat/pull/40) fields.

PR #34, opened 2021-02-04
Add license and license_url to Database

hagenw:
This covers the license part discussed in #27.

It add the following two arguments/attributes to `audformat.Database():
* `license`
* `license_url`

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -36,6 +36,8 @@ class Database(HeaderBase):
         expires: expiry date
         languages: list of languages
         description: database description
+        license: database license
+        license_url: URL of database license
         meta: additional meta fields
 
     Raises:
@@ -101,6 +103,8 @@ class Database(HeaderBase):
             expires: datetime.date = None,
             languages: typing.Union[str, typing.Sequence[str]] = None,
             description: str = None,
+            license: str = None,
+            license_url: str = None,
             meta: dict = None,
     ):
         define.Usage.assert_has_attribute_value(usage)
@@ -120,6 +124,10 @@ class Database(HeaderBase):
         r"""Expiry date"""
         self.languages = languages
         r"""List of included languages"""
+        self.license = license
+        r"""License of database"""
+        self.license_url = license_url
+        r"""URL of database license"""
         self.media = HeaderDict(value_type=Media)
         r"""Dictionary of media information"""
         self.raters = HeaderDict(value_type=Rater)
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -12,6 +12,7 @@ from audformat.core import utils
 from audformat.core.index import segmented_index
 from audformat.core.column import Column
 from audformat.core.common import HeaderBase, HeaderDict
+from audformat.core.define import LICENSE_URLS
 from audformat.core.errors import BadIdError
 from audformat.core.media import Media
 from audformat.core.rater import Rater
@@ -36,7 +37,12 @@ class Database(HeaderBase):
         expires: expiry date
         languages: list of languages
         description: database description
-        license: database license
+        license: database license.
+            You can use a custom license
+            or pick one from :attr:`audformat.define.License`.
+            In the later case,
+            ``license_url`` will be automatically set
+            if it is not given
         license_url: URL of database license
         meta: additional meta fields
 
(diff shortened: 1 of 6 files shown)


frankenjoe:
Thanks for adding this!

Would it make sense to provide an enumeration with standard licences as we do for usage?

hagenw:
The problem with enumeration is that we would then have to use other for all licenses that are not provided by us, but I would prefer that the actual license name is inserted.

On the other hand I agree that it would be nice to avoid different names for the same license `CC0` vs. `CC-0`.

frankenjoe:
There seems to be a package that could be useful: https://pypi.org/project/license/ but it's in a very early development stage.

frankenjoe:
Maybe we can extract them from here: https://spdx.org/licenses/

frankenjoe:
There is also a Python package https://pypi.org/project/spdx/ for it

hagenw:
> Maybe we can extract them from here: https://spdx.org/licenses/

They don't contain all the licenses we want to use, so I would not directly use them.

One simple solution might be to provide a list of licenses in `audformat.define.LICENSES`, but don't force the user to pick one from there. If she is pickes from `audformat.define.LICENSES` we can then automatically fill in the URL.
For a custom license should we maybe throw a warning if the user does not provide an URL then?

frankenjoe:
> One simple solution might be to provide a list of licenses in audformat.define.LICENSES

Yes, let's do that. It also hast the advantage that we don't introduce a new dependency.

> For a custom license should we maybe throw a warning if the user does not provide an URL then?

This probably means we raise a warning for all our internal databases :)

hagenw:
OK, I added `audformat.define.License`:

image

And you can now pick from those, or define your own:

image

audformat/core/database.py
```diff
@@ -101,9 +109,16 @@ def __init__(
             expires: datetime.date = None,
             languages: typing.Union[str, typing.Sequence[str]] = None,
             description: str = None,
+            license: typing.Union[str, define.License] = None,
```
hagenw, line 112: `define.License` as a type cannot be correct as it is no enumeration. But we use the same with `usage` for example.

audformat/core/database.py
```diff
@@ -12,6 +12,7 @@
 from audformat.core.index import segmented_index
 from audformat.core.column import Column
 from audformat.core.common import HeaderBase, HeaderDict
+from audformat.core.define import LICENSE_URLS
```
frankenjoe, line 15: We already do `from audformat.core import define`, so maybe we can remove and replace ` LICENSE_URLS` with `define.LICENSE_URLS` below?

audformat/core/database.py
frankenjoe: I'm fine with it, unless you would like to change to `enum.Enum` as we do in `opensmile.FeatureExtractor`. But there was some reason I didn't like to work `enum.Enum`, though I cannot remember what it was at the moment. The other problem might be that we also derive from `DefineBase` here, so not sure if it is straightforward to change to `enum.Enum`.

audformat/core/database.py
hagenw: You are right, I changed it accordingly.

audformat/core/testing.py
```diff
@@ -217,6 +217,10 @@ def create_db(minimal: bool = False) -> Database:
         return db
 
     db.description = 'A database for unit testing.'
+    db.license = 'CC0'
```
frankenjoe, line 220: Is there a specific reason you don't want to use one of our pre-defined licenses and let it fill in `license_url` automatically?

audformat/core/database.py
hagenw: No, I don't want to change to `enum.Enum`, but I think we can only write `license: str = None` and `usage: str = define.Usage.UNRESTRICTED` and have to refer to `define.License` and `define.Usage` in the docstrings.

hagenw: Because `type(audformat.define.License.CC0_1_0) != audformat.define.License`

audformat/core/testing.py
hagenw: No, I just forgot to change this.
Fixed it now.

audformat/core/database.py
frankenjoe: Ok, maybe that makes more sense. But we actually have it in quite a lot of places, so I suggest to do it in a separate MR and leave it as it is for the moment.

hagenw: OK, I opened https://github.com/audeering/audformat/issues/39

PR #35, opened 2021-02-04
DOC: cache emodb on Github

### LINKED ISSUE #6: Enable caching of emob when building documentation
At the moment the 40MB source file of emodb has to be donwloaded for every documentation test.
There is no need for this and we should enable [caching](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows) of that file.

hagenw:
Closes #6.

Caches the extracted emodb database to avoid downloading it every time the documentation is build.

It is stored under the cache key `emodb-ubuntu` and is used in the test and publish workflows.
We don't need to cache for Windows or Mac as the docuemntation is only build on `ubuntu-18.04` in the test workflow and `ubuntu-latest` in the publish workflow.

Note: it will [only cache for 7 days](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy), but that's still better as nothing as we will trigger the pipeline with every commit we add to an existing pull request.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,6 +13,12 @@ jobs:
       with:
         fetch-depth: 2
 
+    - name: Cache emodb
+      uses: actions/cache@v2
+      with:
+        path: emodb-src
+        key: emodb-ubuntu
+
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,14 +12,22 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 2
+
+    - name: Cache emodb
+      uses: actions/cache@v2
+      with: emodb-src
+      key: emodb
+
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
         python-version: '3.x'
+
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


hagenw:
In the first run of the CI piepline, the cache seems to be stored succesfully:

image

image

I will now try to trigger a second run to see if the cache is actually used.

hagenw:
this dos not look to promising:

image

It seems like it does not store the correct folder inside the cache.

hagenw:
I think I know the problem:

```yml
    - name: Test building documentation
      run: |
        python -m sphinx docs/ docs/_build/ -b html -W
        python -m sphinx docs/ build/sphinx/html -b linkcheck
      if: matrix.os == 'ubuntu-18.04'
```
This means we only download `emodb` for  'ubuntu-18.04' and should also only cache there. I will try.

hagenw:
Great, now I think it has worked:

image

I will do one final cleanup and integrate it into publishing as well.

hagenw:
OK, all done now.

frankenjoe:
Will it automatically update the cache if we change the database version?

hagenw:
This does not cache a versioned database, but the original download from the emodb web page.

It will not automatically update the cache if the file on the original web page changes, but I would also not expect that.

frankenjoe:
Yes, that's not very likely.

Looks good, I will merge.

PR #36, opened 2021-02-04
Database.load() raises error if table file missing

hagenw:
`audformat.Database.load()` now raises an error if no table files can be found.

PR #37, opened 2021-02-04
Database.load() checks which file to load

hagenw:
Checks if it should load from PKL or CSV file.
If PKL file exists it is preferred unless a newer CSV file exists.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -541,15 +541,18 @@ class Table(HeaderBase):
         r"""Load table data from disk.
 
         Tables can be stored as PKL and/or CSV files to disk.
-        It will load from the PKL file
-        if available and its modification date is newer,
-        otherwise the CSV file is loaded.
+        If both files are present
+        it will load the PKL file
+        as long as its modification date is newer,
+        otherwise it will raise an error
+        and ask to delete one of the files.
 
         Args:
             path: file path without extension
 
         Raises:
             RuntimeError: if table file(s) are missing
+            RuntimeError: if CSV file is newer than PKL file
 
         """
         path = audeer.safe_path(path)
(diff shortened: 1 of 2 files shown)


hagenw:
#36 should be merged first, before we continue here.

hagenw:
I updated and fixed the tests.

audformat/core/table.py
```diff
@@ -540,6 +540,11 @@ def load(
     ):
         r"""Load table data from disk.
 
+        Tables can be stored as PKL and/or CSV files to disk.
+        It will load from the PKL file
+        if available and its modification date is newer,
+        otherwise the CSV file is loaded.
```
frankenjoe, line 546: Should we raise a warning if the modification date of the CSV file is newer than that of the PKL file?

frankenjoe: Most likely this means the two are not in sync, right?

hagenw: yes, but I would not raise a warning then, because it might be totally on purpose, e.g. you modified the CSV file.

frankenjoe: yes, but if I do that I should either update the PKL or delete it

frankenjoe: or you think with this approach we are safe enough that it can never fire back if CSV and PKL are not in sync?

hagenw: This was the whole idea of this approach, to make it more safe.

Another solution would be to raise an error an ask the user to first delete the PKL file.

frankenjoe: Yes, that would be the safest solution in my opinion.

hagenw: I changed it. It now raises an error if the CSV file is newer.

audformat/core/table.py
```diff
@@ -556,7 +564,27 @@ def load(
...
+                    and os.path.getmtime(csv_file) > os.path.getmtime(pkl_file)
+            ):
+                print('CSV time:', os.path.getmtime(csv_file))
+                print('PKL time:', os.path.getmtime(pkl_file))
+                raise RuntimeError(
+                    f"The table CSV file '{csv_file}' is newer "
+                    f"than the table PKL file '{pkl_file}'. "
+                    "If you want to load from the CSV file, "
+                    "please delete the PKL file. "
+                    "If you wanto to load from the PKL file, "
```
frankenjoe, line 583:
```suggestion
                    "If you want to load from the PKL file, "
```

PR #40, opened 2021-02-15
Add audformat.Database.author

hagenw:
This adds the `author` attribute to `audformat.Database`.

I'm not completely sure if it is always possible to differentiate it from `source`, but I think in most cases it should be obvious to the user.

### DIFF
--- a/audformat/core/testing.py
+++ b/audformat/core/testing.py
@@ -217,7 +217,7 @@ def create_db(minimal: bool = False) -> Database:
         return db
 
     db.description = 'A database for unit testing.'
-    db.author = 'audEERING GmbH'
+    db.author = 'J. Wagner, H. Wierstorf'
     db.license = define.License.CC0_1_0,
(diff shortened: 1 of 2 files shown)


audformat/core/testing.py
```diff
@@ -217,6 +217,7 @@ def create_db(minimal: bool = False) -> Database:
         return db
 
     db.description = 'A database for unit testing.'
+    db.author = 'audEERING GmbH'
```
hagenw, line 220: Should we use our names here instead?

frankenjoe: Yes, I think that makes more sense. If we want to have the option to store the company name as well, I suggest to introduce another field like `organization`.

hagenw: OK, I switched to real names and created https://github.com/audeering/audformat/pull/41 so we can decide there if we want to add organization as well.

PR #44, opened 2021-02-18
Union of index objects

frankenjoe:
Adds `utils.union()` to create the union of a list of index objects.

### Example:

```python
objs = [
    audformat.segmented_index(['f1', 'f2'], [0, 0], [2, 2]),
    audformat.segmented_index(['f1', 'f2', 'f3'], [1, 0, 0], [2, 2, 2]),
    audformat.filewise_index(['f1', 'f1', 'f2', 'f3']),
]
audformat.utils.union(objs)
```
```
MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:02'),
            ('f1', '0 days 00:00:00',               NaT),
            ('f1', '0 days 00:00:01', '0 days 00:00:02'),
            ('f2', '0 days 00:00:00', '0 days 00:00:02'),
            ('f2', '0 days 00:00:00',               NaT),
            ('f3', '0 days 00:00:00', '0 days 00:00:02'),
            ('f3', '0 days 00:00:00',               NaT)],
```

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -499,7 +499,6 @@ def union(
                     ('f4', '0 days 00:00:00', '0 days 00:00:01')],
                    names=['file', 'start', 'end'])
 
-
     """
     if not objs:
         return filewise_index()
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -517,6 +517,9 @@ def union(
         if index.empty:
             index = segmented_index()
         elif index.levels[2].empty:
+            # If all end values are NaT, pandas stores an empty array and
+            # since pd.Index.union() in that case sets the type to
+            # DatetimeArray, we need to set it back to 'timedelta64[ns]'.
             index.set_levels(
                 [pd.to_timedelta([])],
                 level=[2],
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -113,7 +113,7 @@ def intersect(
             :ref:`table specifications <data-tables:Tables>`
 
     Returns:
-        intersction of index objects
+        intersection of index objects
 
     Raises:
         ValueError: if one or more objects are not conform to
@@ -419,8 +419,10 @@ def to_segmented_index(
             :ref:`table specifications <data-tables:Tables>`
 
     """
-    if obj.empty or index_type(obj) == define.IndexType.SEGMENTED:
+    if index_type(obj) == define.IndexType.SEGMENTED:
         return obj
+    elif obj.empty:
+        return segmented_index()
 
     if isinstance(obj, (pd.Series, pd.DataFrame)):
         index = obj.index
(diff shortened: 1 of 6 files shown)


hagenw:
To be symmertical with `utils.intersect()` I would expect the following output of `audformat.utils.union(objs)`:

```
MultiIndex([('f1', '0 days 00:00:00',               NaT),
            ('f2', '0 days 00:00:00',               NaT),
            ('f3', '0 days 00:00:00',               NaT)],
```
but I agree that for most use cases your implementation makes a lot more sense.

frankenjoe:
Yes, I would prefer to stick to the proposed implementation.

audformat/core/utils.py
```diff
@@ -419,8 +419,10 @@ def to_segmented_index(
             :ref:`table specifications <data-tables:Tables>`
 
     """
-    if obj.empty or index_type(obj) == define.IndexType.SEGMENTED:
+    if index_type(obj) == define.IndexType.SEGMENTED:
         return obj
+    elif obj.empty:
+        return segmented_index()
```
hagenw, line 425: Why this change to an existing function?

audformat/core/utils.py
```diff
@@ -439,3 +441,87 @@ def to_segmented_index(
     obj = obj.reset_index(drop=True)
     obj.index = index
     return obj
+
+
+def union(
+    objs: typing.Sequence[typing.Union[pd.Index]],
```
hagenw, line 447: Why Union?

Maybe:
```suggestion
    objs: typing.Sequence[pd.Index],
```

audformat/core/utils.py
```diff
@@ -439,3 +441,87 @@ def to_segmented_index(
...
+                    ('f2', '0 days 00:00:00',               NaT),
+                    ('f2', '0 days 00:00:00', '0 days 00:00:01'),
+                    ('f3', '0 days 00:00:00',               NaT),
+                    ('f3', '0 days 00:00:00', '0 days 00:00:01'),
+                    ('f3', '0 days 00:00:01', '0 days 00:00:02'),
+                    ('f4', '0 days 00:00:00',               NaT),
+                    ('f4', '0 days 00:00:00', '0 days 00:00:01')],
+                   names=['file', 'start', 'end'])
+
+
```
hagenw, line 502:
```suggestion
```

audformat/core/utils.py
```diff
@@ -439,3 +441,87 @@ def to_segmented_index(
...
+    if index_type(index) == define.IndexType.SEGMENTED:
+        # asserts that start and end are of type 'timedelta64[ns]'
+        if index.empty:
+            index = segmented_index()
+        elif index.levels[2].empty:
+            index.set_levels(
+                [pd.to_timedelta([])],
+                level=[2],
+                inplace=True,
+            )
```
hagenw, line 525: When can it happen that `index.levels[2].empty`?

Does it also mean if I pass two index objects with missing second level the returned index will then contain a second level?

audformat/core/utils.py
frankenjoe: This fixes a bug we had - before when you passed an empty `filewise` index it was returning it instead of an empty `segmented` index.

audformat/core/utils.py
frankenjoe: > When can it happen that index.levels[2].empty?

Yeah, took me a while to understand why we need this. The reason we have to add this is that `pandas` stores an empty array if all `end` values are set to `NaT`. And unfortunately `pd.Index.union()` converts from `timedelta` to `datetime` in that case. Don't ask me why they do it, but this fixes it.

> Does it also mean if I pass two index objects with missing second level the returned index will then contain a second level?

As I just said, the second level is not really missing, it's just containing only `pd.NaT` and in that case `pandas` stores and empty array. Probably for efficiency, but not sure.

audformat/core/utils.py
hagenw: OK, can you then also update the test for `to_segmented_index()` to cover this error.

audformat/core/utils.py
hagenw: OK, makes all sense, but please add a comment explaining why we do it to the code.

frankenjoe: Honestly, the way pandas handles `timedelta` and `datetime` does not make much sense to me :)

I now added a comment explaining why we need this.

audformat/core/utils.py
frankenjoe: done

audformat/core/utils.py
hagenw: Yes, I also don't like how `pandas` handles it.

Issue #32, opened 2021-02-03
Forbid duplicated index entries?


hagenw:
As some methods like `audformat.Table.drop_files()` will fail if the database contains duplicated index entries, we should maybe forbid this in the first place.

The question is if we can find a good place to do this. The first places that come to mind are `Column.set()` and maybe `Database.save()`. I guess, this will still not completely safe as you can also directly assign the dataframe, but maybe this is then not a bug, but a feature for the power user.

frankenjoe:
I think we should add it in three places:

* `audformat.segmented_index()`
* `audformat.filewise_index()`
* `audformat.Table()`

And we should also check how `audformat.Table.extend()` behaves.

In `Column.set()` you cannot add new segments/files, but only set values to the existing index.

hagenw:
BTW, there would be a way to support duplicated index entries in `drop_files`:

We need to check if we have a filewise index or a segmented and for a segmented index we should do:
```python
self._df.drop(index, level='file', inplace=True)
```
instead of
```python
self._df.drop(index, inplace=True)
```
And maybe use a list of files instead of an index.

frankenjoe:
Ok, but do we want to support duplicated entries?

I'd say no. E.g. if we have:

``` python
table = Table(audformat.filewise_index('f1', 'f1', 'f1'))
```

and later we use:

``` python
table.set(value, index=audformat.filewise_index('f1')
```

should we assign the value to the first, second, or third entry or all of them? And what about this case:

``` python
table.set([value] * 2, index=audformat.filewise_index(['f1', 'f1'])
```

Same problem with `get`:

``` python
table.get(audformat.filewise_index(['f1'])  # ?
table.get(audformat.filewise_index(['f1', 'f1'])   # ?
```

And I guess more problems like this will pop up...

hagenw:
I agree, we should not support it.

I just wanted to mention how you could still use `drop_files()` if you have a dataframe that has already dplicated entries.

frankenjoe:
OK!

So I suggest to create a helper function that checks for duplicated entries in the index and raises an error. Then we can call this function in various places where we need to ensure a unique index. Makes sense?

PR #45, opened 2021-02-18
Forbid duplicates

frankenjoe:
Closes https://github.com/audeering/audformat/issues/32

In https://github.com/audeering/audformat/issues/32 we decided that an index conform to `audformat` must not contain duplicates.

This is now checked by `audformat.assert_index()` which is called by `audformat.index_type()`.

### Examples:

```python
index = audformat.segmented_index(['f1', 'f2'])
audformat.assert_index(index)

try:
    audformat.segmented_index(['f1', 'f1', 'f2', 'f2'])
except Exception as ex:
    print(ex)

try:
    audformat.assert_index(pd.Index(['f1']))
except Exception as ex:
    print(ex)

try:
    audformat.assert_index(pd.MultiIndex.from_arrays(
        [
            ['f1'], [0],
        ]
    ))
except Exception as ex:
    print(ex)

try:
    audformat.assert_index(
        pd.MultiIndex.from_arrays(
            [
                ['f1'], [0], [1]
            ], names=['File', 'Start', 'End']
        )
    )
except Exception as ex:
    print(ex)
```
```
Index not conform to audformat. Found duplicates:
('f1', Timedelta('0 days 00:00:00'), NaT)
('f2', Timedelta('0 days 00:00:00'), NaT)
Index not conform to audformat. Found single level with name None, but expected name 'file'.
Index not conform to audformat. Found 2 levels, but expected 1 or 3 levels.
Index not conform to audformat. Found three levels with names ['File', 'Start', 'End'], but expected names ['file', 'start', 'end'].
```

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -220,6 +220,9 @@ def test_pick_files(files, table):
     table = table.pick_files(files, inplace=False)
     if callable(files):
         files = table.files[table.files.to_series().apply(files)]
+        seen = set()
+        seen_add = seen.add
+        files = [x for x in files if not (x in seen or seen_add(x))]
     elif isinstance(files, str):
         files = [files]
     pd.testing.assert_index_equal(
--- a/audformat/__init__.py
+++ b/audformat/__init__.py
@@ -3,6 +3,7 @@ from audformat import errors
 from audformat import utils
 from audformat.core.database import Database
 from audformat.core.index import (
+    assert_index,
     filewise_index,
     segmented_index,
     index_type,
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -52,15 +52,19 @@ def assert_index(
         obj = obj.index
 
     if obj.has_duplicates:
-        duplicates = '\n'.join(
+        max_display = 10
+        duplicates = obj[obj.duplicated()]
+        msg_tail = '\n...' if len(duplicates) > max_display else ''
+        msg_duplicates = '\n'.join(
             [
-                str(duplicate) for duplicate in obj[obj.duplicated()].tolist()
+                str(duplicate) for duplicate
+                in duplicates[:max_display].tolist()
             ]
         )
         raise ValueError(
             'Index not conform to audformat. '
             'Found duplicates:\n'
-            f'{duplicates}'
+            f'{msg_duplicates}{msg_tail}'
         )
 
     num = len(obj.names)
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -119,6 +119,9 @@ def filewise_index(
     Returns:
         filewise index
 
+    Raises:
+        ValueError: if created index contains duplicates
+
     Example:
         >>> filewise_index(['a.wav', 'b.wav'])
         Index(['a.wav', 'b.wav'], dtype='object', name='file')
@@ -126,8 +129,12 @@ def filewise_index(
     """
     if files is None:
         files = []
+
     files = to_array(files)
-    return pd.Index(files, name=define.IndexField.FILE)
+    index = pd.Index(files, name=define.IndexField.FILE)
+    assert_index(index)
+
+    return index
 
 
 def index_type(
@@ -189,6 +196,9 @@ def segmented_index(
         segmented index
 
     Raises:
+        ValueError: if created index contains duplicates
+
+    Raises:
         ValueError: if ``files``, ``start`` and ``ends`` differ in size
 
     Example:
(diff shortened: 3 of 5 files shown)


hagenw:
I see the purpose behind the changes, but it also leads to some problems:

```python
>>> index = audformat.segmented_index(['f1', 'f1', 'f2', 'f2'])
>>> audformat.index_type(index)
...
ValueError: Index not conform to audformat. Found duplicates:
('f1', Timedelta('0 days 00:00:00'), NaT)
('f2', Timedelta('0 days 00:00:00'), NaT)
```
This makes no sense from a users perspective.
I would expect the following instead:
```python
>>> index = audformat.segmented_index(['f1', 'f1', 'f2', 'f2'])
...
ValueError: Index not conform to audformat. Found duplicates:
('f1', Timedelta('0 days 00:00:00'), NaT)
('f2', Timedelta('0 days 00:00:00'), NaT)
```

hagenw:
As a solution I would propose to indeed raise the error also inside `audformat.segmented_index()` and `audformat.filewise_index()`, but maybe at an option to those functions to disable the check and allow creating an index with duplicated entries as some users might want to use it anyway.

frankenjoe:
>  would propose to indeed raise the error also inside audformat.segmented_index()

Oh yes, absolutely. I forgot about that.

> allow creating an index with duplicated entries as some users might want to use it anyway.

You think this is a good idea? If it does not fail when creating the index it will fail later when the index is used e.g. to create a table.

hagenw:
> > allow creating an index with duplicated entries as some users might want to use it anyway.
> 
> You think this is a good idea? If it does not fail when creating the index it will fail later when the index is used e.g. to create a table.

No, it's fine with me to just raise the error. I thought it was on purpose that you didn't included it there.

frankenjoe:
In fact, maybe we should only call `assert_index()` in `audformat.filewise_index()` and `audformat.segmented_index()`. I gues we can assume that most of the time a user uses these functions to create an index, so there is is no need to check it later in every other function again. If the user for some reason manually creates the index, we can say it's her own risk and of course she can always call `assert_index()` manually, too. What do you think?

hagenw:
But in `audformat.index_type()` we state that we raise a `ValueError` if index not conform, so I guess we also have to raise the error there as well.

frankenjoe:
In any case, we can still raise an error if the number of levels or the level names are wrong as we did before. I'm just a bit worried that always checking for duplicates might be slow. Let's do a quick check how greedy that function is.

hagenw:
> In any case, we can still raise an error if the number of levels or the level names are wrong as we did before. I'm just a bit worried that always checking for duplicates might be slow. Let's do a quick check how greedy that function is.

That's a good point, maybe `audformat.index_type()` should not raise an error at all, but return a third index type, e.g. `unknown`. But I guess it's a little bit late to change this now.

frankenjoe:
Checking for duplicates seems not to be a bottleneck:

```python
n = 1000000

index = audformat.segmented_index(
    [str(idx) for idx in range(n)]
)

# without duplicate

t = time.time()
audformat.assert_index(index)
print(time.time()-t)

# with duplicates

index = audformat.segmented_index(
    [str(idx % n // 2) for idx in range(n)]
)
t = time.time()
try:
    audformat.assert_index(index)
except Exception as ex:
    print(ex)
print(time.time()-t)
```
```
0.05043768882751465
Index not conform to audformat. Found duplicates:
('0', Timedelta('0 days 00:00:00'), NaT)
('1', Timedelta('0 days 00:00:00'), NaT)
('2', Timedelta('0 days 00:00:00'), NaT)
('3', Timedelta('0 days 00:00:00'), NaT)
('4', Timedelta('0 days 00:00:00'), NaT)
('5', Timedelta('0 days 00:00:00'), NaT)
('6', Timedelta('0 days 00:00:00'), NaT)
('7', Timedelta('0 days 00:00:00'), NaT)
('8', Timedelta('0 days 00:00:00'), NaT)
('9', Timedelta('0 days 00:00:00'), NaT)
0.0592348575592041
```

frankenjoe:
We now also call `assert_index()` in `segmented_index()` and `filewise_index()`. I updated the example in the description accordingly. Thanks for spotting!

tests/test_table.py
```diff
@@ -220,6 +220,9 @@ def test_pick_files(files, table):
     table = table.pick_files(files, inplace=False)
     if callable(files):
         files = table.files[table.files.to_series().apply(files)]
+        seen = set()
+        seen_add = seen.add
+        files = [x for x in files if not (x in seen or seen_add(x))]
```
hagenw, line 225: What is this doing?

frankenjoe: Removes duplicates. We need this now since otherwise `audformat.create_filewise_index()` in the assertion will fail.

frankenjoe: And we cannot use `set()` here as we need to keep the ordering.

Issue #48, opened 2021-02-23
Introduce Table.update()


frankenjoe:
Let's have a look at the following example:

```python
db1 = audformat.testing.create_db(minimal=True)
db1.schemes['int'] = audformat.Scheme(int)
audformat.testing.add_table(db1, 'table', 'filewise')

db2 = audformat.testing.create_db(minimal=True)
db2.schemes['str'] = audformat.Scheme(str)
audformat.testing.add_table(db2, 'table', 'filewise')

audformat.utils.concat(
    [
        db1['table'].df,
        db2['table'].df
    ],
)
```
```
               int         str
file                          
audio/001.wav   47  NhDdzJX0YA
audio/002.wav   43  qdd5Onf25M
audio/003.wav   76  UIUHk0dGdi
audio/004.wav   97  WCXPkAxL5M
audio/005.wav   25  Aos20DyoUo
```

As we see we have successfully combined tables from two different database using `audformat.utils.concat()`.

Hence, a user might be surprised to learn that adding the two tables with `+` raises an error:

```python
(db1['table'] + db2['table']).get()
```
```
Bad column ID 'str', expected one of ['int']
```

The reason is that the new `Table` we create is actually assigned to `db1`:

https://github.com/audeering/audformat/blob/c045b928ddab9fbb9fc4508b0d4ca242c5f313cb/audformat/core/table.py#L795

But at the same time, we assign the the second column to scheme `str`, which is only known by `db2`:

https://github.com/audeering/audformat/blob/c045b928ddab9fbb9fc4508b0d4ca242c5f313cb/audformat/core/table.py#L790

So currently we can only safely combine tables with `+` from the same database. To overcome this limitation I suggest to not assign the new table to `db1` and set the `scheme_id` and `rater_id` of all columns to `None`. Then we can safely do:

```python
(db1['table'] + db2['table']).get()
```
```
               int         str
file                          
audio/001.wav   49  vaRA2y1rj4
audio/002.wav   42  IhpThA81cj
audio/003.wav   51  2oRdTHlYNr
audio/004.wav   31  ePZfabVHvd
audio/005.wav    5  FiZYPYcQZI
```

The only downside is that if we do:

```python
db['table1'] = db['table1'] + db['table2']  # removes scheme and rater from table1
```

we actually remove the scheme and rater information from `table1`.

Therefore I suggest to introduce a new function `Table.update()` for this use-case:

```python
db['table1'].update(db['table2'])
```

[...]

PR #49, opened 2021-02-24
Table.__add__ does not assign result to Database

frankenjoe:
Relates to #48 

Since the result of `+` is not assigned to a `Database` object, the following now works:

```python
db1 = audformat.testing.create_db(minimal=True)
db1.schemes['int'] = audformat.Scheme(int)
audformat.testing.add_table(db1, 'table', 'filewise')

db2 = audformat.testing.create_db(minimal=True)
db2.schemes['str'] = audformat.Scheme(str)
audformat.testing.add_table(db2, 'table', 'filewise')

(db1['table'] + db2['table']).get()
```
```
               int         str
file                          
audio/001.wav   22  5rIopnlMaW
audio/002.wav   66  O5UXXBIQmL
audio/003.wav   56  LDEXtoA1Yo
audio/004.wav   10  VyaeKKbuuX
audio/005.wav   83  SJpIxbNsIm
```

Also fixes a bug in `utils.to_segmented_index()` if an empty `pd.Series` or `pd.DataFrame` is passed - before the result was a `pd.Index` in that case. An according test to cover this case is added to `test_utils`.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -762,36 +762,13 @@ class Table(HeaderBase):
             ValueError: if values in the same position do not match
 
         """
-        if self == other:
-            return self
-
-        # concatenate table data
         df = utils.concat([self._df, other._df])
         if isinstance(df, pd.Series):
             df = df.to_frame()
 
-        # create table
-        media_id = self.media_id if self.media_id == other.media_id else None
-        split_id = self.split_id if self.split_id == other.split_id else None
-        table = Table(df.index, media_id=media_id, split_id=split_id)
-        table._db = self._db
-
-        # add columns
-        scheme_ids = {}
-        rater_ids = {}
-        for column_id, column in self.columns.items():
-            scheme_ids[column_id] = column.scheme_id
-            rater_ids[column_id] = column.rater_id
-        for column_id, column in other.columns.items():
-            scheme_ids[column_id] = column.scheme_id
-            rater_ids[column_id] = column.rater_id
+        table = Table(df.index)
         for column_id in df:
-            table[column_id] = Column(
-                scheme_id=scheme_ids[column_id],
-                rater_id=rater_ids[column_id],
-            )
-
-        # set table data
+            table[column_id] = Column()
         table._df = df
 
         return table
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -489,8 +489,6 @@ def to_segmented_index(
     """
     if index_type(obj) == define.IndexType.SEGMENTED:
         return obj
-    elif obj.empty:
-        return segmented_index()
 
     if isinstance(obj, (pd.Series, pd.DataFrame)):
         index = obj.index
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -67,7 +67,7 @@ def test_access():
                 )
             ),
         ),
-        # empty + content
+        # content + empty
         (
             [
                 create_table(
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -90,6 +90,7 @@ def test_access():
                 )
             ),
         ),
+        # empty + content
         (
             [
                 create_table(
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -59,6 +59,23 @@ def test_access():
                         dtype=float,
                     )
                 ),
+            ] * 3,
+            create_table(
+                pd.Series(
+                    index=audformat.filewise_index(),
+                    dtype=float,
+                )
+            ),
+        ),
+        # empty + content
+        (
+            [
+                create_table(
+                    pd.Series(
+                        [1.],
+                        index=audformat.filewise_index('f1'),
+                    )
+                ),
                 create_table(
                     pd.Series(
                         index=audformat.filewise_index(),
@@ -68,8 +85,30 @@ def test_access():
             ],
             create_table(
                 pd.Series(
-                    index=audformat.filewise_index(),
-                    dtype=float,
+                    [1.],
+                    index=audformat.filewise_index('f1'),
+                )
+            ),
+        ),
+        (
+            [
+                create_table(
+                    pd.Series(
+                        index=audformat.filewise_index(),
+                        dtype=float,
+                    )
+                ),
+                create_table(
+                    pd.Series(
+                        [1.],
+                        index=audformat.filewise_index('f1'),
+                    )
+                ),
+            ],
+            create_table(
+                pd.Series(
+                    [1.],
+                    index=audformat.filewise_index('f1'),
                 )
             ),
         ),
@@ -113,14 +152,20 @@ def test_access():
             [
                 create_table(
                     pd.Series(
-                        [1., 2., np.nan],
-                        index=audformat.filewise_index(['f1', 'f2', 'f3']),
+                        [1., np.nan],
+                        index=audformat.filewise_index(['f1', 'f2']),
                     )
                 ),
                 create_table(
                     pd.Series(
-                        [2., 3., 4.],
-                        index=audformat.filewise_index(['f2', 'f3', 'f4']),
+                        [2., 3.],
+                        index=audformat.filewise_index(['f2', 'f3']),
+                    )
+                ),
+                create_table(
+                    pd.Series(
+                        [3., 4.],
+                        index=audformat.filewise_index(['f3', 'f4']),
                     )
                 ),
             ],
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -1,3 +1,5 @@
+import typing
+
 import pytest
 import numpy as np
 import pandas as pd
@@ -8,6 +10,20 @@ import audformat
 import audformat.testing
 
 
+def create_table(
+        obj: typing.Union[pd.Series, pd.DataFrame],
+) -> audformat.Table:
+    r"""Helper function to create Table."""
+    table = audformat.Table(obj.index)
+    if isinstance(obj, pd.Series):
+        obj = obj.to_frame()
+    for name in obj:
+        table[name] = audformat.Column()
+        table[name].set(obj[name].values)
+    table._df = table.df.astype(obj.dtypes)
+    return table
+
+
 def test_access():
     db = audformat.testing.create_db()
     for table_id in db.tables.keys():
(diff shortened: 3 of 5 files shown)


audformat/core/table.py
```diff
@@ -762,36 +762,13 @@ def __add__(self, other: 'Table') -> 'Table':
             ValueError: if values in the same position do not match
 
         """
-        if self == other:
-            return self
```
hagenw, line 766: Why are you removing this? I think it covered the case `A + A` correct?

audformat/core/utils.py
```diff
@@ -489,8 +489,6 @@ def to_segmented_index(
     """
     if index_type(obj) == define.IndexType.SEGMENTED:
         return obj
-    elif obj.empty:
-        return segmented_index()
```
hagenw, line 493: You just added this a few days ago in https://github.com/audeering/audformat/pull/44,
was there no reasoning behind it?

tests/test_table.py
```diff
@@ -8,14 +10,224 @@
...
+    ]
+)
+def test_add(tables, expected):
+    table = tables[0]
+    for other in tables[1:]:
+        table += other
+    assert table == expected
+
+
+def test_add_2():  # TODO: turn into test_update()
```
hagenw, line 230: Is the TODO for another pull request?

tests/test_table.py
```diff
@@ -8,14 +10,224 @@
...
+                    [
+                        pytest.DB['files'].df,
+                        pytest.DB['segments'].df,
+                    ]
+                )
+            )
+        )
+    ]
+)
+def test_add(tables, expected):
```
hagenw, line 223: I think we are missing tests for adding an empty table to a table with content and the other way around.

Also, all the tests above check what is happing if you do `a + b`. Can we assume that `a + b + c` will then automatically work as well?

audformat/core/table.py
frankenjoe: Not anymore, because we want `_db`, `scheme_id`, etc. in the result set to `None`.

audformat/core/utils.py
frankenjoe: The fix was not correct as it only covered the case where `obj` is an `Index`, but `obj` can also be `pd.Series` or `pd.DataFrame` and in that case we don't want to return an index.

tests/test_table.py
frankenjoe: Yes, I will handle this in another MR.

audformat/core/table.py
hagenw: Because it should just return a table, and because schemes are not part of the table.
It still feels buggy to me.
Maybe we should in general rethink the objects, and make tables independent of the database. But I guess this is not easy as we had this dicussion before at some point and did not chnage it ;)

audformat/core/utils.py
hagenw: You said it covered the case where `obj` is an empty `Index`. How is this case now covered? or was the fix not neccessary before?

frankenjoe: Before we had:

```python
if obj.empty or index_type(obj) == define.IndexType.SEGMENTED:
             return obj
```

and there we had the problem that it would return an `filewise` index if the input was an object with a filewise index.

audformat/core/table.py
frankenjoe: If we make `tables` independent of the database we have to store schemes and raters with the table, but that means we cannot share them between tables anymore. I was also thinking about a solution where we add the option to temporarily store scheme and rater objects with a table until it is added to a database and then move them to the database. But that means adding a lot of complicated code so I'm more in favor of implementing the proposed `update()` method, which  is much easier to implement and maintain.

audformat/core/utils.py
frankenjoe: So basically I fixec a bug and introduced another bug. But let's blame the reviewer for not spotting  :)

audformat/core/table.py
hagenw: I would not move schemes and raters into the tables, but leaf them in the database.
What we have to change is that the information which table belongs to which schemes/raters is not stored inside the table, but inside the database.

tests/test_table.py
frankenjoe: I added the proposed tests.

audformat/core/table.py
hagenw: But I guess, this is what you propose here anyway. `A + A` will be different from `A` only when `A` was assigned to a database before. And if we reassing `A + A` to a database it will again be the same as `A`, correct?

frankenjoe: Ok, I see your point. But that means that we have to completely break the API, cause at the moment we do:

```python
table = Table(scheme_id='scheme')
db['table'] = table
```

and with your proposal have to change e.g. to:

```python
table = Table()
db.add_table('table, scheme_id='scheme')
```

frankenjoe: Actually the example above is not correct, cause `scheme_id` is part of `Column` at the moment, so instead of:

table = Table()
table['column'] = Column(scheme_id='scheme')
db['table'] = table
``` 

we have to do:

table = Table()
table['column'] = Column()
db.add_table('table, scheme_id={'column': 'scheme'})

Comment (frankenjoe) [reply audformat/core/table.py:766]: So I propose we first give the solution proposed in #48 a chance and if we are not happy with it, we can still think about a `audformat 2.0` someday :)

Comment (hagenw) [reply audformat/core/table.py:766]: BTW, it's even more complicated, because `scheme_id` is only part of `Column`, but for `Table` we have `media_id` and `split_id`.

I think we should at least think about breaking the API here, if it simplifies a lot of other stuff and makes it easier for the user to understand. But I have to first check or try to imagine how it would then be ;)

File: tests/test_table.py [hunk 5]
```diff
@@ -8,14 +10,269 @@
...
+                ),
+            ] * 3,
+            create_table(
+                pd.Series(
+                    index=audformat.filewise_index(),
+                    dtype=float,
+                )
+            ),
+        ),
+        # empty + content
```
hagenw, line 70:
```suggestion
        # content + empty
```

tests/test_table.py
```diff
@@ -8,14 +10,269 @@
...
+                ),
+            ],
+            create_table(
+                pd.Series(
+                    [1.],
+                    index=audformat.filewise_index('f1'),
+                )
+            ),
+        ),
+        (
```
hagenw, line 93:
```suggestion
        # empty + content
        (
```

tests/test_table.py
hagenw: The `a + b + c` test is now only for empty tables, would be nice to also have one for tables with content.

audformat/core/table.py
hagenw: I would also propose to first continue with the current solution.

I will create an issue regarding the `Table` discussion, remember we have also not reached `1.0.0` yet, so there is time to change it.

tests/test_table.py
frankenjoe: It's here: https://github.com/audeering/audformat/pull/49/files#diff-956061cd60efd54943151485eaa77ab422337bf1298e1f6094f118cd138b7c04R153

hagenw:
As `A + A` will return a detached table, what will be the result of something like:

```python
(db1['table'] + db2['table']).get(map={'speaker': ['age', 'gender']})
```

For me it is ok, if this is not working in this case, but we have to make sure that we raise a good error message.

audformat/core/table.py
hagenw: I created https://github.com/audeering/audformat/issues/50

tests/test_table.py
hagenw: Sorry, missed this. Thanks.

frankenjoe:
```python
db1 = audformat.testing.create_db(minimal=True)
db1.schemes['gender'] = audformat.Scheme(labels={0: 'female', 1: 'male'})
audformat.testing.add_table(db1, 'table', 'filewise')

db2 = audformat.testing.create_db(minimal=True)
db2.schemes['str'] = audformat.Scheme(str)
audformat.testing.add_table(db2, 'table', 'filewise')

table = db1['table'] + db2['table']
print(table['gender'].get(map='gender'))
```
ValueError: Column 'gender' is not assigned to a scheme.
```

**Comment by frankenjoe** [comment]
And:

```python
db1 = audformat.testing.create_db(minimal=True)
db1.schemes['speaker'] = audformat.Scheme(
    labels={
        0: {'gender': 'female', 'age': 20},
        1: {'gender': 'male', 'age': 30},
    }
)
audformat.testing.add_table(db1, 'table', 'filewise')

db2 = audformat.testing.create_db(minimal=True)
db2.schemes['str'] = audformat.Scheme(str)
audformat.testing.add_table(db2, 'table', 'filewise')

table = db1['table'] + db2['table']
print(table.get(map={'speaker': ['age', 'gender']}))
```
```
RuntimeError: Table is not assigned to a database.
```

frankenjoe:
And if we assign `table` again to a `Database`:

```python
db3 = audformat.testing.create_db(minimal=True)
db3['table'] = table
db3['table'].get(map={'speaker': ['age', 'gender']})
```
```
ValueError: Column 'speaker' is not assigned to a scheme.
```

hagenw:
In your example it might be easy to grasp for a user, but in this case it is not that simple
as now both databases have the same gender scheme:

```python
gender = audformat.Scheme(labels={0: 'female', 1: 'male'})
db1 = audformat.testing.create_db(minimal=True)
audformat.testing.add_table(db1, 'table', 'filewise')
db1.schemes['gender'] = gender

db2 = audformat.testing.create_db(minimal=True)
db2.schemes['str'] = audformat.Scheme(str)
db2.schemes['gender'] = gender
audformat.testing.add_table(db2, 'table', 'filewise')

table = db1['table'] + db2['table']
print(table['gender'].get(map='gender'))
```

```
ValueError: Column 'gender' is not assigned to a scheme.
```

So, maybe we change it to something like:
```
RuntimeError: 'map=' cannot be used with a detached table.
```
The downside is that a user also does not know what a detached table is ;)
So maybe we stay with the current error message for now.

frankenjoe:
Maybe:

```
RuntimeError: 'map=' cannot be used with a table that is not part of Database
```

?

frankenjoe:
But I'm also fine with keeping the current error message :)

hagenw:
BTW, the following does also not work, which is not what I would have expected:

```python
db1 = audformat.testing.create_db(minimal=True)
db1.schemes['gender'] = audformat.Scheme(labels={0: 'female', 1: 'male'})
audformat.testing.add_table(db1, 'table', 'filewise')
db1['table'].get(map='gender')
```
this results in
```
AttributeError                            Traceback (most recent call last)
<ipython-input-22-6923e1bd3313> in <module>
----> 1 db1['table'].get(map='gender')

~/git/audeering/audformat/audformat/core/table.py in get(self, index, map, copy)
    493                 result_is_copy = True  # to avoid another copy
    494 
--> 495             for column, mapped_columns in map.items():
    496                 mapped_columns = audeer.to_list(mapped_columns)
    497                 if len(mapped_columns) == 1:

AttributeError: 'str' object has no attribute 'items'
```

frankenjoe:
Yes, because It has to be:

```
db1['table'].get(map={'gender': 'gender'})
```

hagenw:
> Yes, because It has to be:
> 
> ```
> db1['table'].get(map={'gender': 'gender'})
> ```

Right, or

```python
db1['table']['gender'].get(map='gender')
```

hagenw:
> And if we assign `table` again to a `Database`:
> 
> ```python
> db3 = audformat.testing.create_db(minimal=True)
> db3['table'] = table
> db3['table'].get(map={'speaker': ['age', 'gender']})
> ```
> 
> ```
> ValueError: Column 'speaker' is not assigned to a scheme.
> ```

But if we assign it to a database with the correct scheme I would expect that it should work again, which it does not:

```python
db1 = audformat.testing.create_db(minimal=True)
db1.schemes['gender'] = audformat.Scheme(labels={0: 'female', 1: 'male'})
audformat.testing.add_table(db1, 'table', 'filewise')

db2 = audformat.testing.create_db(minimal=True)
db2.schemes['str'] = audformat.Scheme(str)
audformat.testing.add_table(db2, 'table', 'filewise')

table = db1['table'] + db2['table']
table['gender'].get(map='gender')
db1['table_combined'] = table
db1['table_combined'].get(map={'gender': 'gender'})
```

```
ValueError: Column 'gender' is not assigned to a scheme.
```

hagenw:
So basically you canot not combine tables and map schemes.
What you have to do instead is map schemes first and the combine the two dataframes, which might not provide the same result as `db1['table'] + db2['table']`. So I think the current solution is not ideal.

frankenjoe:
> But if we assign it to a database with the correct scheme I would expect that it should work again, which it does not:

you first have to assign the scheme again:

```python
table = db1['table'] + db2['table']
db1['table_combined'] = db1['table'] + db2['table']
db1['table_combined']['gender'].scheme_id = 'gender'
```

Or with #48 you can do:

```python
db1['table_combined'] = Table()
db1['table_combined'].update(db1['table'])
db1['table_combined'].update(db2['table'])
db1['table_combined'].get(map={'gender': 'gender'})
```

hagenw:
> > But if we assign it to a database with the correct scheme I would expect that it should work again, which it does not:
> 
> you first have to assign the scheme again:
> 
> ```python
> table = db1['table'] + db2['table']
> db1['table_combined'] = db1['table'] + db2['table']
> db1['table_combined']['gender'].scheme_id = 'gender'
> db1['table_combined'].get(map={'gender': 'gender'})
> ```
> 
> Or with #48 you can do:
> 
> ```python
> db1['table_combined'] = Table()
> db1['table_combined'].update(db1['table'])
> db1['table_combined'].update(db2['table'])
> db1['table_combined'].get(map={'gender': 'gender'})
> ```

Ah,, yes, forgot about `update()`, so tyhis is sufficient then.

PR #51, opened 2021-02-24
Add overwrite argument to utils.concat()

frankenjoe:
Relates to #48 

### Example:

```python
obj1 = pd.Series(
    [0., 0.],                 # assigns 0. to 'f2'
    index=audformat.filewise_index(['f1', 'f2']),
    name='float',
)
obj2 = pd.DataFrame(
    {
        'float': [1., 2.],    # assigns 1. to 'f2'
        'string': ['a', 'b'],
    },
    index=audformat.segmented_index(['f2', 'f3']),
)
```

Usually the following would raise an error:

```python
try:
    audformat.utils.concat([obj1, obj2])
except Exception as ex:
    print(ex)
```
```
Found overlapping data:
                 left  right
file start  end             
f2   0 days NaT   0.0    1.0
```

But with `overwrite=True` it gives:

```python
audformat.utils.concat([obj1, obj2], overwrite=True)
```
```
                 float string
file start  end              
f1   0 days NaT    0.0    NaN
f2   0 days NaT    1.0      a
f3   0 days NaT    2.0      b
```

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -174,7 +174,8 @@ def concat(
                             if len(overlap) > max_display \
                             else ''
                         raise ValueError(
-                            'Found overlapping data:\n'
+                            "Found overlapping data in column "
+                            f"'{column.name}':\n"
                             f"{msg_overlap}{msg_tail}"
                         )
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -81,14 +81,14 @@ def concat(
         ...         'float': [np.nan, 2.],
         ...         'string': ['a', 'b'],
         ...     },
-        ...     index=segmented_index(['f2', 'f3']),
+        ...     index=filewise_index(['f2', 'f3']),
         ... )
         >>> concat([obj1, obj2])
-                         float string
-        file start  end
-        f1   0 days NaT    0.0    NaN
-        f2   0 days NaT    1.0      a
-        f3   0 days NaT    2.0      b
+              float string
+        file
+        f1      0.0    NaN
+        f2      1.0      a
+        f3      2.0      b
         >>> obj1 = pd.Series(
         ...     [0., 0.],
         ...     index=filewise_index(['f1', 'f2']),
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -32,9 +32,11 @@ def concat(
 
     Columns with the same identifier are combined to a single column.
     This requires that both columns have the same dtype
-    and unless ``overwrite`` is set to ``True``,
+    and if ``overwrite`` is set to ``False``,
     values in places where the indices overlap have to match
     or one column contains ``NaN``.
+    If ``overwrite`` is set to ``True``,
+    the value of the last object in the list is kept.
 
     Args:
         objs: objects conform to
@@ -70,13 +72,13 @@ def concat(
         f2   0 days NaT    1.0      a
         f3   0 days NaT    2.0      b
         >>> obj1 = pd.Series(
-        ...     [0., np.nan],
+        ...     [0., 1.],
         ...     index=filewise_index(['f1', 'f2']),
         ...     name='float',
         ... )
         >>> obj2 = pd.DataFrame(
         ...     {
-        ...         'float': [1., 2.],
+        ...         'float': [np.nan, 2.],
         ...         'string': ['a', 'b'],
         ...     },
         ...     index=segmented_index(['f2', 'f3']),
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -18,6 +18,8 @@ from audformat.core.index import (
 
 def concat(
         objs: typing.Sequence[typing.Union[pd.Series, pd.DataFrame]],
+        *,
+        overwrite: bool = False,
 ) -> typing.Union[pd.Series, pd.DataFrame]:
     r"""Concatenate objects.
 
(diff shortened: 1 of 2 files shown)


frankenjoe:
@hagenw I'm not really happy with the argument name `overwrite`, but I cannot really think of a better name at the moment. Do you have an idea?

frankenjoe:
Maybe, `allow_overlap`?

hagenw:
I think we need to improve on the error message:
```
Found overlapping data:
                 left  right
file start  end             
f2   0 days NaT   0.0    1.0
```

Do the following:
```python
obj1 = pd.Series(
    [0., 0.],                 # assigns 0. to 'f2'
    index=audformat.filewise_index(['f1', 'f2']),
    name='float',
)
obj2 = pd.DataFrame(
    {
        'float': [1., 2.],    # assigns 1. to 'f2'
        'string': ['a', 'b'],
    },
    index=audformat.segmented_index(['f2', 'f3']),
)
obj3 = pd.DataFrame(
    {
        'float': [1., 3.],  # assigns 3. to 'f2'
        'string': ['a', 'c'],
    },
    index=audformat.segmented_index(['f4', 'f2']),
)
```
If you then run:
```python
audformat.utils.concat([obj1, obj2, obj3])
```
we get
```
ValueError: Found overlapping data:
                 left  right
file start  end             
f2   0 days NaT   0.0    1.0
```
But it's not obvious to the user what `left` and `right` refer to as we have more than two objects.
It's still possible to figure it out for 3 objects, but a user can also concatenate hundreds of objects.

hagenw:
BTW, this works as expected, which is great:

```python
>>> audformat.utils.concat([obj1, obj2, obj3], overwrite=True)
                 float string
file start  end              
f1   0 days NaT    0.0    NaN
f2   0 days NaT    3.0      c
f3   0 days NaT    2.0      b
f4   0 days NaT    1.0      a

>>> audformat.utils.concat([obj1, obj3, obj2], overwrite=True)
                 float string
file start  end              
f1   0 days NaT    0.0    NaN
f2   0 days NaT    1.0      a
f3   0 days NaT    2.0      b
f4   0 days NaT    1.0      a
```

hagenw:
> Maybe, `allow_overlap`?

`allow_overlap` is not better than `overwrite` as it does not tell the user what will happen on overlap. `overwrite` is ok, but it is not obvious what it is referring to (on overlap) and which dataframe takes over (the dataframe further to the right).

But for now, I also don't have a better idea than `overwrite`.

audformat/core/utils.py
```diff
@@ -67,6 +71,42 @@ def concat(
...
+        ...     [0., 1.],
+        ...     index=filewise_index(['f1', 'f2']),
+        ...     name='float',
+        ... )
+        >>> obj2 = pd.DataFrame(
+        ...     {
+        ...         'float': [np.nan, 2.],
+        ...         'string': ['a', 'b'],
+        ...     },
+        ...     index=segmented_index(['f2', 'f3']),
```
hagenw, line 84: Maybe change this also to a filewise index to show that in this case it will also return a filewise index.

frankenjoe:
> But it's not obvious to the user what left and right refer to as we have more than two objects.

please make a suggestion

tests/test_utils.py
```diff
@@ -343,13 +386,14 @@
...
             ],
+            False,
             None,
             marks=pytest.mark.xfail(raises=ValueError),
         ),
     ],
 )
-def test_concat(objs, expected):
-    obj = utils.concat(objs)
+def test_concat(objs, overwrite, expected):
```
hagenw, line 395: I think you should also add a few tests that combine three objects.

hagenw (approved):

frankenjoe:
E.g. would that be better:

```
ValueError: Found overlapping data in column 'float':
                 left  right
file start  end             
f2   0 days NaT   0.0    1.0
```

hagenw:
> E.g. would that be better:
> 
> ```
> ValueError: Found overlapping data in column 'float':
>                  left  right
> file start  end             
> f2   0 days NaT   0.0    1.0
> ```

Yes, and maybe enumerate the dataframes/series, e.g.
```
ValueError: Found overlapping data in column 'float':
                 obj1   obj2
file start  end             
f2   0 days NaT   0.0    1.0
```
Then you could also use `obj2 obj3` and it is more obvious where the error is happening.

hagenw:
On the other hand, if you just use two objects, `left right` is of course easier to understand.

tests/test_utils.py
frankenjoe: we now have:

```python
# combine series and data frame
        (
            [
                pd.Series(
                    [1., 2.],
                    name='c1',
                ),
                pd.Series(
                    ['a', np.nan, 'd'],
                    audformat.filewise_index(['f1', 'f2', 'f4']),
                    name='c2',
                ),
                pd.DataFrame(
                    {
                        'c1': [np.nan, 3.],
                        'c2': ['b', 'c'],
                    },
                    audformat.segmented_index(['f2', 'f3']),
                ),
            ],
            False,
            pd.DataFrame(
                {
                    'c1': [1., 2., 3., np.nan],
                    'c2': ['a', 'b', 'c', 'd']
                },
                audformat.segmented_index(['f1', 'f2', 'f3', 'f4']),
            ),
        ),
```

frankenjoe:
> Yes, and maybe enumerate the dataframes/series, e.g.

for `right` it's easy, but `left` can already be a concatenation of several columns

hagenw:
> > Yes, and maybe enumerate the dataframes/series, e.g.
> 
> for `right` it's easy, but `left` can already be a concatenation of several objects

Good point, then let's stay with `left` and `right`.

audformat/core/utils.py
frankenjoe: Good idea. I changed it accordingly.

PR #60, opened 2021-04-08
Add allow_nat argument to utils.to_segmented_index()

frankenjoe:
Adds  the option to replace `NaT` with file duration when converting to a segmented index.

image

### Example:

```python
import audformat

root = '/media/jwagner/Data/audb/emodb/1'
db = audformat.Database.load(root)
audformat.utils.to_segmented_index(db['emotion'].get())
```
```
                              emotion @emotion
file            start  end                    
wav/03a01Fa.wav 0 days NaT  happiness      0.9
wav/03a01Nc.wav 0 days NaT    neutral      1.0
wav/03a01Wa.wav 0 days NaT      anger     0.95
wav/03a02Fc.wav 0 days NaT  happiness     0.85
wav/03a02Nc.wav 0 days NaT    neutral      1.0
...                               ...      ...
wav/16b10Lb.wav 0 days NaT    boredom      1.0
wav/16b10Tb.wav 0 days NaT    sadness      0.9
wav/16b10Td.wav 0 days NaT    sadness     0.95
wav/16b10Wa.wav 0 days NaT      anger      1.0
wav/16b10Wb.wav 0 days NaT      anger      1.0
```
```python
audformat.utils.to_segmented_index(db['emotion'].get(), allow_nat=False, root=root)
```
```
                                                    emotion @emotion
file            start  end                                          
wav/03a01Fa.wav 0 days 0 days 00:00:01.898250     happiness      0.9
wav/03a01Nc.wav 0 days 0 days 00:00:01.611250       neutral      1.0
wav/03a01Wa.wav 0 days 0 days 00:00:01.877812500      anger     0.95
wav/03a02Fc.wav 0 days 0 days 00:00:02.006250     happiness     0.85
wav/03a02Nc.wav 0 days 0 days 00:00:01.439812500    neutral      1.0
...                                                     ...      ...
wav/16b10Lb.wav 0 days 0 days 00:00:03.442687500    boredom      1.0
wav/16b10Tb.wav 0 days 0 days 00:00:03.500625       sadness      0.9
wav/16b10Td.wav 0 days 0 days 00:00:03.934187500    sadness     0.95
wav/16b10Wa.wav 0 days 0 days 00:00:02.414125         anger      1.0
wav/16b10Wb.wav 0 days 0 days 00:00:02.522499999      anger      1.0
```

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -525,6 +525,8 @@ def to_segmented_index(
         *,
         allow_nat: bool = True,
         root: str = None,
+        num_workers: typing.Optional[int] = 1,
+        verbose: bool = False,
 ) -> typing.Union[pd.Index, pd.Series, pd.DataFrame]:
     r"""Convert to segmented index.
 
@@ -545,7 +547,12 @@ def to_segmented_index(
             :ref:`table specifications <data-tables:Tables>`
         allow_nat: if set to ``False``, ``end=NaT`` is replaced with file
             duration
-        root: root directory to files in the index
+        root: root directory under which the files referenced in the index
+            are stored
+        num_workers: number of parallel jobs.
+            If ``None`` will be set to the number of processors
+            on the machine multiplied by 5
+        verbose: show progress bar
 
     Returns:
         object with segmented index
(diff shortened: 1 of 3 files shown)


frankenjoe:
@hagenw I have two questions:

1. you think it's a good idea to make this part of `to_segmented_index()` or you think it's better to have a separate function, e.g.:

```python
segmented_index = audformat.to_segmented_index(filewise_index)
segmented_index_without_nat = audformat.utils.replace_nat(segmented_index)
```

I'm a bit undecided, but probably it's more convenient to have only one function. It's also not clear what we should do if we pass a `filewise` index to `audformat.utils.replace_nat()`, probably simply return it.

2. I guess it's not a bad idea to add an option for multi-threading for speed up on large tables.

hagenw:
1. I think it makes sense to add it to `to_segmented_index()` as it seems a valid use case to me. Do we have another function that could add `NaT` to an index? Then it might be worth to make an extra function out of it.
2. Multi-threading also sounds like a good idea. The only thing that is a little bit strange is that we then have 4 arguments in `to_segmented_index()` all covering a rare usecase.

frankenjoe:
>     2\. all covering a rare usecase.

We will actually need it quite a lot soon - I would like to add a couple of other utility functions soon (like inverting an index etc.) where it's needed.

frankenjoe:
>     1. that could add `NaT` to an index

I'm not aware of any at the moment. So let's keep it within `segmented_index()` for now and if it makes sense we can later move it to a separate function that we call within `segmented_index()`.

frankenjoe:
>     2\. Multi-threading also sounds like a good idea.

Ok, I will add it. Putting WIP in the meantime.

frankenjoe:
@hagenw I added an option for multithreading and removed the WIP.

PR #66, opened 2021-04-29
Add audformat.utils.join_labels()

hagenw:
As discussed in https://github.com/audeering/audformat/pull/66#pullrequestreview-648192732 this adds `audformat.utils.join_labels()` as a helper function to join scheme labels before merging two databases.

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -286,6 +286,89 @@ def intersect(
     return index
 
 
+def join_labels(
+        labels: typing.Sequence[typing.Union[typing.List, typing.Dict]],
+):
+    r"""Combine scheme labels.
+
+    This might be helpful,
+    if you would like to combine two databases
+    that have the same schene,
+    but with different labels:
+
+    .. code-block:: python
+
+        labels = audformat.utils.join_labels(
+            [
+                db.schemes['scheme'].labels,
+                db_new.schemes['scheme'].labels,
+            ]
+        )
+        db.schemes['scheme'].replace_labels(labels)
+        db_new.schemes['scheme'].replace_labels(labels)
+        db.update(db_new)
+
+    Args:
+        labels: sequence of labels to join
+
+    Returns:
+        joined labels
+
+    Raises:
+        ValueError: if labels are of different type
+        ValueError: if labels are dicts
+            and contain different values for the same key
+        RuntimeError: if label type is not ``list`` or ``dict``
+
+    Example:
+        >>> join_labels([{'a': 0, 'b': 1}, {'b': 1, 'c': 2}])
+        {'a': 0, 'b': 1, 'c': 2}
+
+    """
+    if len(labels) == 0 or isinstance(labels, dict):
+        return labels
+
+    if not isinstance(labels, list):
+        labels = list(labels)
+
+    label_type = type(labels[0])
+    joined_labels = labels[0]
+
+    for label in labels[1:]:
+        if type(label) != label_type:
+            raise ValueError(
+                f"Labels are of different type:\n"
+                f"{label_type}\n"
+                f"!=\n"
+                f"{type(label)}"
+            )
+
+    if label_type == dict:
+        for label in labels[1:]:
+            for key, value in label.items():
+                if key not in joined_labels:
+                    joined_labels[key] = value
+                elif joined_labels[key] != value:
+                    raise ValueError(
+                        f"Values for key '{key}' are different:\n"
+                        f"{joined_labels[key]}\n"
+                        f"!=\n"
+                        f"{value}"
+                    )
+    elif label_type == list:
+        joined_labels = list(
+            set(list(joined_labels) + audeer.flatten_list(labels[1:]))
+        )
+        joined_labels = sorted(audeer.flatten_list(joined_labels))
+    else:
+        raise RuntimeError(
+            f"Supported label types are 'list' and 'dict', "
+            f"but your is '{label_type}'"
+        )
+
+    return joined_labels
+
+
 def map_language(language: str) -> str:
     r"""Map language to ISO 639-3.
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -309,20 +309,21 @@ def join_labels(
         db.update(db_new)
 
     Args:
-        labels: sequence of labels to join
+        labels: sequence of labels to join.
+            For dictionary labels,
+            labels further to the right
+            can overwrite previous labels
 
     Returns:
         joined labels
 
     Raises:
         ValueError: if labels are of different type
-        ValueError: if the labels are dicts
-            and contain different values for the same key
         RuntimeError: if label type is not ``list`` or ``dict``
 
     Example:
-        >>> join_labels([{'a': 0, 'b': 1}, {'b': 1, 'c': 2}])
-        {'a': 0, 'b': 1, 'c': 2}
+        >>> join_labels([{'a': 0, 'b': 1}, {'b': 2, 'c': 2}])
+        {'a': 0, 'b': 2, 'c': 2}
 
     """
     if len(labels) == 0 or isinstance(labels, dict):
@@ -346,15 +347,8 @@ def join_labels(
     if label_type == dict:
         for label in labels[1:]:
             for key, value in label.items():
-                if key not in joined_labels:
+                if key not in joined_labels or joined_labels[key] != value:
                     joined_labels[key] = value
-                elif joined_labels[key] != value:
-                    raise ValueError(
-                        f"Values for key '{key}' are different:\n"
-                        f"{joined_labels[key]}\n"
-                        f"!=\n"
-                        f"{value}"
-                    )
     elif label_type == list:
         joined_labels = list(
             set(list(joined_labels) + audeer.flatten_list(labels[1:]))
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -326,7 +326,7 @@ def join_labels(
         {'a': 0, 'b': 2, 'c': 2}
 
     """
-    if len(labels) == 0 or isinstance(labels, dict):
+    if len(labels) == 0:
         return labels
 
     if not isinstance(labels, list):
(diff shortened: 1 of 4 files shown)


frankenjoe:
> This is a proposal of how to add the `join_scheme_labels` argument suggested in [#62 (comment)](https://github.com/audeering/audformat/pull/62#issuecomment-828212230).

Puhh, good you pushed this early, a minute ago I started to work on the same issue :)

Glad to only have to review this one :)

audformat/core/database.py
```diff
@@ -514,11 +522,51 @@ def join_field(
                 elif value2:
                     self.__dict__[field] = value2
 
+        def update_scheme(
+                other: Database,
+                scheme: str,
+        ):
+            r"""Update schemes to contain the same labels"""
+            self_type = type(self.schemes[scheme].labels)
```
frankenjoe, line 530: I suggest to use `self.schemes.dtype` instead

audformat/core/database.py
```diff
@@ -514,11 +522,51 @@ def join_field(
                 elif value2:
                     self.__dict__[field] = value2
 
+        def update_scheme(
```
frankenjoe, line 525: One thing that is not so nice about implementing it as a method of `Scheme` is that it will change both - `self` and `other`, which might be a bit unusual. Maybe it makes more sense to have it as a utility function, e.g. `utils.join_schemes(...)`?

hagenw: Good point.

In my opinion it shouldn't update `other`, so maybe we have to work with a copy instead, if this is possible?

audformat/core/database.py
hagenw: I changed it accordingly

audformat/core/database.py
frankenjoe: But since changing the scheme labels also changes the table content this means we have to copy the whole database, don't think that is a good idea.

So maybe it's better that the user has to explicitly change the schemes in both database before calling `update()`. Something like:

```python
labels = audformat.utils.join-labels([db.schemes['scheme'].labels, [db_new.schemes['scheme'].labels])
db.schemes['scheme'].replace_labels(labels)
db_new.schemes['scheme'].replace_labels(labels)
db.update(db_new)
```

The downside is that it's of course a bit lengthy, but at least there are now hidden side-effects the user is not aware of.

hagenw: It's a good point.

I updated the pull request and propose now only to introduce a helper function.

audformat/core/utils.py
```diff
@@ -286,6 +286,95 @@ def intersect(
...
+        ValueError: if the labels are dicts
+            and contain different values for the same key
+        RuntimeError: if label type is not ``list`` or ``dict``
+
+    Example:
+        >>> join_labels([{'a': 0, 'b': 1}, {'b': 1, 'c': 2}])
+        {'a': 0, 'b': 1, 'c': 2}
+
+    """
+    if isinstance(labels, dict):
```
frankenjoe, line 329: Why do you think we need this check? The docstring clearly states that the user has to pass a list.

audformat/core/utils.py
```diff
@@ -286,6 +286,95 @@ def intersect(
...
+        raise ValueError(
+            "'labels' has to be a sequence, yours is a 'dict'"
+        )
+
+    if len(labels) == 0 or isinstance(labels, dict):
```
frankenjoe, line 334:
```suggestion
    if len(labels) == 0:
```

audformat/core/utils.py
```diff
@@ -286,6 +286,95 @@ def intersect(
...
+        return labels
+
+    if not isinstance(labels, list):
```
frankenjoe, line 337: Again, why do we need this when we expect a list as input?

frankenjoe:
Currently the following is possible:

```python
labels_1 = {
    0: {'age': 20},
}

labels_2 = {
    '0': {'age': 30},
}

labels = audformat.utils.join_labels([labels_1, labels_2])
print(labels)
```
```
{0: {'age': 20}, '0': {'age': 30}}
```

So I wonder if we should check for same key type when we merge dictionaries?

```

**Comment by frankenjoe** [comment]
Currently, it's not possible to update dictionaries with new values:

```python
labels_1 = {
    0: {'age': 20},
}

labels_2 = {
    0: {'gender': 'male'},
}

labels = audformat.utils.join_labels([labels_1, labels_2])
print(labels)
```
```
ValueError: Values for key '0' are different:
{'age': 20}
!=
{'gender': 'male'}
```

It would be nice if we get:

```
{0: {'age': 20, 'gender': 'male'}}
```

hagenw:
> Currently, it's not possible to update dictionaries with new values:
> 
> ```python
> labels_1 = {
>     0: {'age': 20},
> }
> 
> labels_2 = {
>     0: {'gender': 'male'},
> }
> 
> labels = audformat.utils.join_labels([labels_1, labels_2])
> print(labels)
> ```
> 
> ```
> ValueError: Values for key '0' are different:
> {'age': 20}
> !=
> {'gender': 'male'}
> ```
> 
> It would be nice if we get:
> 
> ```
> {0: {'age': 20, 'gender': 'male'}}
> ```

Good point, I now allow for overwriting and updated the documentation screenshot in the description.

audformat/core/utils.py
hagenw: No, you are right, I removed it.

audformat/core/utils.py
hagenw: We state in the docs that we accept a sequence as input, so it might be that we get tuple, which we need to convert to a list at this stage.

hagenw:
> Currently the following is possible:
> 
> ```python
> labels_1 = {
>     0: {'age': 20},
> }
> 
> labels_2 = {
>     '0': {'age': 30},
> }
> 
> labels = audformat.utils.join_labels([labels_1, labels_2])
> print(labels)
> ```
> 
> ```
> {0: {'age': 20}, '0': {'age': 30}}
> ```
> 
> So I wonder if we should check for same key type when we merge dictionaries? An easy way to do so is trying to create a `Scheme` from the combined labels, which will raise an error if the new labels are malformed:
> 
> ```python
> audformat.Scheme(labels=labels)
> ```
> 
> ```
> ValueError: All labels must be of the same data type.
> ```

Have all dictionary keys to be of the same type in a scheme?
If yes, we should indeed check for it, if not we shouldn't.

hagenw:
> ```python
> '0': {'age': 30}
> ```

```python
>>> audformat.Scheme(labels={0: {'age': 20}, '0': {'age': 30}})
ValueError                                Traceback (most recent call last)
<ipython-input-2-db9ca95f1e49> in <module>
----> 1 scheme1 = audformat.Scheme(labels={0: {'age': 20}, '0': {'age': 30}})

~/git/audeering/audformat/audformat/core/scheme.py in __init__(self, dtype, labels, minimum, maximum, description, meta)
     83 
     84         if labels is not None:
---> 85             dtype_labels = self._dtype_from_labels(labels)
     86             if dtype is not None and dtype != dtype_labels:
     87                 raise ValueError(

~/git/audeering/audformat/audformat/core/scheme.py in _dtype_from_labels(self, labels)
    292         if not all(isinstance(x, dtype) for x in list(labels)):
    293             raise ValueError(
--> 294                 'All labels must be of the same data type.'
    295             )
    296 

ValueError: All labels must be of the same data type.
```
which means we should indeed make sure that this cannot happen.

hagenw:
> ```python
> '0': {'age': 30}
> ```

Implemented it:

```python
>>> audformat.utils.join_labels([{0: {'age': 20}}, {'0': {'age': 30}}])
...
ValueError: All labels must be of the same data type.
```
also added a test for it.

audformat/core/utils.py
frankenjoe: But are you sure it's necessary? Or is it the `audeer.flatten_list()` that requires a `list` as input?

hagenw: At the moment yes. Before I tried to implement a version without `audeer.flatten_list()`, but there I also got an error for tuples.
So, I haven't found a better way for now. As we want to return always a list we have to convert at some point anyway.

audformat/core/utils.py
```diff
@@ -286,6 +287,87 @@ def intersect(
...
+            for key, value in label.items():
+                if key not in joined_labels or joined_labels[key] != value:
+                    joined_labels[key] = value
+    elif label_type == list:
+        joined_labels = list(
+            set(list(joined_labels) + audeer.flatten_list(labels[1:]))
+        )
+        joined_labels = sorted(audeer.flatten_list(joined_labels))
+    else:
+        raise RuntimeError(
```
frankenjoe, line 359: Would it make more sense to raise a `ValueError` or `TypeError` here?

audformat/core/utils.py
```diff
@@ -286,6 +287,87 @@ def intersect(
...
+
+    if not isinstance(labels, list):
+        labels = list(labels)
+
+    label_type = type(labels[0])
+    joined_labels = labels[0]
+
+    for label in labels[1:]:
+        if type(label) != label_type:
+            raise ValueError(
```
frankenjoe, line 341: Or a `TypeError`?

frankenjoe: But actually in `Scheme` we also raise `ValueError` when the types do not match, so maybe it's better to stick with `ValueError`.

audformat/core/utils.py
hagenw: It's in general not very pretty the code for joining the labels. I guess there must be an easier solution.

audformat/core/utils.py
frankenjoe: Probably `ValueError` (see comment above).

audformat/core/utils.py
hagenw: It's in general not very pretty the code for joining the labels. I guess there must be an easier solution.

audformat/core/utils.py
hagenw: I switched to `ValueError`

audformat/core/utils.py
hagenw: I would vote to stick with `ValueError`

PR #71, opened 2021-05-06
Don't compress table pickle files

### LINKED ISSUE #64: Speed up caching of tables
As shown in https://github.com/audeering/audb/pull/38#issuecomment-828990575 using uncompressed pickle files is much faster when storing large dataframes. As we don't care about size of the cached tables, but only about speed for loading them we should change the behavior for caching tables as well.

I think we can even implement it in a backward compatible way. Loading a compressed pickle file without specifying the compression, it should fail and we can use `try`-`except` to catch that and then load using `compression='xz'.

hagenw:
Closes #64.

As we discovered with `audb` storing pickle files without copmression makes saving and loading them much faster,
so we will disable compression here as well as we don't bother too much about the size of the pickle file, but mostly about the reading speed of a table.

It's implemented with a `try`-`except` to still be compatible with tables stored with `xz` compression.
It also adds a test for that case.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1,5 +1,6 @@
 import copy
 import os
+import pickle
 import typing
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


frankenjoe:
Makes sense, no comments.

PR #73, opened 2021-05-07
Add copy_media to audformat.Database.update()

hagenw:
This adds the `copy_media=False` argument to `audformat.Database.update()` which allows you to automatically copy media files when updating a database by another one.

I have set it to `False` as default, which is the saver choice even though a user would maybe expect that media files are copied when updating a database.

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -1,6 +1,7 @@
 import datetime
 import itertools
 import os
+import shutil
 import typing
 
 import oyaml as yaml
@@ -429,6 +430,7 @@ class Database(HeaderBase):
             self,
             others: typing.Union['Database', typing.Sequence['Database']],
             *,
+            copy_media: bool = False,
             overwrite: bool = False,
     ) -> 'Database':
         r"""Update database with other database(s).
@@ -454,6 +456,8 @@ class Database(HeaderBase):
 
         Args:
             others: database object(s)
+            copy_media: if ``True`` it copies the media files
+                associated with ``others`` to the current database root folder
             overwrite: overwrite table values where indices overlap
 
         Returns:
@@ -465,6 +469,9 @@ class Database(HeaderBase):
                 same ID is found
             ValueError: if table data cannot be combined (e.g. values in
                 same position overlap)
+            RuntimeError: if ``copy_media=True``,
+                but one of the databases in ``others``
+                contains files but no root folder
 
         """
 
@@ -557,6 +564,21 @@ class Database(HeaderBase):
                     self[table_id].update(table, overwrite=overwrite)
                 else:
                     self[table_id] = table.copy()
+
+        # copy media files
+        if copy_media and self.root:
+            for other in others:
+                if len(other.files) > 0 and other.root is None:
+                    raise RuntimeError(
+                        f"The database '{other.name}' has no root folder."
+                    )
+                for file in other.files:
+                    # Joining works also if the paths are already absolute
+                    src_file = os.path.join(other.root, file)
+                    dst_file = os.path.join(self.root, file)
+                    dst_dir = os.path.dirname(dst_file)
+                    audeer.mkdir(dst_dir)
+                    shutil.copy(src_file, dst_file)
 
         return self
 
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -310,7 +310,7 @@ def test_string():
                       'languages: [deu, eng]'
 
 
-def test_update():
+def test_update(tmpdir):
 
     # original database
 
@@ -331,6 +331,9 @@ def test_update():
             'labels': ('labels', None),
         },
     )
+    db_root = audeer.mkdir(os.path.join(tmpdir, 'db'))
+    db.save(db_root)
+    audformat.testing.create_audio_files(db, db_root, file_duration='0.1s')
 
     assert db.update(db) == db
 
@@ -353,6 +356,9 @@ def test_update():
             'labels': ('labels', None),
         },
     )
+    db_root = audeer.mkdir(os.path.join(tmpdir, 'other1'))
+    other1.save(db_root)
+    audformat.testing.create_audio_files(other1, db_root, file_duration='0.1s')
 
     # database with new table
 
@@ -365,6 +371,9 @@ def test_update():
         audformat.define.IndexType.SEGMENTED,
         columns={'str': ('str', 'rater2')},
     )
+    db_root = audeer.mkdir(os.path.join(tmpdir, 'other2'))
+    other2.save(db_root)
+    audformat.testing.create_audio_files(other2, db_root, file_duration='0.1s')
 
     # raises error because schemes do not match
 
@@ -387,7 +396,7 @@ def test_update():
 
     with pytest.raises(ValueError):
         db.update(others, overwrite=False)
-    db.update(others, overwrite=True)
+    db.update(others, overwrite=True, copy_media=True)
 
     pd.testing.assert_frame_equal(db['table'].df, df)
     assert db['table_new'] == other2['table_new']
@@ -397,6 +406,33 @@ def test_update():
             assert db.raters[rater_id] == rater
         for scheme_id, scheme in other.schemes.items():
             assert db.schemes[scheme_id] == scheme
+
+    # don't copy files if self.root is not given
+
+    db_root = db.root
+    db._root = None
+    original_files = db.files
+    db.update(others, overwrite=True, copy_media=True)
+    db._root = db_root
+    for other in others:
+        for file in other.files:
+            if file in original_files:
+                assert os.path.exists(os.path.join(db.root, file))
+            else:
+                assert not os.path.exists(os.path.join(db.root, file))
+
+    # fail if one of the others has no root folder
+
+    db_root = other1._root
+    other1._root = None
+    with pytest.raises(RuntimeError):
+        db.update(others, overwrite=True, copy_media=True)
+    other1._root = db_root
+
+    # test media files
+
+    for file in db.files:
+        assert os.path.exists(os.path.join(db.root, file))
 
     # test other fields
 
@@ -420,7 +456,7 @@ def test_update():
         organization='other',
         source='other',
     )
-    db.update(other)
+    db.update(other, copy_media=True)
 
     assert db.author == f'{db_author}, {other.author}'
     assert db.name == db_name


audformat/core/database.py
```diff
@@ -558,6 +565,21 @@ def join_field(
                 else:
                     self[table_id] = table.copy()
 
+        # copy media files
+        if copy_media and self.root:
```
frankenjoe, line 569: I would check for `self.root is not None` here as you do later when you check for `other.root`.

audformat/core/database.py
```diff
@@ -558,6 +565,21 @@ def join_field(
...
+            for other in others:
+                if len(other.files) > 0 and other.root is None:
+                    raise RuntimeError(
+                        f"The database '{other.name}' has no root folder."
+                    )
+                for file in other.files:
+                    # Joining works also if the paths are already absolute
```
frankenjoe, line 576: Yes, but then `dst_file` points to `src_file` :)

audformat/core/database.py
```diff
@@ -558,6 +565,21 @@ def join_field(
...
                     self[table_id] = table.copy()
 
+        # copy media files
+        if copy_media and self.root:
+            for other in others:
+                if len(other.files) > 0 and other.root is None:
+                    raise RuntimeError(
+                        f"The database '{other.name}' has no root folder."
+                    )
+                for file in other.files:
```
frankenjoe, line 575: In principle we could use multi-threading here, the only nasty thing is that we have to create all target folders before we start the jobs to avoid the usual threading problem with `mkdir`.

audformat/core/database.py
hagenw: No, why should this happen.

The comment refers to:

```python
>>> os.path.join('/a', '/a/b')
'/a/b'
```

hagenw: `dst_file` and `src_file` should be different as `self.root` and `other.root` should be different.

audformat/core/database.py
hagenw: Yes, I was thinking about it before, but we also didn't had a `verbose` argument, so I thought I will skip it for now.
Or do you think we should incorporate it directly?

audformat/core/database.py
hagenw: Changed it accordingly.

audformat/core/database.py
frankenjoe: yes, but the comment suggests that paths can be absolute, but that is not true, they have to be relative.

audformat/core/database.py
frankenjoe: No, it's fine to skip it for now.

audformat/core/database.py
hagenw: No `/a/b` in my example is absolute and it works. If you load a database with `audb.load()` you can get absolute paths.

frankenjoe: Ok, I'm confused.

```
file = '/a/b/file.wav'
dst_file = os.path.join(self.root, file)  # ->  '/a/b/file.wav'
src_file = os.path.join(other.root, file) # -> '/a/b/file.wav'
```

No?

hagenw: Ah, ok, now I see the problem.

I have to fix it for `dst_file`, will have a look.

frankenjoe: Something like:

```python
if file.startswith(other.root):
    file = file.replace(other.root, '')
```

hagenw: I fixed the absolute path by removing it from `file` and added a test for it.

frankenjoe: But I guess there is still the problem that the new entries in the tables are still pointing to the old location?

hagenw: That's true and it gets even more complicated:

* if `self.db` stores the tables absolute, we have to replace the old absolute path with the new absolute path, or the add the new `db.root` path if other was relative
* if `self.db` stores the tables relative, we have to remove the absolute path

The other think is, we cannot make those changes before by applying it on `other` as this would change the original `other`  database which we don't want. So we first need to join and afterwards fix all the tables in `self`.

frankenjoe: Oh and it's even worse - with absolute paths, we get duplicated entries for every file that exists in both databases :) So maybe it is easier if we simply forbid absolute paths in `update()`?

frankenjoe: hehe, kind of same thought at the same time :)

hagenw: Yes, for a start it seems the easiest solution, I will adapt the code and forbid absolute path in `others` and `self`.

hagenw: OK, I'm struggling to get it to work.
I will pause here and continue on Monday.

hagenw: Just a remark on what is not working yet:

The test if the file is absolute in `other` happens to late at the moment, we have to fail before `copy_media=True` because the tables are joined already:

```
Index(['/tmp/pytest-of-hwierstorf/pytest-125/test_update0/other2/audio/001.wav',
       '/tmp/pytest-of-hwierstorf/pytest-125/test_update0/other2/audio/002.wav',
       '/tmp/pytest-of-hwierstorf/pytest-125/test_update0/other2/audio/003.wav',
       '/tmp/pytest-of-hwierstorf/pytest-125/test_update0/other2/audio/004.wav',
       '/tmp/pytest-of-hwierstorf/pytest-125/test_update0/other2/audio/005.wav',
       'audio/000.wav', 'audio/001.wav', 'audio/002.wav', 'audio/003.wav',
       'audio/004.wav', 'audio/005.wav'],
      dtype='object', name='file')
```

frankenjoe: By don't you move the test to the front of the function?

frankenjoe: And the test should of course be applied to `self` and `other`.

frankenjoe: And it should be applied independently of `copy_media`.

audformat/core/database.py
```diff
@@ -755,6 +791,16 @@ def load_header_from_yaml(header: dict) -> 'Database':
 
         return db
 
+    def _path_absolute(
+        self,
+    ) -> bool:
+        r"""Check if paths in tables are absolute."""
+        return (
```
frankenjoe, line 798: Maybe it's better to simply apply `os.path.isabs()` independent of `self.root`.

hagenw: I switched to `os.path.isabs()` and do not only check the first `files[0]` entry, but now check if any of `files` conatins an absolute path. This is slower, but I think as a user can add whatever paths she likes, there is no way around checking all entries.

audformat/core/database.py
hagenw: I moved the checks for absolute path now to the beginning of `update()` and it works.

hagenw:
I added now an extra public method to check if all file paths are given in a unique way (do not be absolute, or contain `.` or `..` for referring to a folder).

audformat/core/database.py
```diff
@@ -208,6 +209,34 @@ def segments(self) -> pd.MultiIndex:
         assert isinstance(index, pd.MultiIndex)
         return index.drop_duplicates()
 
+    def contains_nonunique_path(
+        self,
+    ) -> bool:
+        r"""Check if all paths are specified in a unique way.
+
+        If some path is given as an asolute path,
```
frankenjoe, line 217:
```suggestion
        If some path is given as an absolute path,
```

audformat/core/database.py
```diff
@@ -208,6 +209,34 @@ def segments(self) -> pd.MultiIndex:
         assert isinstance(index, pd.MultiIndex)
         return index.drop_duplicates()
 
+    def contains_nonunique_path(
```
frankenjoe, line 212: To be in-line with `db.drop_file()`, `db.files`, should we maybe call it `contains_nonunique_file()`. And would it also make sense to return `True` if all files are `unique` instead? Then we could simplify to  `contains_unique_files()`.

audformat/core/database.py
```diff
@@ -208,6 +209,34 @@ def segments(self) -> pd.MultiIndex:
...
+        self,
+    ) -> bool:
+        r"""Check if all paths are specified in a unique way.
```
frankenjoe, line 215: But isn't an absolute path also unique?

hagenw: No, as it can also be replaced by the relative path + `db.root`.

But I'm very open if you find a better way for the method.

audformat/core/database.py
hagenw: I would prefer to stay with the current behavior as it is faster (in almost all cases) to test if one file has a non-unique path than to check that all files have unique paths.

frankenjoe: But that's fine, just invert the return value :)

audformat/core/database.py
frankenjoe: Maybe `db.is_portable()`?

audformat/core/database.py
hagenw: Good point, I changed to `contains_unique_files()`

audformat/core/database.py
frankenjoe: But if you think `contains_unique_files()` is clear, then I'm also fine with it.

hagenw: OK, I changed to `is_portable()` and updated the description and title of this pull request.

audformat/core/database.py
```diff
@@ -254,6 +255,35 @@ def drop_tables(
         for table_id in table_ids:
             self.tables.pop(table_id)
 
+    def is_portable(
+        self,
+    ) -> bool:
+        r"""Check if a database can be moved to another loaction.
```
frankenjoe, line 261:
```suggestion
        r"""Check if a database can be moved to another location.
```

audformat/core/database.py
```diff
@@ -254,6 +255,35 @@ def drop_tables(
...
+
+        To be portable a database is supposed to not contain any
```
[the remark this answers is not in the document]
hagenw: Changed to
```rst
        To be portable,
        media may be referenced with an absolute path,
        or contain ``.`` or ``..`` to specify a folder.
```

audformat/core/database.py
```diff
@@ -465,6 +498,12 @@ def update(
                 same ID is found
             ValueError: if table data cannot be combined (e.g. values in
                 same position overlap)
+            RuntimeError: if ``copy_media=True``,
+                but one of the databases in ``others``
+                was not saved (contains files but no root folder)
+            RuntimeError: if any involved database is not portable
```
frankenjoe, line 504: Maybe simplify to just "if any involved database is not portable"?

audformat/core/database.py
```diff
@@ -558,6 +607,22 @@ def join_field(
...
+                    raise RuntimeError(
+                        f"You can only use update with saved databases. "
+                        f"The database '{other.name}' was not saved yet."
+                    )
+                for file in other.files:
+                    src_file = os.path.join(other.root, file)
+                    dst_file = os.path.join(self.root, file)
+                    dst_dir = os.path.dirname(dst_file)
+                    audeer.mkdir(dst_dir)
+                    shutil.copy(src_file, dst_file)
```
frankenjoe, line 624: Before we copy, should we check if `src_file` exists and raise a custom error message in that case?

audformat/core/database.py
```diff
@@ -558,6 +607,22 @@ def join_field(
                 else:
                     self[table_id] = table.copy()
 
+        # copy media files
+
+        if copy_media and self.root is not None:
```
frankenjoe, line 612: Maybe we should raise an error if `copy_media=True` but `self.root==None`?

audformat/core/database.py
frankenjoe: Ok, it's just that with the following "and" it sounds like `portable` and containing absolute paths etc. are two different things.

audformat/core/database.py
hagenw: You are right, that seems to be the safer option. I changed it accordingly.

audformat/core/database.py
hagenw: I think the error message `shutil.copy()` returns should be enough:

```
>>> shutil.copy('non-existing.txt', '.')
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-6-f7a24d9d4024> in <module>
----> 1 shutil.copy('non-existing.txt', '.')

~/.envs/audb/lib/python3.6/shutil.py in copy(src, dst, follow_symlinks)
    243     if os.path.isdir(dst):
    244         dst = os.path.join(dst, os.path.basename(src))
--> 245     copyfile(src, dst, follow_symlinks=follow_symlinks)
    246     copymode(src, dst, follow_symlinks=follow_symlinks)
    247     return dst

~/.envs/audb/lib/python3.6/shutil.py in copyfile(src, dst, follow_symlinks)
    118         os.symlink(os.readlink(src), dst)
    119     else:
--> 120         with open(src, 'rb') as fsrc:
    121             with open(dst, 'wb') as fdst:
    122                 copyfileobj(fsrc, fdst)

FileNotFoundError: [Errno 2] No such file or directory: 'non-existing.txt'
```

audformat/core/database.py
hagenw: Yes, the `and` makes it indeed hard. We could replace it by something `which means`, but I have no deleted it as it is also the safer option if we at some point change what `is_potable()` is checking the docstring here is still correct.

audformat/core/database.py
```diff
@@ -535,6 +572,16 @@ def join_field(
             assert_equal(other, 'license')
             assert_equal(other, 'usage')
 
+        # can only join databases with relatvie paths
+        for database in [self] + others:
+            if not database.is_portable():
+                raise RuntimeError(
+                    f"You can only use update with databases "
```
frankenjoe, line 579: I would either change to "You can only update with databases...." or "You can only use update() with databases..."

hagenw: I used "You can only update with... " already at another place, so I changed it to that one.

audformat/core/database.py
```diff
@@ -535,6 +572,15 @@ def join_field(
             assert_equal(other, 'license')
             assert_equal(other, 'usage')
 
+        # can only join databases with relatvie paths
+        for database in [self] + others:
+            if not database.is_portable():
+                raise RuntimeError(
+                    f"You can only update with databases that are portable "
```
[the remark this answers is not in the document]
hagenw: Yes, sorry, I did not really clean this up.
Changed now.

PR #75, opened 2021-05-07
Fix nullable dtype in utils.concat()

### LINKED ISSUE #74: Problem when concatenating non-nullable dtypes
`utils.concat()` does not properly work for dtypes that are not nullable:

E.g.:

```python
try:
    y1 = pd.Series(1, audformat.filewise_index('f1'))
    y2 = pd.Series(1, audformat.filewise_index('f2'))
    audformat.utils.concat([y1, y2])
except Exception as ex:
    print(ex)
```
```
Found overlapping data in column 'None':
      left  right
file             
f2       0      1
```

and:

```python
try:
    y1 = pd.Series(True, audformat.filewise_index('f1'))
    y2 = pd.Series(True, audformat.filewise_index('f2'))
    audformat.utils.concat([y1, y2])
except Exception as ex:
    print(ex)
```
```
Found overlapping data in column 'None':
       left  right
file              
f2    False   True
```

The problem is that we first create an empty table in which we then insert the data. However, before we insert the data we check for an overlap, which is a problem if the empty table does not contain NA values, but e.g. 0 in case of `int` or `False` in case of `bool`. The function then detects an overlap and raises an error.

The issue can be avoided by switching to a nullable dtype. E.g. the following works:

```python
y1 = pd.Series(1., audformat.filewise_index('f1'), dtype='Int64')
y2 = pd.Series(1., audformat.filewise_index('f2'), dtype='Int64')
audformat.utils.concat([y1, y2])

y1 = pd.Series(True, audformat.filewise_index('f1'), dtype='boolean')
y2 = pd.Series(True, audformat.filewise_index('f2'), dtype='boolean')
audformat.utils.concat([y1, y2])
```

frankenjoe:
Closes #74 

Solved by using nullable dtypes.

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -189,9 +189,15 @@ def concat(
             # drop NaN to avoid overwriting values from other column
             column = column.dropna()
         else:
+            if column.dtype.name.startswith('int'):
+                dtype = 'Int64'
+            elif column.dtype.name == 'bool':
+                dtype = 'boolean'
+            else:
+                dtype = column.dtype
             columns_reindex[column.name] = pd.Series(
                 index=index,
-                dtype=column.dtype,
+                dtype=dtype,
             )
         columns_reindex[column.name][column.index] = column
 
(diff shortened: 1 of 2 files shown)


hagenw:
Changes make sense, the only downside is that we are now changing the datatype.
E.g. if you join two series of type `int` the result will not be of type `int`.

frankenjoe:
>  if you join two series of type `int` the result will not be of type `int`.

yes, but that makes sense, because the result of `concat()` can contain `NaN`

PR #77, opened 2021-05-11
Remove root argument from testing.create_audio_files()

### LINKED ISSUE #72: audformat.testing.create_audio_files does not set db.root
Example:

```python
import audformat.testing

db.raters['rater'] = audformat.Rater()
db.schemes['str'] = audformat.Scheme(str)
audformat.testing.add_table(db, 'table', 'segmented', columns={'str': ('str', 'rater')})
audformat.testing.create_audio_files(db, './database', file_duration='0.1s')

db.root
```

results in

```
```

But it should return `audeer.safe_path('./database')` instead.

frankenjoe:
Closes #72 

`testing.create_audio_files(db, ...)` now automatically creates files in `db.root`. An error is raised if `db` was not saved or is not portable.

### DIFF
--- a/audformat/core/testing.py
+++ b/audformat/core/testing.py
@@ -8,11 +8,11 @@ from typing import (
     Tuple,
     Optional,
 )
+import warnings
 
 import numpy as np
 import pandas as pd
 
-import audeer
 import audiofile as af
 
 from audformat.core import define
@@ -150,6 +150,7 @@ def add_table(
 
 def create_audio_files(
         db: Database,
+        root: str = None,
         *,
         sample_generator: Callable[[float], float] = None,
         sampling_rate: int = 16000,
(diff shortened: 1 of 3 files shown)


audformat/core/testing.py
```diff
@@ -150,7 +150,6 @@ def add_table(
 
 def create_audio_files(
         db: Database,
-        root: str,
```
frankenjoe, line 153: This is of course a breaking change. The question is - should we officially deprecate it. But since it is a testing function it might be ok to simply remove it.

frankenjoe: Ok, it's now set to `None` by default and a warning is raised if it is set:

```
UserWarning: The argument 'root' is deprecated, 'db.root' will be used.'
```

hagenw:
Could you rebase the master as it the changes show that it will revert the `is_portable` stuff.

audformat/core/testing.py
hagenw: Great, your solution looks good.

PR #80, opened 2021-05-12
Add audformat.utils.join_schemes()

hagenw:
As we would need to implement this helper function in every data project, I thought it is better to add it directly here.

image

I also simplified the docstring of `audformat.utils.join_labels()` where we mentioned the code that is now provided by `audformat.utils.join_schemes()`.

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -364,9 +364,9 @@ def join_schemes(
         dbs: typing.Sequence[Database],
         scheme_id: str,
 ):
-    r"""Join and update scheme of two databases.
+    r"""Join and update scheme of databases.
 
-    This joins the given scheme of two databases
+    This joins the given scheme of several databases
     using :func:`audformat.utils.join_labels`
     and replaces the scheme in each database
     with the joined one.
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -361,8 +361,7 @@ def join_labels(
 
 
 def join_schemes(
-        db1: Database,
-        db2: Database,
+        dbs: typing.Sequence[Database],
         scheme: str,
 ):
     r"""Join and update scheme of two databases.
@@ -377,30 +376,25 @@ def join_schemes(
     with :meth:`audformat.Database.update`.
 
     Args:
-        db1: database
-        db2: database
-        scheme: scheme with labels contained in both databases
+        dbs: sequence of databases
+        scheme: scheme ID of a scheme with labels
+            that should be joined
 
     Example:
         >>> db1 = Database('db1')
         >>> db2 = Database('db2')
         >>> db1.schemes['scheme_id'] = Scheme(labels=['a'])
         >>> db2.schemes['scheme_id'] = Scheme(labels=['b'])
-        >>> join_schemes(db1, db2, 'scheme_id')
+        >>> join_schemes([db1, db2], 'scheme_id')
         >>> db1.schemes
         scheme_id:
           dtype: str
           labels: [a, b]
 
     """
-    labels = join_labels(
-        [
-            db1.schemes[scheme].labels,
-            db2.schemes[scheme].labels,
-        ]
-    )
-    db1.schemes[scheme].replace_labels(labels)
-    db2.schemes[scheme].replace_labels(labels)
+    labels = join_labels([db.schemes[scheme].labels for db in dbs])
+    for db in dbs:
+        db.schemes[scheme].replace_labels(labels)
 
 
 def map_language(language: str) -> str:
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -10,6 +10,7 @@ import audeer
 import audiofile
 
 from audformat.core import define
+from audformat.core.database import Database
 from audformat.core.index import index_type
 from audformat.core.index import (
     filewise_index,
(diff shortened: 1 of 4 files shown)


frankenjoe:
Oh I thought it was already a utility function. So yes, this change makes absolute sense.

hagenw:
> Oh I thought it was already a utility function. So yes, this change makes absolute sense.

No, before we focused on just joining the schemes without touching any database.

frankenjoe:
Ah sorry, here you actually join the scheme objects. I guess we need to mention that this can have side-effects on the tables.

audformat/core/utils.py
```diff
@@ -376,6 +360,49 @@ def join_labels(
     return joined_labels
 
 
+def join_schemes(
+        db1: Database,
```
frankenjoe, line 364: Since update supports a list of `Database` objects, shouldn't we also allow the same here?

frankenjoe: And would it make sense to actually use the new function in `Database.update()` then?

hagenw: The difference is that `update()` is already assigned to a database.
But in principle you are right, we could replace `db1` and `db2` by a single list `dbs` that can contain one or more databases.
Was that your idea?

frankenjoe: Ah sorry, forget my last comment. We decided earlier that it's safer to let the user do it because of the side-effects it can have.

frankenjoe: > Was that your idea?

yes

frankenjoe: Then it's also in line with `join_labels()`.

hagenw: I changed to a sequence of databases.

audformat/core/utils.py
```diff
@@ -376,6 +360,43 @@ def join_labels(
     return joined_labels
 
 
+def join_schemes(
+        dbs: typing.Sequence[Database],
+        scheme: str,
```
frankenjoe, line 365: Should we rename to `scheme_id` so that it is more obvious that not a `Scheme` object but an ID is expected?

hagenw: Good idea, I renamed it to `scheme_id`

audformat/core/utils.py
```diff
@@ -376,6 +360,43 @@ def join_labels(
     return joined_labels
 
 
+def join_schemes(
+        dbs: typing.Sequence[Database],
+        scheme_id: str,
+):
+    r"""Join and update scheme of two databases.
```
frankenjoe, line 367: Could you please update the docstring as we can now combine the scheme of more than two databases.

audformat/core/utils.py
```diff
@@ -376,6 +360,43 @@ def join_labels(
...
+        >>> db1.schemes['scheme_id'] = Scheme(labels=['a'])
+        >>> db2.schemes['scheme_id'] = Scheme(labels=['b'])
+        >>> join_schemes([db1, db2], 'scheme_id')
+        >>> db1.schemes
+        scheme_id:
+          dtype: str
+          labels: [a, b]
+
+    """
+    labels = join_labels([db.schemes[scheme_id].labels for db in dbs])
```
frankenjoe, line 395: maybe immediately return if `len(dbs) <= 1`?

audformat/core/utils.py
hagenw: done

audformat/core/utils.py
```diff
@@ -376,6 +360,43 @@ def join_labels(
...
+):
+    r"""Join and update scheme of databases.
+
+    This joins the given scheme of several databases
+    using :func:`audformat.utils.join_labels`
+    and replaces the scheme in each database
+    with the joined one.
+
+    This might be useful,
+    if you want to combine both databases
```
frankenjoe, line 375:
```suggestion
    if you want to combine databases
```

audformat/core/utils.py
hagenw: I don't think it will be any speed issue if the code gets executed, or are you worried that the database object is touched with `replace_labels`?

frankenjoe: Ok, as long as `join_labels` works with an empty list we can also leave it as it is.

frankenjoe:
> I guess we need to mention that this can have side-effects on the tables.

What do you think, should we maybe add a sentence similar to https://github.com/audeering/audformat/blob/a7ebb6c09815bf522c5a655337d17825443a4e93/audformat/core/scheme.py#L212

audformat/core/utils.py
hagenw: I added a test for it, it does work.

hagenw:
> > I guess we need to mention that this can have side-effects on the tables.
> 
> What do you think, should we maybe add a sentence similar to
> 
> https://github.com/audeering/audformat/blob/a7ebb6c09815bf522c5a655337d17825443a4e93/audformat/core/scheme.py#L212
> 
> ?

I changed the docstring to:

```rst
    This joins the given scheme of several databases
    using :func:`audformat.utils.join_labels`
    and replaces the scheme in each database
    with the joined one.
    The dtype of all :class:`audformat.Column` objects
    that reference the scheme in the databases
    will be updated.
    Removed labels are set to ``NaN``.

    This might be useful,
    if you want to combine databases
    with :meth:`audformat.Database.update`.
```

frankenjoe:
> I changed the docstring to:

Great, thanks! Looks all good now.

PR #84, opened 2021-06-23
Raise custom BadKeyError when key is not found

frankenjoe:
Introduces custom error class `BadKeyError`, which prints valid keys in the error message.

### Example

At the moment when a table is requested that does not exist, we get:

```python
import audformat.testing


db['bad-table']
```
```
Traceback (most recent call last):
  File "/media/jwagner/Data/Git/pyaudformat/my/my.py", line 16, in <module>
    db['bad']
  File "/media/jwagner/Data/Git/pyaudformat/audformat/core/database.py", line 652, in __getitem__
    return self.tables[table_id]
  File "/media/jwagner/Data/Git/pyaudformat/audformat/core/common.py", line 84, in __getitem__
    value = super().__getitem__(key)
KeyError: 'bad-table'
```

Now it prints the names of available tables in the error message:

```
...
  File "/media/jwagner/Data/Git/pyaudformat/audformat/core/common.py", line 81, in __getitem__
    raise BadKeyError(key, list(self))
audformat.core.errors.BadKeyError: "Bad key 'bad-table', expected one of ['files', 'segments']"
```

This also works in other places, e.g. with schemes:

```python
db.schemes['bad-scheme']
```
```
...
audformat.core.errors.BadKeyError: "Bad key 'bad-scheme', expected one of ['bool', 'date', 'float', 'int', 'label', 'label_map_int', 'label_map_str', 'string', 'time']"
```

### DIFF
--- a/audformat/core/errors.py
+++ b/audformat/core/errors.py
@@ -1,38 +1,6 @@
 import typing
 
 
-class BadValueError(ValueError):
-    """Raised when a value is not in a list of pre-defined strings.
-
-    Args:
-        invalid_value: value causing the error
-        valid_values: list of valid strings
-
-    """
-    def __init__(self, invalid_value: str, valid_values: typing.Sequence[str]):
-        message = (
-            f"Bad value '{invalid_value}', "
-            f"expected one of {list(valid_values)}"
-        )
-        super().__init__(message)
-
-
-class BadTypeError(ValueError):
-    r"""Raised when a value has an unexpected type.
-
-    Args:
-        invalid_value: value causing the error
-        expected_type: expected value type
-
-    """
-    def __init__(self, invalid_value: typing.Any, expected_type: type):
-        message = (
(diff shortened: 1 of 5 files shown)


audformat/core/errors.py
```diff
@@ -1,23 +1,46 @@
...
-    """Raised when a value is not in a list of pre-defined strings.
+class BadIdError(ValueError):
+    r"""Raised when a field identifier is unknown.
 
     Args:
-        invalid_value: value causing the error
-        valid_values: list of valid strings
+         name: name of the field
+         invalid_id: identifier causing the error
+         dictionary: dictionary with valid identifiers
```
hagenw, line 10:
```suggestion
        name: name of the field
        invalid_id: identifier causing the error
        dictionary: dictionary with valid identifiers
```

hagenw:
Sounds like a food idea to me.

PR #98, opened 2021-08-19
Speedup audformat.utils.union()

### LINKED ISSUE #97: audformat.utils.union() takes too long
This function is used in `audinterface` to combine segments detected by a `Segment` object, like VAD.
But it can take very long for a typical database, see https://github.com/audeering/audinterface/issues/26

To verify this I created 100 `MultiIndex` indexes that each contain 10.000 entries (1.000 files with 10 start and end times each).
I then benchmarked how long it takes to join a certain number of those index entries.

```python
import time

import numpy as np
import pandas as pd

import audformat


def chunks(lst, n):
    """Yield successive n-sized chunks from lst."""
    for i in range(0, len(lst), n):
        yield lst[i:i + n]


def measure(idx):
    start = time.time()
    i = audformat.utils.union(idx)
    end = time.time()
    return end - start


files = [f'a{n}' for n in range(1000)]
starts = [pd.to_timedelta(t, unit='s') for t in range(1000)]
ends = [pd.to_timedelta(n + 10 * np.random.rand()) for n in range(1000)]
# Split into lists of 10 entries each
starts = list(chunks(starts, 10))
ends = list(chunks(ends, 10))

# Create tuples
idx_tuples = []
for start, end in zip(starts, ends):
    idx_tuples.append([(f, s, e) for s, e in zip(start, end) for f in files])
idx = []
for idx_tuple in idx_tuples:
    idx.append(pd.MultiIndex.from_tuples(idx_tuple, names=['file', 'start', 'end']))
```

Then we get:

```python
>>> measure(idx[:10])
9.416870594024658
>>> measure(idx[:20])
32.61396503448486
>>> measure(idx[:30])
71.54374980926514
>>> measure(idx[:40])
128.81993174552917
>>> measure(idx[:50])
203.0398144721985
>>> measure(idx[:60])
274.86833333969116
```

First I test how fast an alternative method would work, that don't handles the union aspect:

```python
def measure(idx):
    start = time.time()
    idx = pd.concat([i.to_frame() for i in idx])
    i = pd.MultiIndex.from_frame(idx)
    end = time.time()
    return end - start
```

This returns:
```python
>>>  measure(idx[:10])
0.03077840805053711
>>> measure(idx[:20])
0.08491826057434082
>>> measure(idx[:30])
0.11141705513000488
>>> measure(idx[:40])
0.13216853141784668
>>> measure(idx[:50])
0.1430981159210205
>>> measure(idx[:60])
0.15899324417114258
```

We clearly need to improve on this. Hopefully we can find a solution that also handles the union part nicely.

hagenw:
Closes #97 

This avoids calling `index.union()` several times and replaces it by transforming each index to a dataframe and using `pd.concat` to combine them and remove duplicated index entries afterwards.

This decreases the execution times reported in #97 to:

```python
>>> measure(idx[:10])
0.043624162673950195
>>> measure(idx[:20])
0.09886717796325684
>>> measure(idx[:30])
0.137054443359375
>>> measure(idx[:40])
0.21222424507141113
>>> measure(idx[:50])
0.2929871082305908
>>> measure(idx[:60])
0.36236071586608887
```

Now we are similar to the 0.15899324417114258 for `idx[:60]` when not caring about duplicates and not sorting the results,
and a lot better than the 274.86833333969116 it took before.

There is a small change in the order of the returned arrays, as now we always put `np.NaT` before any number as this is the default behavior of `index.sortlevel()` which we cannot change.

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -874,24 +874,10 @@ def union(
 
     # Combine all MultiIndex entries and drop duplicates afterwards,
     # faster than using index.union(),
-    # compare https://github.com/audeering/audformat/issues/97
+    # compare https://github.com/audeering/audformat/pull/98
     df = pd.concat([o.to_frame() for o in objs])
     index = df.index
     index = index.drop_duplicates()
     index, _ = index.sortlevel()
-
-    if isinstance(index, pd.MultiIndex) and len(index.levels) == 3:
-        # asserts that start and end are of type 'timedelta64[ns]'
-        if index.empty:
-            index = segmented_index()
-        elif index.levels[2].empty:
-            # If all end values are NaT, pandas stores an empty array and
-            # since pd.Index.union() in that case sets the type to
-            # DatetimeArray, we need to set it back to 'timedelta64[ns]'.
-            index.set_levels(
-                [pd.to_timedelta([])],
-                level=[2],
-                inplace=True,
-            )
 
     return index
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -872,9 +872,13 @@ def union(
     if len(set(types)) != 1:
         objs = [to_segmented_index(obj) for obj in objs]
 
-    index = objs[0]
-    for obj in objs[1:]:
-        index = index.union(obj)
+    # Combine all MultiIndex entries and drop duplicates afterwards,
+    # faster than using index.union(),
+    # compare https://github.com/audeering/audformat/issues/97
+    df = pd.concat([o.to_frame() for o in objs])
+    index = df.index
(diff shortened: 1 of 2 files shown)


hagenw:
I tested without droping duplicates and sorting and the execution time is roughly the same, so this is not the remaining bottleneck.

hagenw:
If I remove the `if` statement at the end of the function that converts to timedelta we are going down with execution time to 0.28009748458862305 for `measure(idx[:60])`.
The good news is that we do no longer need the `if` statement as we no longer need to fix the data type of `end` values when not using `index.union()`.

PR #101, opened 2021-08-20
Support pandas>=1.3

### LINKED ISSUE #93: Tests fail for pandas>=1.3
I created a dummy pull request to check the current master: #92 

And it seems that something had broken our tests: https://github.com/audeering/audformat/pull/92/checks?check_run_id=3162446359

hagenw:
Closes #93.

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -140,14 +140,13 @@ def concat(
         # if we already have a column with that name, we have to merge them
         if column.name in columns_reindex:
 
+            dtype_1 = columns_reindex[column.name].dtype
+            dtype_2 = column.dtype
+
             # assert same dtype
-            if not same_dtype(
-                    columns_reindex[column.name].dtype, column.dtype
-            ):
-                dtype_1 = columns_reindex[column.name].dtype
+            if not same_dtype(dtype_1, dtype_2):
                 if dtype_1.name == 'category':
                     dtype_1 = repr(dtype_1)
-                dtype_2 = column.dtype
                 if dtype_2.name == 'category':
                     dtype_2 = repr(dtype_2)
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


hagenw:
You simply fixed it by pushing the latest `master` changes?

Maybe they fixed bugs in `pandas` in the mean time?
So the question would be if it would fail with some `pandas` versions, e.g. `pandas==1.3.0`

frankenjoe:
> So the question would be if it would fail with some `pandas` versions, e.g. `pandas==1.3.0`

I'm currently trying to figure out which versions are causing problems

frankenjoe:
```
1.1.5 OK
1.2.0 OK
1.2.5 OK
1.3.0 FAILED tests/test_database.py::test_update - ValueError: Cannot set a Categorical with another, without identical categories
1.3.1 FAILED tests/test_database.py::test_update - ValueError: Cannot set a Categorical with another, without identical categories
1.3.2 FAILED tests/test_database.py::test_update - ValueError: Cannot set a Categorical with another, without identical categories
1.3.3 FAILED tests/test_database.py::test_update - ValueError: Cannot set a Categorical with another, without identical categories
1.3.4 OK
```

So, I seems they introduces a change in `1.3.0` that they fixed in `1.3.4`.

I suggest we change to `pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3` then.

hagenw:
Cool, thanks for testing. So it seems indeed that they introduced bugs and not we ;)

The question is if you should also test if we can revert https://github.com/audeering/audformat/pull/101/commits/b7e87e9b3200be775234a187bcedd4e011f4c698 as well.

Otherwise I agree with you to simply disable the broken pandas versions and link to this merge request in `setup.cfg`.

frankenjoe:
> The question is if you should also test if we can revert [b7e87e9](https://github.com/audeering/audformat/commit/b7e87e9b3200be775234a187bcedd4e011f4c698) as well.

I tested it and it still fails with `1.3.4`.

Issue #103, opened 2021-10-13
Allow Database.load() to not load the tables into memory


hagenw:
For databases with lots of annotations it can happen that we have lots of CSV files and there size is altogether >5GB.
When loading such a database it can take a very long time (>30 minutes).

One solution would be to not load all the CSV files to the corresponding dataframes, but only when those are requested by `Column.get()`, `Table.get()` or by `Table.df`.

PR #106, opened 2021-11-01
Load table data on demand

frankenjoe:
Closes https://github.com/audeering/audformat/issues/103

Changes default value of `load_data` in `Database.load()` from `True` to `False`. 

* `load_data=True`, all table data is loaded immediately
* `load_data=False`, table data is loaded on demand when used for the first time, e.g. `get()` is called

image

Note that the previous behavior of `load_data=False` was to use an empty table and not load table data all.

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -702,10 +702,16 @@ class Database(HeaderBase):
         Args:
             root: root directory
             name: base name of header and table files
-            load_data: if ``True`` :class:`audformat.Table`
-                will load all table data immediately.
-                Otherwise,
-                table data will be loaded on demand
+            load_data: by default,
+                :class:`audformat.Table`
+                data is only loaded on demand,
+                e.g. when
+                :meth:`audformat.Table.get`
+                is called for the first time.
+                Set to ``True`` to load all
+                :class:`audformat.Table`
+                data immediately
+
             num_workers: number of parallel jobs.
                 If ``None`` will be set to the number of processors
                 on the machine multiplied by 5
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -702,7 +702,7 @@ class Database(HeaderBase):
         Args:
             root: root directory
             name: base name of header and table files
-            load_data: by default,
+            load_data: if ``False``,
                 :class:`audformat.Table`
                 data is only loaded on demand,
                 e.g. when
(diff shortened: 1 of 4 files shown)


hagenw:
I did a small test using `librispeech` and it really helps if you are interested in only one (or a few tables) tables:

```python
>>> timeit.timeit("db = audformat.Database.load('~/audb/librispeech/3.1.0/d3b62a9b', load_data=False); len(db['train-other-500'].df)", number=4, setup="import audformat")
3.2327790140006982
>>> timeit.timeit("db = audformat.Database.load('~/audb/librispeech/3.1.0/d3b62a9b', load_data=True); len(db['train-other-500'].df)", number=4, setup="import audformat")
8.110380816000543
```

If you are addressing all tables performance is more or less identical:

```python
>>> timeit.timeit("db = audformat.Database.load('~/audb/librispeech/3.1.0/d3b62a9b', load_data=False); len(db.files)", number=4, setup="import audformat")
14.637679298000876
>>> timeit.timeit("db = audformat.Database.load('~/audb/librispeech/3.1.0/d3b62a9b', load_data=True); len(db.files)", number=4, setup="import audformat")
14.66365661299642
```

So I think that it should be fine to change the default behavior of `audofrmat.Database.load()`.

Note, that calling `db.files` on a database with all table loaded takes already 5.726854896001896.

hagenw:
> Note that the previous behavior of `load_data=False` was to use an empty table and not load table data all.

Does this mean `table.df` returned an empty dataframe then, and now it would load the actual data frame?

frankenjoe:
> Does this mean `table.df` returned an empty dataframe then, and now it would load the actual data frame?

yes

frankenjoe:
So I think the new behavior is also safer, before we were assuming that a user would never use the tables when requested with `load_data=False`, but now it's fine to do so.

audformat/core/database.py
```diff
@@ -700,8 +702,16 @@ def load(
...
+            load_data: by default,
+                :class:`audformat.Table`
+                data is only loaded on demand,
+                e.g. when
+                :meth:`audformat.Table.get`
+                is called for the first time.
+                Set to ``True`` to load all
+                :class:`audformat.Table`
+                data immediately
+
```
hagenw, line 714:
```suggestion
            load_data: if ``False``,
                :class:`audformat.Table`
                data is only loaded on demand,
                e.g. when
                :meth:`audformat.Table.get`
                is called for the first time.
                Set to ``True`` to load all
                :class:`audformat.Table`
                data immediately
```

PR #107, opened 2021-11-02
Speed up creation of Database.files index

hagenw:
This contains two parts:

* When calling `audformat.Database.files` we used `pandas.Index.union` before with is slow as discussed in https://pandas.pydata.org/docs/reference/index.html#api. I replace it here with `audformat.utils.union()` instead
* It also has to call `audformat.Table.files` which was slow due to using `df.index.empty` which takes a long time for no reason. I replaced it by `len(df.index) == 0`

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -178,10 +178,10 @@ class Database(HeaderBase):
             files
 
         """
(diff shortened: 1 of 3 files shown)


hagenw:
BTW, the speed up was going from 8.4 s down to 0.1 s.

Issue #111, opened 2021-11-03
Speed up conversion to segmented table without NaT


frankenjoe:
When converting from a `filewise` to a `segmented` index we usually call `utils.to_segmented_index()` where we have the option to set the end of the segments to the file duration. This can be a bottleneck for large tables since the file duration has to be calculated for every file. However, usually we get our tables from a database that we load with [audb](https://github.com/audeering/audb) where the duration of every file is stored in the dependency table. So I wonder if we find a way to benefit from this information to speed up the conversion to segmented tables.

Possible solution might be that `audb` attaches a table with fie duration to the `Database` object it returns and we add a `segmented` option to `Table.get()` and `Column.get()`. If set to `True`, we return a segmented index where we can now access the file duration directly from the attached duration table.

frankenjoe:
As mentioned in #113 coding should be straight-forward, so it's basically a question which design we want to choose. I see two main solutions:

### Solution 1: private member

Add private member  `Database._duration`, which stores a table with duration per file and which is set by `audb`. To also make the information accessable to the user, we could add `Database.duration(file)`.

### Solution 2: column in db['files']

If it does not already exist, `audb` adds `db['files']['duration']`. Since a user will see the column, there is no need to extend the interface.

@hagenw which solution would you prefer or do you see another solution?

hagenw:
I would go for "Solution 1" as I don't think that the actual tables should be changed.

I also don't see a direct need for making it accessible to the user. As a first step I would update `audformat.utils.to_segmented_index(allow_nat=False)` that it first checks `Database._duration()` and only uses `audiofile.duration()` if information is missing. A user can always extend a database, so we can not rely only `Database._duration()`.

Above you proposed to maybe add a `segmented` argument to our `get()` methods. At the moment I'm not completely convinced by that. Having to use ``audformat.utils.to_segmented_index()` instead highlights more that this operation is not trivial and might take some time. But we can discuss about this also after we have implemented using the duration values from the dependency table.

hagenw:
There is one advantage I see in adding `Database.duration(file)` though. If it is called on a file without an entry in `Database._duration` it could add it there and afterwards it will be faster when called again, which is not the case for `audiofile.duration(os.path.join(db.root), file))`.

frankenjoe:
> As a first step I would update `audformat.utils.to_segmented_index(allow_nat=False)`

I don't see this work, the function does not know about `Database`.

PR #114, opened 2021-11-04
Add Database.files_duration()

frankenjoe:
As discussed in #111 this adds `Database.file_duration()`.

image

Uses `audiofile.duration()` to calculate the duration of a file and caches the result Internally in `Database._files_duration`. I decided to name it `files_duration()` to not confuse it with `utils.duration()` which returns the duration of the segments in a table.

I decided to use a dictionary instead of a `pd.Series`, as the latter is much slower as the following test shows:

```python
ts = [pd.to_timedelta(random.random(), unit='s') for _ in range(10000)]

d = {}
tt = time.time()
for idx, t in enumerate(ts):
    d[f'{idx}'] = t
time.time() - tt
```
```
0.0021936893463134766
```

```python
y = pd.Series()
tt = time.time()
for idx, t in enumerate(ts):
    y[f'{idx}'] = t
time.time() - tt
```
```
9.314390182495117
```

### Example

```python
db = audb.load(
    'emodb',
    version='1.1.1',
    format='wav',
    mixdown=True,
    sampling_rate=16000,
    only_metadata=True,
    full_path=False,
)
```

```python
db.files_duration(db.files).sum()
```
```
0 days 00:24:47.092187490
```

```python
db.files_duration(db['emotion'].files[:10])
```
```
file
wav/03a01Fa.wav      0 days 00:00:01.898250
wav/03a01Nc.wav      0 days 00:00:01.611250
wav/03a01Wa.wav   0 days 00:00:01.877812500
wav/03a02Fc.wav      0 days 00:00:02.006250
wav/03a02Nc.wav   0 days 00:00:01.439812500
wav/03a02Ta.wav   0 days 00:00:01.735687500
wav/03a02Wb.wav      0 days 00:00:02.123625
wav/03a02Wc.wav   0 days 00:00:01.498062500
wav/03a04Ad.wav   0 days 00:00:01.504874999
wav/03a04Fd.wav   0 days 00:00:01.696812500
Name: file, dtype: timedelta64[ns]
```

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -1,9 +1,11 @@
 import datetime
+import functools
 import itertools
 import os
 import shutil
 import typing
 
+import audiofile
 import oyaml as yaml
 import pandas as pd
 
@@ -11,7 +13,6 @@ import audeer
 
 from audformat.core import define
 from audformat.core import utils
-from audformat.core.index import segmented_index
 from audformat.core.column import Column
 from audformat.core.common import HeaderBase, HeaderDict
 from audformat.core.errors import BadIdError
@@ -166,6 +167,7 @@ class Database(HeaderBase):
         )
         r"""Dictionary of tables"""
 
+        self._file_duration = {}
         self._name = None
         self._root = None
 
@@ -288,6 +290,40 @@ class Database(HeaderBase):
             table_ids = [table_ids]
         for table_id in table_ids:
             self.tables.pop(table_id)
+
+    def file_duration(
+            self,
+            file: str = None,
+    ) -> pd.Timedelta:
+        r"""Duration of file or database.
+
+        Args:
+            file: if a file name is given,
+                returns duration of this file.
+                If ``None`` returns the summed duration
+                of all files in the database
+
+        Returns:
+            duration
+
+        """
+        if file is None:
+            return functools.reduce(
+                lambda d, f: d + self.file_duration(f),
+                self.files,
+                pd.to_timedelta(0),
+            )
+        elif file not in self._file_duration:
+            if not os.path.isabs(file) and self.root is not None:
+                path = os.path.join(self.root, file)
+            else:
+                path = file
+            dur = audiofile.duration(path)
+            dur = pd.to_timedelta(dur, unit='s')
+            self._file_duration[file] = dur
+            return dur
+        else:
+            return self._file_duration[file]
 
     def map_files(
             self,
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -303,9 +303,9 @@ class Database(HeaderBase):
         database is reloaded from disk.
 
         Use ``db.files_duration(db.files).sum()``
-        to get the total duration of a database.
-        Or ``db.files_duration(db['table'].files).sum()``
-        to get the total duration of a table.
+        to get the total duration of all files in a database.
+        Or ``db.files_duration(db[table_id].files).sum()``
+        to get the total duration of all files assigned to a table.
 
         Args:
             files: file names
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -293,8 +293,8 @@ class Database(HeaderBase):
     def files_duration(
             self,
             files: typing.Union[str, typing.Sequence[str]],
-    ) -> typing.Union[pd.Timedelta, pd.Series]:
-        r"""Duration of file(s) in the database.
+    ) -> pd.Series:
+        r"""Duration of files in the database.
 
         Note that durations are cached,
         i.e. changing the files on disk after calling
@@ -308,28 +308,25 @@ class Database(HeaderBase):
         to get the total duration of a table.
 
         Args:
-            files: for a single file name
-                returns the duration of the file.
-                For a sequence of file names
-                a mapping from file to duration
+            files: file names
 
         Returns:
-            duration(s)
+            mapping from file to duration
 
         """
-        if not isinstance(files, str):
-            return pd.Series(files, index=files).map(self.files_duration)
-        elif files not in self._files_duration:
-            if not os.path.isabs(files) and self.root is not None:
-                path = os.path.join(self.root, files)
-            else:
-                path = files
-            dur = audiofile.duration(path)
-            dur = pd.to_timedelta(dur, unit='s')
-            self._files_duration[files] = dur
-            return dur
-        else:
-            return self._files_duration[files]
+        def duration(file: str) -> pd.Timedelta:
+            if file not in self._files_duration:
+                if not os.path.isabs(file) and self.root is not None:
+                    path = os.path.join(self.root, file)
+                else:
+                    path = file
+                dur = audiofile.duration(path)
+                dur = pd.to_timedelta(dur, unit='s')
+                self._files_duration[file] = dur
+            return self._files_duration[file]
+
+        files = audeer.to_list(files)
+        return pd.Series(files, index=files).map(duration)
 
     def map_files(
             self,
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -298,16 +298,16 @@ class Database(HeaderBase):
     ) -> pd.Series:
         r"""Duration of files in the database.
 
-        Note that durations are cached,
-        i.e. changing the files on disk after calling
-        this function can lead to wrong results.
-        The cache is cleared when the
-        database is reloaded from disk.
-
         Use ``db.files_duration(db.files).sum()``
         to get the total duration of all files in a database.
         Or ``db.files_duration(db[table_id].files).sum()``
         to get the total duration of all files assigned to a table.
+
+        .. note:: Durations are cached,
+            i.e. changing the files on disk after calling
+            this function can lead to wrong results.
+            The cache is cleared when the
+            database is reloaded from disk.
 
         Args:
             files: file names
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -339,7 +339,7 @@ class Database(HeaderBase):
                 # expand file path
                 if root is None:
                     raise ValueError(
-                        "Found relative file name "
+                        f"Found relative file name "
                         f"{file}, "
                         f"but db.root is None. "
                         f"Please save database or "
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -166,7 +166,7 @@ class Database(HeaderBase):
         )
         r"""Dictionary of tables"""
 
-        self._file_duration = {}
+        self._files_duration = {}
         self._name = None
         self._root = None
 
@@ -319,17 +319,17 @@ class Database(HeaderBase):
         """
         if not isinstance(files, str):
             return pd.Series(files, index=files).map(self.files_duration)
-        elif files not in self._file_duration:
+        elif files not in self._files_duration:
             if not os.path.isabs(files) and self.root is not None:
                 path = os.path.join(self.root, files)
             else:
                 path = files
             dur = audiofile.duration(path)
             dur = pd.to_timedelta(dur, unit='s')
-            self._file_duration[files] = dur
+            self._files_duration[files] = dur
             return dur
         else:
-            return self._file_duration[files]
+            return self._files_duration[files]
 
     def map_files(
             self,
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -1,5 +1,4 @@
 import datetime
-import functools
 import itertools
 import os
 import shutil
@@ -291,39 +290,46 @@ class Database(HeaderBase):
         for table_id in table_ids:
             self.tables.pop(table_id)
 
-    def file_duration(
+    def files_duration(
             self,
-            file: str = None,
-    ) -> pd.Timedelta:
-        r"""Duration of a single file or all files in the database.
+            files: typing.Union[str, typing.Sequence[str]],
+    ) -> typing.Union[pd.Timedelta, pd.Series]:
+        r"""Duration of file(s) in the database.
+
+        Note that durations are cached,
+        i.e. changing the files on disk after calling
+        this function can lead to wrong results.
+        The cache is cleared when the
+        database is reloaded from disk.
+
+        Use ``db.files_duration(db.files).sum()``
+        to get the total duration of a database.
+        Or ``db.files_duration(db['table'].files).sum()``
+        to get the total duration of a table.
 
         Args:
-            file: if a file name is given,
-                returns duration of the file.
-                If ``None`` returns the summed duration
-                of all files in the database
+            files: for a single file name
+                returns the duration of the file.
+                For a sequence of file names
+                a mapping from file to duration
 
         Returns:
-            duration
+            duration(s)
 
         """
-        if file is None:
-            return functools.reduce(
-                lambda d, f: d + self.file_duration(f),
-                self.files,
-                pd.to_timedelta(0),
-            )
-        elif file not in self._file_duration:
-            if not os.path.isabs(file) and self.root is not None:
-                path = os.path.join(self.root, file)
+        if not isinstance(files, str):
+            return pd.Series(files, index=files).map(self.files_duration)
+        elif files not in self._file_duration:
+            if not os.path.isabs(files) and self.root is not None:
+                path = os.path.join(self.root, files)
             else:
-                path = file
+                path = files
             dur = audiofile.duration(path)
             dur = pd.to_timedelta(dur, unit='s')
-            self._file_duration[file] = dur
+            self._file_duration[files] = dur
             return dur
         else:
-            return self._file_duration[file]
+            return self._file_duration[files]
 
     def map_files(
             self,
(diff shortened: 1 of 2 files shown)


audformat/core/database.py
```diff
@@ -289,6 +291,40 @@ def drop_tables(
...
+                pd.to_timedelta(0),
+            )
+        elif file not in self._file_duration:
+            if not os.path.isabs(file) and self.root is not None:
+                path = os.path.join(self.root, file)
+            else:
+                path = file
+            dur = audiofile.duration(path)
+            dur = pd.to_timedelta(dur, unit='s')
+            self._file_duration[file] = dur
```
hagenw, line 323: Above you allow that `file` is an absolute path, which is ok for running `audiofile` on it. But when we then store it allowing also relative or absolute path, it means that it `self._file_duration` can get twice as big as `db.files` in the end (or even bigger if the user moves around the database in between).

tests/test_database.py
```diff
@@ -163,6 +164,28 @@ def test_drop_and_pick_tables():
...
+
+    db = pytest.DB
+
+    # duration of single file
+
+    file = db.files[0]
+    full_file = os.path.join(db.root, file)
+    dur = pd.to_timedelta(audiofile.duration(full_file), unit='s')
+    assert db.file_duration(file) == dur
+    assert db.file_duration(full_file) == dur
```
hagenw, line 177: I think you should repeat the last two lines in order to check that a second call still returns the same duration.

audformat/core/database.py
hagenw: One way around this would be that we allow only the syntax that is used in the current index (e.g. `db.files`).
But again, `db.files` is generated on the fly and it will change if a user updates a table in between.

hagenw:
I agree that it might be dangerous to name it `Database.duration()`. The problem with the current solution of `Database.file_duration()` is that I would have never guessed that I can use it to get the overall duration. Instead I would have tried the following:
```python
duration = np.sum([db.file_duration(file) for file in db.files])
```
I don't have an obvious solution for the problem. We could use `Database.files_duration()`, but this does also not look very nice.

audformat/core/database.py
frankenjoe: I guess whether `file` is absolute or relative depends mainly on how the files are stored in the table. We could of course always make it relative by removing `db.root`, but I'm not sure if this really worth the effort. In the tables we also usually store absolute paths.

frankenjoe: I also don't think moving around a database is a common use case, is it?

frankenjoe:
> The problem with the current solution of `Database.file_duration()` is that I would have never guessed that I can use it to get the overall duration.

We could also make `file` a mandatory argument and add a property `Database.total_file_duration` instead.

audformat/core/database.py
frankenjoe: On the other hand - always storing the relative path would indeed solve the problem :)

tests/test_database.py
frankenjoe: done

hagenw:
How do you solve the problem of an updated file, that has already a duration entry?
I guess at the moment the idea is that this is only correctly handled in `audb`, but if a user modifies a local database we don't check for this.

frankenjoe:
>  The only alternative to make it safe inside `audformat` would be to always call `audiofile` as the underlying file could always be changed/updated by the user

Yes, but then we don't gain anything. The ultimate goal is that we make the conversion from filewise to segmented index without `NaT` faster.

> We could of cause also argue that this is not the way you should use audformat and simply document it?

Yes, if a user changes a file under the hood, the labels will also no longer fit. So this is of course a general problem.

frankenjoe:
> We could use `Database.files_duration()`, but this does also not look very nice.

I now renamed the function to `Database.files_duration()`. It now also accepts a sequence of files and in that case returns a mapping from file to duration. A user can now use `db.files_duration(db.files).sum()` to get the total duration of a database.

frankenjoe:
I updated the description and added an example there. Would be nice to add those to the docstring directly. So maybe we should think about loading a database with `audb` when we build the documentation.

hagenw:
> The ultimate goal is that we make the conversion from filewise to segmented index without setting end=NaT faster.

How can we achieve this for `audformat.utils.to_segmented_index()` as this has no access to the `Database` object?
The same problem we phase for `audformat.utils.duration()`.

frankenjoe:
> How can we achieve this for `audformat.utils.to_segmented_index()` as this has no access to the `Database` object?
> The same problem we phase for `audformat.utils.duration()`.

Yes, that's why I suggested to add a `as_segmented` option to `Table.get()` and `Column.get()`. I.e. the user can replace:

```python
df = db['table'].get()
df = utils.to_segmented_index(df, allow_nat=False)
```

with:

```
df = db['table'].get(as_segmented=True, allow_nat=False)
```

audformat/core/database.py
```diff
@@ -289,6 +290,47 @@ def drop_tables(
...
+        Note that durations are cached,
+        i.e. changing the files on disk after calling
+        this function can lead to wrong results.
+        The cache is cleared when the
+        database is reloaded from disk.
+
+        Use ``db.files_duration(db.files).sum()``
+        to get the total duration of a database.
+        Or ``db.files_duration(db['table'].files).sum()``
+        to get the total duration of a table.
```
hagenw, line 308: It's not so nice that we now have two functions that can do the same thing. E.g. `db.files_duration(db.files).sum()`
and `audformat.utils.duration(db.files)`. As I don't have to add the `.sum()` at the end, I think I would pick the later as a user as I will not be aware that it cannot use the caching.

audformat/core/database.py
```diff
@@ -289,6 +290,47 @@ def drop_tables(
         for table_id in table_ids:
             self.tables.pop(table_id)
 
+    def files_duration(
+            self,
+            files: typing.Union[str, typing.Sequence[str]],
+    ) -> typing.Union[pd.Timedelta, pd.Series]:
```
hagenw, line 296: I think it is not a good idea to return different things for a single file or several files. I would change to always return a series.

audformat/core/database.py
frankenjoe: The advantage I see is that you can easily apply other operations, e.g. `db.files_duration(db.files).max()` or  `db.files_duration(db.files).mean()`.

hagenw:
> > How can we achieve this for `audformat.utils.to_segmented_index()` as this has no access to the `Database` object?
> > The same problem we phase for `audformat.utils.duration()`.
> 
> Yes, that's why I suggested to add a `as_segmented` option to `Table.get()` and `Column.get()`. I.e. the user can replace:
> 
> ```python
> df = db['table'].get()
> df = utils.to_segmented_index(df, allow_nat=False)
> ```
> 
> with:
> 
> ```python
> df = db['table'].get(as_segmented=True, allow_nat=False)
> ```

Ah, yes, found it at https://github.com/audeering/audformat/issues/111#issue-1043261735 thanks.
This seems to be a possible solution.

We then just need to train the users when to use `audformat.utils.to_segmented_index()` and when not ;)

audformat/core/database.py
hagenw: Yes, I also see the advantage. I'm just worried that most of the users will simply continue using `audformat.utils.duration()`, but I think we simply have to life with it. It's also still great to have it as it provides the possibility to calculate duration for data frames that are not part of any database.

hagenw: I think it is not obvious what "the total duration of a table" means for a segmented table.
Maybe we should change to
```suggestion
        to get the total duration of all files assigned to a table.
```

audformat/core/database.py
```diff
@@ -289,6 +290,47 @@ def drop_tables(
...
+
+        Note that durations are cached,
+        i.e. changing the files on disk after calling
+        this function can lead to wrong results.
+        The cache is cleared when the
+        database is reloaded from disk.
+
+        Use ``db.files_duration(db.files).sum()``
+        to get the total duration of a database.
+        Or ``db.files_duration(db['table'].files).sum()``
```
hagenw, line 307: I would change it to
```suggestion
        Or ``db.files_duration(db[table_id].files).sum()``
```

audformat/core/database.py
frankenjoe: Changed to:

image

audformat/core/database.py
frankenjoe: Changed it accordingly.

audformat/core/database.py
frankenjoe: Changed it accordingly.

hagenw:
The example you provide above fails for me, because due to `only_metadata=True` there are no files:

```python
db = audb.load(
    'emodb',
    version='1.1.1',
    format='wav',
    mixdown=True,
    sampling_rate=16000,
    only_metadata=True,
    full_path=False,
)
db.files_duration(db['emotion'].files[:10])
```
results in
```
In [9]: db.files_duration(db['emotion'].files[:10])                                        
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-a1e8b2b5ac20> in <module>
----> 1 db.files_duration(db['emotion'].files[:10])
                                                  
~/git/audeering/audformat/audformat/core/database.py in files_duration(self, files)
    327                                                                                                                                         
    328         files = audeer.to_list(files)                                     
--> 329         return pd.Series(files, index=files).map(duration)       
    330                                                         
    331     def map_files(                                        
                                                                            
~/.envs/audformat/lib/python3.6/site-packages/pandas/core/series.py in map(self, arg, na_action)
   3981         dtype: object                                                                     
   3982         """                                                         
-> 3983         new_values = super()._map_values(arg, na_action=na_action)
   3984         return self._constructor(new_values, index=self.index).__finalize__(
   3985             self, method="map"        
                                                                               
~/.envs/audformat/lib/python3.6/site-packages/pandas/core/base.py in _map_values(self, mapper, na_action)
   1158                                                                                
   1159         # mapper is a function
-> 1160         new_values = map_f(values, mapper) 
   1161                                                                                     
   1162         return new_values
       
pandas/_libs/lib.pyx in pandas._libs.lib.map_infer()
                                                                                                                       
~/git/audeering/audformat/audformat/core/database.py in duration(file)
    321                 else:
    322                     path = file
--> 323                 dur = audiofile.duration(path)
    324                 dur = pd.to_timedelta(dur, unit='s')
    325                 self._files_duration[file] = dur
                                 
...

RuntimeError: Error opening '/home/audeering.local/hwierstorf/audb/emodb/1.1.1/fe182b91/wav/03a01Nc.wav': System error.
```

I guess we need to handle this case. The question is if we should return an empty Series then, or a Series containing all the files, but no durations or 0 as duration?

audformat/core/database.py
```diff
@@ -289,6 +290,44 @@ def drop_tables(
...
+            mapping from file to duration
+
+        """
+        def duration(file: str) -> pd.Timedelta:
+            if file not in self._files_duration:
+                if not os.path.isabs(file) and self.root is not None:
+                    path = os.path.join(self.root, file)
+                else:
+                    path = file
+                dur = audiofile.duration(path)
```
hagenw, line 323: I think we should add another check here.

Let us assume that all paths are stored absolute in `db._files_duration` because the database was loaded with `full_path=True` and then we called `db.files_duration(db.files)`. If I then call `db.files_duration(relative_path_to_file)` it will generate the correct path with
```python
path = os.path.join(self.root, file)
```
but after that we should not directly call `audiofile.duration(path)` but could check again if `path` is inside `self._files_duration` and use that value instead.

audformat/core/database.py
hagenw: To not have to deal with https://github.com/audeering/audformat/pull/114#discussion_r744586885 maybe we indeed only store the relative path of the files. 

> I also don't think moving around a database is a common use case, is it?

It is if you do `db.update(db_new)`, where `db_new` is moved to the new location. Will it in general work if both `db` and `db_new` contain already entries in `db._files_duration`?

frankenjoe:
> I guess we need to handle this case. The question is if we should return an empty Series then, or a Series containing all the files, but no durations or 0 as duration?

Doesn't it make sense to raise an error if files do not exist? If we silently ignore it, the user may not notice it and work with wrong results.

audformat/core/database.py
frankenjoe: But `db.update()` just add some new files, I don't see why this is a problem.

hagenw:
> > I guess we need to handle this case. The question is if we should return an empty Series then, or a Series containing all the files, but no durations or 0 as duration?
> 
> Doesn't it make sense to raise an error if files do not exist? If we silently ignore it, the user may not notice it and work with wrong results.

It would be ok to raise an error, but it will have consequences. E.g. we should then not set `allow_nat=False` in `get()`, because users will work a lot with `only_metadata=True` and still might want to convert to a segmented table. BTW, for some tables it will even work if only the metadata is downloaded as `audb` will fill in the `db._files_duration` table, but this will not be the case for all databases.

audformat/core/database.py
hagenw: The question was more what will happen with its `_files_duration` entries. Will they be merged with the ones from `db`? If yes, it will only make sense if the path is stored relative as the absolute path would now be different.

frankenjoe: Mhh, yes if we store the file paths relative we could indeed update `_files_duration`. Otherwise, it would simply re-calculate the duration for the new files when `files_duration()` is called.

At the moment I'm not completely convinced if we should really store the relative path, though. Cause then there is of course the question what `files_duration()` should return. Always the relative path? Always the full path? At the moment it will return it the same way it's stored in the tables. And I don't think that we will often have the case that the user requests a database with absolute path, but then calls `files_duration()` or vice versa. At least as long as the user passes `db.files` or `db[table_id].files` this is not the case. And even if the user does it, the result will still be correct, though it has re-calculate the duration in that case. But as I said, I don't see it as a common use case.

hagenw: I'm also fine with your proposal. To summarize it:

* `db.update(db_new)` will simply ignore the entries in `db_new._files_duration`
* in `db._files_duration` we store relative and absolute paths as different entries. This is not a problem as the user should not access this dictionary
* we could simply recalculate the duration if a user has called first absolute path and then relative (or vice versa), but I think we could even avoid it, see https://github.com/audeering/audformat/pull/114#discussion_r744586885

frankenjoe:
> E.g. we should then not set `allow_nat=False` as default in `get()`

Yes, in `utils.to_segmented_table()` we also have `allow_nat=True` as the default value.

frankenjoe:
>  BTW, for some tables it will even work if only the metadata is downloaded as `audb` will fill in the `db._files_duration` table, but this will not be the case for all databases.

True, for all tables loaded with `audb` it will work even if the files were not downloaded!

frankenjoe:
> No, not for all, we still have databases published that don't have duration stored in the dependency table:

Ah yeah, right. I guess it applies to databases we have directly converted from the old `audb` repository. But for databases published with `audb.publish()` it should work, correct?

hagenw:
> > No, not for all, we still have databases published that don't have duration stored in the dependency table:
> 
> Ah yeah, right. I guess it applies to databases we have directly converted from the old `audb` repository. But for databases published with `audb.publish()` it should work, correct?

Yes, at some point we decided to switch to use `audiofile.duration(file, sloppy=True)` and that the duration should also be stored for MP3 and MP4 files when running `audb.publish()`. At the beginning we set duration to `0.0` for MP3 and MP4 files. But in general, I don't think this is something we have to worry about here.

audformat/core/database.py
hagenw: If I see it correctly the first two points are already implemented by the current behavior, correct?

frankenjoe: Yes, and with the latest commit the third point should be implemented, too. See my comment at #114.

audformat/core/database.py
frankenjoe: I added a check as you suggested. In addition I applied two more changes:

1. I set the name of the returned `pd.Series` to `file`
2. I added an argument `root` to cover the case where file names are relative but the database was not saved yet or loaded from disk (e.g. when a user prepares a database for publishing). In this case the user has to provide `root` or otherwise an error is raised.

audformat/core/database.py
```diff
@@ -289,6 +290,82 @@ def drop_tables(
...
+            i.e. changing the files on disk after calling
+            this function can lead to wrong results.
+            The cache is cleared when the
+            database is reloaded from disk.
+
+        Args:
+            files: file names
+            root: root directory under which the files are stored.
+                Provide if file names are relative and
+                database was not saved or loaded from disk
```
hagenw, line 316: Maybe we should add here, what the default behavior is, e.g.

```suggestion
                database was not saved or loaded from disk.
                If ``None`` :attr:`audformat.Database.root` is used
```

frankenjoe: Good idea.

audformat/core/database.py
hagenw: Great, thanks.

PR #115, opened 2021-11-09
Database.files_duration: always store full path

frankenjoe:
In #114 we had discussed if it makes sense to store duration with the relative path. But the more obvious solution is to always store the full path. As the following examples show, we can now request durations with a relative and absolute path and independent of the `full_path` argument we never get duplicated entries in the dictionary:

```python
db = audb.load(
    'emodb',
    version='1.1.1',
    format='wav',
    mixdown=True,
    sampling_rate=16000,
    only_metadata=True,
    full_path=False,
)

file = db.files[0]
db.files_duration(file)

full_file = os.path.join(db.root, file)
db.files_duration(full_file)

db._files_duration
```
```
{'/media/jwagner/Data/audb/emodb/1.1.1/fe182b91/wav/03a01Fa.wav': Timedelta('0 days 00:00:01.898250')}
```

```python
db = audb.load(
    'emodb',
    version='1.1.1',
    format='wav',
    mixdown=True,
    sampling_rate=16000,
    only_metadata=True,
    full_path=True,
)

file = full_file[len(db.root) + 1:]
db.files_duration(file)

full_file = os.path.join(db.root, file)
db.files_duration(full_file)

db._files_duration
```
```
{'/media/jwagner/Data/audb/emodb/1.1.1/fe182b91/wav/03a01Fa.wav': Timedelta('0 days 00:00:01.898250')}
```

With the current master you get in both cases:

```python
{'wav/03a01Fa.wav': Timedelta('0 days 00:00:01.898250'), '/media/jwagner/Data/audb/emodb/1.1.1/fe182b91/wav/03a01Fa.wav': Timedelta('0 days 00:00:01.898250')}
```

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -330,14 +330,10 @@ class Database(HeaderBase):
 
         def duration(file: str) -> pd.Timedelta:
 
-            # check cache
-            if file in self._files_duration:
-                return self._files_duration[file]
-
+            # expand file path
             if os.path.isabs(file):
                 full_file = file
             else:
-                # expand file path
                 if root is None:
                     raise ValueError(
                         f"Found relative file name "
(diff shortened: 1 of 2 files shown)


hagenw:
Great, I just checked that the user still sees the relative path:

```python
>>> file
'wav/03a01Fa.wav'
>>> db.files_duration(file)
wav/03a01Fa.wav   0 days 00:00:01.898250
Name: file, dtype: timedelta64[ns]
>>> db._files_duration
{'/home/audeering.local/hwierstorf/audb/emodb/1.1.1/fe182b91/wav/03a01Fa.wav': Timedelta('0 days 00:00:01.898250')}
```

PR #116, opened 2021-11-09
Get table or column data with a segmented index

### LINKED ISSUE #111: Speed up conversion to segmented table without NaT

frankenjoe:
Closes #111 

### Column

image

### Table

image

To avoid re-implementing a lot stuff we already have, the methods call `utils.to_segmented_index()`, which now has a new argument `files_duration`:

image

### DIFF
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -154,7 +154,13 @@ class Column(HeaderBase):
                 assigned to a scheme that contains a dict mapping
                 speaker IDs to age entries, ``map='age'``
                 will replace the ID values with the age of the speaker
-            as_segmented: always return a segmented index
+            as_segmented: if set to ``True``
+                and column has a filewise index,
+                the index of the returned column
+                will be converted to a segmented index.
+                ``start`` will be set to ``0`` and
+                ``end`` to ``NaT`` or to the file duration
+                if ``allow_nat`` is set to ``False``
             allow_nat: if set to ``False``,
                 ``end=NaT`` is replaced with file duration
             root: root directory under which the files are stored.
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -171,6 +171,7 @@ class Column(HeaderBase):
             labels
 
         Raises:
+            FileNotFoundError: if file is not found
             RuntimeError: if column is not assigned to a table
             ValueError: if trying to map without a scheme
             ValueError: if trying to map from a scheme that has no labels
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -131,6 +131,10 @@ class Column(HeaderBase):
             *,
             map: str = None,
             copy: bool = True,
+            as_segmented: bool = False,
+            allow_nat: bool = True,
+            num_workers: typing.Optional[int] = 1,
+            verbose: bool = False,
     ) -> pd.Series:
         r"""Get labels.
 
@@ -149,6 +153,13 @@ class Column(HeaderBase):
                 assigned to a scheme that contains a dict mapping
                 speaker IDs to age entries, ``map='age'``
                 will replace the ID values with the age of the speaker
+            as_segmented: always return with a ``segmented`` index
+            allow_nat: if set to ``False``,
+                ``end=NaT`` is replaced with file duration
+            num_workers: number of parallel jobs.
+                If ``None`` will be set to the number of processors
+                on the machine multiplied by 5
+            verbose: show progress bar
 
         Returns:
             labels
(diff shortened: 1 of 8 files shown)


audformat/core/column.py
```diff
@@ -149,11 +154,24 @@ def get(
                 assigned to a scheme that contains a dict mapping
                 speaker IDs to age entries, ``map='age'``
                 will replace the ID values with the age of the speaker
+            as_segmented: always return a segmented index
```
hagenw, line 157: I think we need to be more specific here (and in the table documentation at the same place) what will happen. Maybe:

```suggestion
            as_segmented: always return a segmented index.
                ``start`` will be set to ``0``.
                ``end`` to ``NaT`` or to the duration of the file
                if ``allow_nat`` is set to ``False``
```

frankenjoe: What about:

image

audformat/core/table.py
```diff
@@ -545,6 +569,24 @@ def get(
                 if column not in mapped_columns:
                     result.drop(columns=column, inplace=True)
 
+        # if necessary, convert to segmented index and replace NaT
+        is_segmented = index_type(result.index) == define.IndexType.SEGMENTED
+        if (not is_segmented and as_segmented)\
+                or (is_segmented and not allow_nat):
+            files_duration = None
```
hagenw, line 576:
```suggestion
        if (
                (not is_segmented and as_segmented)
                or (is_segmented and not allow_nat)
        ):
            files_duration = None
```

tests/test_column.py
```diff
@@ -115,6 +115,44 @@ def test_filewise(num_files, values):
         column.set([], index=index)
 
 
+def test_get_as_segmented():
+
+    db = pytest.DB
+    db._files_duration = {}
```
hagenw, line 121: Why do you have to set this, shouldn't there be a default value?

hagenw: The same comment holds for the other places in the tests where this is explicitly set.

audformat/core/column.py
hagenw: Yes, perfect.

hagenw: Maybe instead of `labels` we use `table` and `column`?

tests/test_column.py
frankenjoe: We don't have to set it, I just wanted to make sure it's used. Should I add a separate test instead?

frankenjoe: Ah wait a moment. I misunderstood your question. We have to reset it since `pytest.DB` is shared by several tests and we always want to start from an empty dictionary.

hagenw: To make sure I get this correctly: it's default value is `None` and then it will not be used inside the tests. You force it to be used by setting it to `{}`.

Maybe it will be easier to understand if we make it an argument to the test or add a separate test.
I would test for:
* not setting it
* setting it to `{}`
* setting a different file duration then the actual one for a given file and test if this was used afterwards

audformat/core/column.py
frankenjoe: I now changed to:

image

and

image

tests/test_column.py
frankenjoe: The default value is `{}`.

frankenjoe: But since we share `pytest.DB` we have to reset it manually to start from an empty dictionary again.

hagenw: Ah, ok, then please add a comment why you are resetting.

On top of that I think it would still be good to add a test with `db._files_duration` containing an entry and to check that this entry is used when setting `end`

frankenjoe: > Ah, ok, then please add a comment why you are resetting.

I will remove the line, since we already reset at the end of each function and there is already a comment

> On top of that I think it would still be good to add a test with db._files_duration containing an entry and to check that this entry is used when setting end

done, I also check now that non-existing entries are added

hagenw: Very cool, thanks.

PR #117, opened 2021-11-11
Use yaml.CLoader for faster header reading

hagenw:
Try to speed up YAML header file loading by using the `CLoader`.

I use a `try`-`except` statement to handle missing `CLoader`.

In addition, I added a `pyyaml >=5.4.1` requirement as locally this version was providing the `CLoader` for me.
I could also try lower versions, but I don't think there is a need to support lower versions.

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -805,7 +805,7 @@ class Database(HeaderBase):
 
         with open(path, 'r') as fp:
 
-            header = yaml.load(fp, Loader=yaml.Loader)
+            header = yaml.load(fp, Loader=yaml.CLoader)
             db = Database.load_header_from_yaml(header)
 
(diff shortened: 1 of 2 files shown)


hagenw:
According to https://pyyaml.org/wiki/PyYAMLDocumentation there is a way to deal with this:

```python
try:
    from yaml import CLoader as Loader, CDumper as Dumper
except ImportError:
    from yaml import Loader, Dumper
```
but I'm wondering why in some of my local environments it fails to find it and in others not.

hagenw:
For a measurement of speed increase have a look at https://github.com/audeering/audb/issues/127#issuecomment-966338947

frankenjoe:
Cool, another speed up :)

PR #118, opened 2021-11-11
WIP: Remove oyaml dependency

hagenw:
Seems by just replacing `oyaml` with `pyyaml` we get a lot of failing tests.

### DIFF
--- a/audformat/core/common.py
+++ b/audformat/core/common.py
@@ -1,5 +1,5 @@
 import inspect
-import oyaml as yaml
+import yaml
 from typing import Sequence, Callable
 import textwrap
 from collections import OrderedDict
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -5,7 +5,7 @@ import shutil
 import typing
 
 import audiofile
-import oyaml as yaml
+import yaml
 import pandas as pd
 
 import audeer
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ install_requires =
     audeer >=1.8.0,<2.0.0
     audiofile >=0.4.0
     iso-639
-    oyaml
+    pyyaml >=3.4.1
     # https://github.com/audeering/audformat/pull/101
     pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3
 setup_requires =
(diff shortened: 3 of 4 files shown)


hagenw:
`pyyaml` is always returning something like 

```
     scheme_id:                                                                                                                                                                                
    -  dtype: str                                                                                                                                                                              
    -  labels: [a, b]                                                                                                                                                                          
    +  !!python/object/apply:collections.OrderedDict 
```

so I guess we have to update our string representations?

frankenjoe:
> so I guess we have to update our string representations?

I think what's happening is that by `pyyaml` does not properly serialize `collections.OrderedDict` but only `dict`. Whereas `oyaml`, which internally builds on `collections.OrderedDict` for ordered dictionaries naturally can cope with it. So I guess, if we find that `pyyaml` does preserve ordering, we have to replace  `collections.OrderedDict` with `dict`.

frankenjoe:
But then we have to see if deriving from `dict` will actually work as expected, there seems to be cases where it is ok, but not always, see https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/

hagenw:
Why do we have to derive from `dict` and cannot simply use `dict`?

frankenjoe:
> Why do we have to derive from `dict` and cannot simply use `dict`?

Because of the `HeaderDict` class: https://github.com/audeering/audformat/blob/9613efcccf0c45fb78432c237d784067ccc03b17/audformat/core/common.py#L16

hagenw:
I see one way around would be to convert to `dict` before dumping and after reading from the YAML file, but I'm not sure if this is worse the effort.
Maybe as long as `oyaml` works with the newest versions of `pyyaml` we simply stay with the current solution?

frankenjoe:
I did  a quick check with the latest version of `pyyaml` (6.0) and turns out it anyway does not preserve the order of dictionaries:

```python
>>> import yaml
>>> d = {'this': 0, 'is': 1, 'an': 2, 'awesome': 3, 'dict': 4} 
>>> print(yaml.dump(d))
an: 2
awesome: 3
dict: 4
is: 1
this: 0
```

So we anyway have to stick to `oyaml`.

hagenw:
Mh, ok. I guess they support the order, but they seem to order it themselves before storing.

I will simply close this and we hope `oyaml` will work the next few years, otherwise we need to take it over ;)

PR #119, opened 2021-11-16
Add audformat.utils.expand_file_path()

frankenjoe:
image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -285,6 +285,47 @@ def duration(
     return (ends - starts).sum()
 
 
+def expand_file_path(
+        index: pd.Index,
+        root: str,
+) -> pd.Index:
+    r"""Expand index with relative file path.
+
+    Args:
+        index: index with relative file path conform to
+            :ref:`table specifications <data-tables:Tables>`
+        root: directory expand the file path with
+
+    Returns:
+        index with absolute file path
+
+    Raises:
+        ValueError: if index is not conform to
+            :ref:`table specifications <data-tables:Tables>`
+
+    Example:
+        >>> index = filewise_index(['f1', 'f2'])
+        >>> index
+        Index(['f1', 'f2'], dtype='object', name='file')
+        >>> expand_file_path(index, '/some/where')  # doctest: +SKIP
+        Index(['/some/where/f1', '/some/where/f2'], dtype='object', name='file')
+
+    """  # noqa: E501
+    if len(index) == 0:
+        return index
+
+    root = audeer.safe_path(root)
+    is_segmented = index_type(index) == define.IndexType.SEGMENTED
+
+    if is_segmented:
+        files = root + os.path.sep + index.levels[0]
+        index = index.set_levels(files, level=0)
+    else:
+        index = root + os.path.sep + index
+
+    return index
+
+
 def hash(
         obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
 ) -> str:
@@ -313,7 +354,7 @@ def hash(
 
 
 def intersect(
-    objs: typing.Sequence[typing.Union[pd.Index]],
+        objs: typing.Sequence[typing.Union[pd.Index]],
 ) -> pd.Index:
     r"""Intersect index objects.
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -289,7 +289,7 @@ def expand_file_path(
         index: pd.Index,
         root: str,
 ) -> pd.Index:
-    r"""Expand index with relative file path.
+    r"""Expand relative path in index with root.
 
     Args:
         index: index with relative file path conform to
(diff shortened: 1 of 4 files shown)


frankenjoe:
@hagenw sorry, need to rebase first..

frankenjoe:
> @hagenw sorry, need to rebase first..

done

audformat/core/utils.py
```diff
@@ -285,8 +285,49 @@ def duration(
...
+
+    Raises:
+        ValueError: if index is not conform to
+            :ref:`table specifications <data-tables:Tables>`
+
+    Example:
+        >>> index = filewise_index(['f1', 'f2'])
+        >>> index
+        Index(['f1', 'f2'], dtype='object', name='file')
+        >>> expand_file_path(index, '/some/where')  # doctest: +SKIP
```
hagenw, line 310: There should be no reason to skip this test here as it should be independent of the operating system and structure of underlying file system.

audformat/core/utils.py
```diff
@@ -285,8 +285,49 @@ def duration(
     return (ends - starts).sum()
 
 
+def expand_file_path(
+        index: pd.Index,
+        root: str,
+) -> pd.Index:
+    r"""Expand index with relative file path.
```
hagenw, line 292: It's not obvious from this line if you expand a relative path or if you expand with a relative path.

Maybe we change to 

```suggestion
    r"""Expand relative path in index with root.
```

I think it's also not completely right, because we just add `root` add the beginning. Which mean if you set `root='/root` and you have already `/data/test.wav` as file path, you will end up with `/root/data/test.wav` which is not really an expansion of a relative file path.

audformat/core/utils.py
```diff
@@ -285,8 +285,49 @@ def duration(
...
+    """  # noqa: E501
+    if len(index) == 0:
+        return index
+
+    root = audeer.safe_path(root)
+    is_segmented = index_type(index) == define.IndexType.SEGMENTED
+
+    if is_segmented:
+        files = root + os.path.sep + index.levels[0]
+        index = index.set_levels(files, level=0)
```
hagenw, line 322: In #120 I do this in a single line:

```python
index.set_levels(root + index.levels[0], level='file')
```
I think your solution is easier to read, but could you test that there is no difference in speed between one line vs. two lines?

tests/test_utils.py
```diff
@@ -528,6 +529,72 @@ def test_duration(obj, expected_duration):
...
+                ],
+                ['1s', '3s'],
+                ['2s', '4s'],
+            ),
+        )
+    ]
+)
+def test_expand_file_path(tmpdir, index, root, expected):
+    expanded_index = audformat.utils.expand_file_path(index, root)
+    pd.testing.assert_index_equal(expanded_index, expected)
```
hagenw, line 595: Could you add tests for cases 
* where you have already an absolute path in the index
* where you provide `'some/where/'` as root (I know you have already `./`, but I would also test this with an actual folder).

audformat/core/utils.py
frankenjoe: In both solutions it has to calculate `root + index.levels[0]` first, so I don't see how there could be any measureable difference.

audformat/core/utils.py
hagenw: Ah, yes, you are right due to `os.path.sep` it might indeed end up as:
```
/some/where\f1
```
That's in general a problem we have, publishing a database on Windows and using it on Linux will fail.

audformat/core/utils.py
frankenjoe: But let's simply switch to your solution :)

audformat/core/utils.py
hagenw: At least I expect that it will not work ;)

audformat/core/utils.py
frankenjoe: done

tests/test_utils.py
frankenjoe: I added a test with absolute path in the index and two more with `'some/where/` and `'some/where`.

audformat/core/utils.py
```diff
@@ -285,8 +285,48 @@ def duration(
...
+def expand_file_path(
+        index: pd.Index,
+        root: str,
+) -> pd.Index:
+    r"""Expand relative path in index with root.
+
+    Args:
+        index: index with relative file path conform to
+            :ref:`table specifications <data-tables:Tables>`
+        root: directory expand the file path with
```
hagenw, line 297: Maybe:

```suggestion
        root: directory added in front of the relative file path
```

tests/test_utils.py
hagenw: Cool, thanks.

hagenw:
Needs to be rebased locally.

PR #121, opened 2021-11-16
Add audformat.utils.hash()

frankenjoe:
image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -290,8 +290,9 @@ def hash(
 ) -> str:
     r"""Create hash from object.
 
-    Objects with same entries in a different order,
-    will produce the same hash string.
+    Objects with the same elements
+    produce the same hash string
+    independent of the ordering of the elements.
 
     Args:
         obj: object
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -285,10 +285,10 @@ def duration(
     return (ends - starts).sum()
 
 
-def hash_index(
+def hash(
         obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
 ) -> str:
-    r"""Create hash from index.
+    r"""Create hash from object.
 
     Indices with same entries in a different order,
     will produce the same hash string.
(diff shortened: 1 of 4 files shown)


hagenw:
For me it also works if I pass a Series or data frame, but it returns a slightly different value in those cases:

```python
>>> y = pd.Series([1, 2])
>>> df = pd.DataFrame([1, 2])
>>> audformat.utils.hash_index(y)
'61872891661014892'
>>> audformat.utils.hash_index(df)
'61872891661014892'
>>> audformat.utils.hash_index(y.index)
'6238072747940578789'
>>> audformat.utils.hash_index(df.index)
'6238072747940578789'
```

I see two solutions:
* raise an error if not an index is given
* add support for series and dataframe, but hash the index in those cases

hagenw:
I know that we usually don't raise an error if you provide the wrong input type, but the problem is that under `audformat.utils` we have a lot of functions that work the same for index, series and data frame and most likely user will expect the same for `audformat.utils.hash_index()`.
We might also rename it to `audformat.utils.hash()` and say that it is a feature that we get different results for `df` and `df.index`.

frankenjoe:
> We might also rename it to `audformat.utils.hash()` and say that it is a feature that we get different results for `df` and `df.index`.

Sorry, just saw your comment now. I just changed to always hashing the index. But I guess your suggestion makes sense, because you can always do `hash(obj.index)` if you only want to cache the index.

frankenjoe:
> > We might also rename it to `audformat.utils.hash()` and say that it is a feature that we get different results for `df` and `df.index`.
> 
> Sorry, just saw your comment now. I just changed to always hashing the index. But I guess your suggestion makes sense, because you can always do `hash(obj.index)` if you only want to cache the index.

Changed to `hash()`.

audformat/core/utils.py
```diff
@@ -285,6 +285,32 @@ def duration(
     return (ends - starts).sum()
 
 
+def hash(
+        obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
+) -> str:
+    r"""Create hash from object.
+
+    Objects with same entries in a different order,
+    will produce the same hash string.
```
hagenw, line 294: Should we add a test for this or simply trust `pandas` here?

hagenw:
> > > We might also rename it to `audformat.utils.hash()` and say that it is a feature that we get different results for `df` and `df.index`.
> > 
> > 
> > Sorry, just saw your comment now. I just changed to always hashing the index. But I guess your suggestion makes sense, because you can always do `hash(obj.index)` if you only want to cache the index.
> 
> Changed to `hash()`.

Cool, thanks.

audformat/core/utils.py
frankenjoe: This is not really related to `pandas` but the fact that we calculate the sum of the hashes we get for each row.

But since it's never a bad idea to have tests, so I added some now :)

I also rephrased to: 

image

hagenw: Great

PR #122, opened 2021-11-16
Add audformat.utils.replace_file_extension()

hagenw:
Add a utility function to change any existing file extension to the requested one.
As we might need this at different places and you can implement it less efficiently, I thought it would also be good to have it here.

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -695,7 +695,7 @@ def set_file_extension(
         index: index with file path
             conform to :ref:`table specifications <data-tables:Tables>`
         extension: new file extension
-            
+
     Returns:
         updated index
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -687,14 +687,18 @@ def set_file_extension(
 ) -> pd.Index:
     r"""Change the file extension of index entries.
 
-    It replaces all existing file extension
+    It replaces all existing file extensions
     in the index file path
     by the new provided one.
+    The part after the last dot
+    is considered as current file extension.
 
     Args:
         index: index with file path
             conform to :ref:`table specifications <data-tables:Tables>`
-        extension: new file extension
+        extension: new file extension.
+            If set to ``''``,
+            the current file extension is removed
 
     Returns:
         updated index
@@ -703,13 +707,19 @@ def set_file_extension(
         >>> idx = filewise_index(['f1.wav', 'f2.flac'])
         >>> set_file_extension(idx, 'mp3')
         Index(['f1.mp3', 'f2.mp3'], dtype='object', name='file')
+        >>> idx = filewise_index(['f1.wav.gz', 'f2.wav.gz'])
+        >>> set_file_extension(idx, '')
+        Index(['f1.wav', 'f2.wav'], dtype='object', name='file')
 
     """
     if len(index) == 0:
         return index
 
     cur_ext = re.compile(r'\.[a-zA-Z0-9]+$')  # match file extension
-    new_ext = f'.{extension}'
+    if extension:
+        new_ext = f'.{extension}'
+    else:
+        new_ext = ''
     is_segmented = index_type(index) == define.IndexType.SEGMENTED
 
     if is_segmented:
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -712,7 +712,7 @@ def set_file_extension(
         >>> index = filewise_index(['f1.wav.gz', 'f2.wav.gz'])
         >>> set_file_extension(index, '')
         Index(['f1.wav', 'f2.wav'], dtype='object', name='file')
-        >>> set_file_extension(index, 'flac', pattern=r'\.wav\.gz')
+        >>> set_file_extension(index, 'flac', pattern=r'\.wav\.gz$')
         Index(['f1.flac', 'f2.flac'], dtype='object', name='file')
 
     """
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -712,7 +712,7 @@ def set_file_extension(
         >>> index = filewise_index(['f1.wav.gz', 'f2.wav.gz'])
         >>> set_file_extension(index, '')
         Index(['f1.wav', 'f2.wav'], dtype='object', name='file')
-        >>> set_file_extension(index, 'flac', pattern='.wav.gz')
+        >>> set_file_extension(index, 'flac', pattern=r'\.wav\.gz')
         Index(['f1.flac', 'f2.flac'], dtype='object', name='file')
 
     """
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -687,8 +687,15 @@ def set_file_extension(
 ) -> pd.Index:
     r"""Change the file extension of index entries.
 
-    Args:
+    It replaces all existing file extension
+    in the index file path
+    by the new provided one.
 
+    Args:
+        index: index with file path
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -930,6 +930,51 @@ def test_read_csv(csv, result):
...
+        ),
+        (
+            audformat.filewise_index(['f1.WAV', 'f2.WAV']),
+            'MP3',
+            audformat.filewise_index(['f1.MP3', 'f2.MP3']),
+        ),
+        (
+            audformat.filewise_index(['f1', 'f2.wv']),
+            'mp3',
+            audformat.filewise_index(['f1', 'f2.mp3']),
```
hagenw, line 964: Here I'm not completely sure what should happen for a file name without extension. Should it also be changed to `f1.mp3`?

tests/test_utils.py
```diff
@@ -930,6 +930,51 @@ def test_read_csv(csv, result):
...
+        ),
+        (
+            audformat.filewise_index(['f1.wav', 'f2.wav']),
+            '',
+            audformat.filewise_index(['f1.', 'f2.']),
```
hagenw, line 969: Here, I'm also not sure if we should return `['f1', 'f2']` instead?

frankenjoe: Yes, I think that makes more sense.

tests/test_utils.py
frankenjoe: > I did not call the function `replace_file_extension()` as we don't ask to replace a particular file extension

But that means I cannot replace `tar.gz` with `zip`, for instance.

frankenjoe:
> I did not call the function `replace_file_extension()` as we don't ask to replace a particular file extension

Which means I cannot replace `tar.gz` with `zip`, for instance.

hagenw:
> > I did not call the function `replace_file_extension()` as we don't ask to replace a particular file extension
> 
> Which means I cannot replace `tar.gz` with `zip`, for instance.

It depends. It's simply up to us how we define the regexp matching pattern to find the file extension. But I think it will be hard to find a pattern that will always work. So at the moment, I would say it will just replace the extension after the last dot.
If I fix what `extension=''` returns, you could replace `tar.gz` by calling `index = set_file_extension(index, '')` and afterwards `index = set_file_extension(index, 'zip')`.

tests/test_utils.py
hagenw: I changed it and extended the docstring, see the description above.

frankenjoe:
>  It's simply up to us how we define the regexp matching pattern to find the file extension.

Would it maybe make sense that we give the user the option to overwrite the regex pattern?

hagenw:
> > It's simply up to us how we define the regexp matching pattern to find the file extension.
> 
> Would it maybe make sense that we give the user the option to overwrite the regex pattern?

Good idea, this also implicitly documents the used matching pattern. I changed it accordingly.
The only problem is that the formatting in the docstring looks slightly odd:

image

Or is this the correct representation of
```python
r'\.[a-zA-Z0-9]+$'
```
?

frankenjoe:
> The only problem is that the formatting in the docstring looks slightly odd

Maybe use a regular string in the argument and then convert to a raw string later with `pattern = rf'{pattern'}`?

hagenw:
> > The only problem is that the formatting in the docstring looks slightly odd
> 
> Maybe use a regular string in the argument and then convert to a raw string later with `pattern = rf'{pattern}'`?

No, this does not really help. So maybe we just stay with the current solution?

hagenw:
The big problem is that if a user copies the string from the solution and modifies it slightly, e.g. `pattern='\\\\.[A-Z]+$'`, it will fail.

frankenjoe:
> No, this does not really help. So maybe we just stay with the current solution?

Ok, so the problem is really that `sphinx` replaces `\` with `\\\\` correct?

hagenw:
> > No, this does not really help. So maybe we just stay with the current solution?
> 
> Ok, so the problem is really that `sphinx` replaces `\` with `\\\\` correct?

Yes, it replaces first `r'\'` with `'\\'` which is still correct and then `\` with `\\`

frankenjoe:
> Yes, it replaces first `r'\'` with `'\\'` which is still correct and then `\` with `\\`

We could set it to `None` and then say in the docstring what is the default value?

audformat/core/utils.py
```diff
@@ -640,6 +641,62 @@ def same_dtype(d1, d2) -> bool:
...
+        updated index
+
+    Example:
+        >>> idx = filewise_index(['f1.wav', 'f2.flac'])
+        >>> set_file_extension(idx, 'mp3')
+        Index(['f1.mp3', 'f2.mp3'], dtype='object', name='file')
+        >>> idx = filewise_index(['f1.wav.gz', 'f2.wav.gz'])
+        >>> set_file_extension(idx, '')
+        Index(['f1.wav', 'f2.wav'], dtype='object', name='file')
+        >>> set_file_extension(idx, 'flac', pattern='.wav.gz')
```
frankenjoe, line 675: For best practise maybe change to ` pattern='\.wav\.gz$'?

audformat/core/utils.py
```diff
@@ -640,6 +641,62 @@ def same_dtype(d1, d2) -> bool:
...
+            the current file extension is removed
+        pattern: regexp pattern to match current extensions.
+            In contrast to ``extension``,
+            you have to include ``'.'``
+
+    Returns:
+        updated index
+
+    Example:
+        >>> idx = filewise_index(['f1.wav', 'f2.flac'])
```
frankenjoe, line 669: For consistency with the other examples I would rename to `index`,

hagenw: There is not really consitency ;)

image

image

image

image

But I agree, that `index` makes the most sense here. I changed it accordingly

frankenjoe: Agree, `i1` and `i2` is really bad naming. I would like to say it wasn't me who added those examples, but I'm afraid that'd be a lie :)

audformat/core/utils.py
hagenw: Good that you reminded me of that one. I changed to `pattern=r'\.wav\.gz$'`.

frankenjoe: In principle, the function can now be used to replace any parts of the file path. But I guess it still makes sense to stick keep the name `audformat.utils.set_file_extension()` as this is the most likely use case.

hagenw:
> > Yes, it replaces first `r'\'` with `'\\'` which is still correct and then `\` with `\\`
> 
> We could set it to `None` and then say in the docstring what is the default value?

The problem with using `None` and stating the default string in the documentation is that you might update the default string later in the code and forget about it in the docs. But I also see no other meaningful way at the moment.

audformat/core/utils.py
hagenw: I would now also be fine with using `replace_file_extension()`, but I don't have an opinion which is better.

hagenw:
> > > Yes, it replaces first `r'\'` with `'\\'` which is still correct and then `\` with `\\`
> > 
> > 
> > We could set it to `None` and then say in the docstring what is the default value?
> 
> The problem with using `None` and stating the default string in the documentation is that you might update the default string later in the code and forget about it in the docs. But I also see no other meaningful way at the moment.

I switched now to `pattern=None`.

audformat/core/utils.py
frankenjoe: I think I would prefer `replace_file_extension()` now.

hagenw: Renamed it

PR #124, opened 2021-11-17
Release 0.12.1

hagenw:
After https://github.com/audeering/audformat/pull/122 is merged I think we should make a new release in order to continue work on `auglib` and `audb`.


hagenw:
#122 is merged now, so we can make a new release.

PR #126, opened 2021-11-17
Do no longer test for duplicates in assert_index()

### LINKED ISSUE #125: assert_index() can be a bottleneck
See https://github.com/audeering/audb/pull/149#issuecomment-971586169

Possible solutions:

* Removing it everywhere and leave it completely to the user to call assert_index().
* Separate `obj.has_duplicates` from `assert_index()` (e.g. add `assert_no_duplicates()`) and call it only if really needed.
*  Add an argument to `assert_index()` (and functions that call it) to disable checking for duplicates.

hagenw:
Closes #125 

As discussed below, we decided to no longer test for duplicates in `audformat.assert_index()`.
Instead, we add `audformat.assert_no_duplicates()`, that can then be used inside `audb.publish()` to ensure only tables without duplicates are published.

This modification brings down the execution time in https://github.com/audeering/audb/pull/149 from 20 s to 0.004 s.

image

I also mention the new function in the docstring of `audformat.assert_index()`

image

### DIFF
--- a/audformat/__init__.py
+++ b/audformat/__init__.py
@@ -4,6 +4,7 @@ from audformat import utils
 from audformat.core.database import Database
 from audformat.core.index import (
     assert_index,
+    assert_no_duplicates,
     filewise_index,
     segmented_index,
     index_type,
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -117,14 +117,12 @@ def assert_index(
 def assert_no_duplicates(
     obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
 ):
-    r"""Assert object is contains no duplicates in its index.
+    r"""Assert object contains no duplicates in its index.
 
     The :ref:`table specifications <data-tables:Tables>`
     allow no duplicated index entries.
     To save time we do not test for this
-    when initializing :class:`audformat.Table`.
-    This function provides you the possibility
-    to manual check for it.
+    in :func:`audformat.assert_index`.
 
     Args:
         obj: object
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -35,13 +35,21 @@ def to_timedelta(times):
 
 
 def assert_index(
-    obj: typing.Union[pd.Index, pd.Series, pd.DataFrame]
+    obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
+    *,
+    allow_duplicates: bool = False,
 ):
     r"""Assert object is conform to :ref:`table specifications
     <data-tables:Tables>`.
 
     Args:
         obj: object
+        allow_duplicates: if ``True``
+            it will not check for duplicated entries,
+            which is faster,
+            but should only be used
+            if you know
+            that the index does not contain duplicates
 
     Raises:
         ValueError: if not conform to
@@ -51,7 +59,7 @@ def assert_index(
     if isinstance(obj, (pd.Series, pd.DataFrame)):
         obj = obj.index
 
-    if obj.has_duplicates:
+    if not allow_duplicates and obj.has_duplicates:
         max_display = 10
         duplicates = obj[obj.duplicated()]
         msg_tail = '\n...' if len(duplicates) > max_display else ''
@@ -183,7 +191,7 @@ def index_type(
     if isinstance(obj, (pd.Series, pd.DataFrame)):
         obj = obj.index
 
-    assert_index(obj)
+    assert_index(obj, allow_duplicates=True)
 
     if len(obj.names) == 1:
         return define.IndexType.FILEWISE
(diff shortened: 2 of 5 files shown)


frankenjoe:
I'm not so happy with the name `allow_duplicates` as it implies that in that case duplicates are fine. Maybe `check_duplicates` would be a better name?

frankenjoe:
We could also assume a user will always create an `audformat` index using `segmented_index()` and `filewise_index()`. So actually it it should be sufficient to call `assert_index()` in those two functions and everywhere else we can assume that we have a proper index.

frankenjoe:
(sorry accidentally closed by hitting the wrong button, was not on purpose:)

hagenw:
I'm not so sure that we an simply assume a user will always create an index using `audformat.filewise_index()` or `audformat.segmented_index()`. She can also manually create an index and assign to a table, if this is then published our assumption that databases do not contain duplicated index entries does alos no longer hold.

I think we should not only check in `Table.__init__()` as I propose here, but also inside `Table.extend_index()`.

hagenw:
> I'm not so happy with the name `allow_duplicates` as it implies that in that case duplicates are fine. Maybe `check_duplicates` would be a better name?

I also first thought about `check_duplicates` and maybe we just switch to it. But your argument against `allow_duplicates` is not really an argument, because if you call `assert_index(index, allow_duplicates=True)` it is exactly how you describe it: in this case duplicates are fine. Internally, we make sure that we call it only on an index that has no duplicates, but as this function is part of the API a user can call it on whatever index.

frankenjoe:
> I'm not so sure that we an simply assume a user will always create an index using `audformat.filewise_index()` or `audformat.segmented_index()`. She can also manually create an index and assign to a table, if this is then published our assumption that databases do not contain duplicated index entries does no longer hold.

This we can capture by calling `assert_index()` in `audb.publish()'. Since publishing a large database will anyway take a lot of time, the little overhead would be fine. But since in all functions we explicitly say that the index / table has to be conform with  `audformat`, I don't think there is a need to check it.

> it is exactly how you describe it: in this case duplicates are fine. 

Well, duplicates are never fine, because it's against the specification :)

> I think we should not only check in Table.__init__() as I propose here, but also inside Table.extend_index().

Exactly, and I think there are more functions where it makes sense. But then in 99% of the cases the user as probably used  `audformat.filewise_index()` or `audformat.segmented_index()` to create the index, i.e. `assert_index()` is called twice on the same index. That's why I would propose we only call it in `audformat.filewise_index()` and `audformat.segmented_index()`.

hagenw:
> But since in all functions we explicitly say that the index / table has to be conform with audformat, I don't think there is a need to check it.

Yes, but users simply don't care. E.g. I would first try to create the index with `filewise_index()` this will then return an error. As a next step I will try to create the index manually and then pass it on to `Table()`, voila, it works :)

In principle I agree that it would be fine to just check inside `audb.publish()`. But there is one problem: as it is a very costly operation and we assume that most of the indices are created with `filewise_index()` and or `segmented_index()` anyway, it would mean calling it twice for most of the tables. And as I know from experience you might have to run your publication scripts several times in order to debug them. Hence I would vote for doing the check only in `audformat` or only in `audb.publish()`. 
As it is part of the audformat specification it seems to make more sense to run the check inside `audformat`.

frankenjoe:
> But there is one problem: as it is a very costly operation and we assume that most of the indices are created with `filewise_index()` and or `segmented_index()` anyway, it would mean calling it twice for most of the tables.

Yes, but that is exactly my argument for not calling `assert_index()` in `Table()`.

frankenjoe:
>  Hence I would vote for doing the check only in audformat or only in audb.publish()

I would say it's safer to do it in `audb.publish()` cause a user could still directly modifying `Table._df` for instance. So the only way to really guarantee published tables follow the specifications is by calling it in `publish()`.

hagenw:
OK, how about the following solution:

* we provide a function to check for duplicates in `audformat` (e.g. adding `check_duplicates=False` to `assert_index()` or as a new function `assert_no_duplicates()`)
* we check for duplicates only inside `audb.publish()`

frankenjoe:
>     * we provide a function to check for duplicates in `audformat` (e.g. adding `check_duplicates=False` to `assert_index()` or as a new function `assert_no_duplicates()`)

Ok, so basically we keep the calls to `assert_index()` as is, but we remove the check for duplicates so the overhead calling this function is basically zero and it will also speed up `filewise_index()` and `segmented_index()`. Then we introduce `assert_no_duplicates()` as a new function which we only call in `audb.publish()` since this is the place where we really want to make sure we have no duplicates in the table.

Yes that sounds like a good solution to me!

hagenw:
OK, I changed the code accordingly and updated the description.

frankenjoe:
Looks great. Looking forward to your new speed tests in `audb` :)

PR #127, opened 2021-11-18
Release 0.12.2

hagenw:


hagenw:
I did a brief check and our changes here are indeed fixing the problem described in https://github.com/audeering/audb/pull/149 and are also not slower in other circumstances, so I will publish a new release to continue in `audb`.

PR #133, opened 2022-01-04
Remove deprecated pandas code

hagenw:
This avoids deprecation warnings when running the tests.

### DIFF
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -268,10 +268,10 @@ class Scheme(HeaderBase):
             for table in self._db.tables.values():
                 for column in table.columns.values():
                     if column.scheme_id == self._id:
-                        column.get(copy=False).cat.set_categories(
+                        y = column.get(copy=False)
+                        y.cat.set_categories(
                             new_categories=self.labels,
                             ordered=False,
-                            inplace=True,
                         )
 
     def _dtype_from_labels(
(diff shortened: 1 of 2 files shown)


audformat/core/scheme.py
```diff
@@ -268,10 +268,10 @@ def replace_labels(
             for table in self._db.tables.values():
                 for column in table.columns.values():
                     if column.scheme_id == self._id:
-                        column.get(copy=False).cat.set_categories(
+                        y = column.get(copy=False)
+                        y.cat.set_categories(
                             new_categories=self.labels,
                             ordered=False,
-                            inplace=True,
                         )
```
hagenw, line 275: Unfortunately this does not work. But as usual I guess, I don't understand the `copy=False` stuff.
My solution here to avoid `inplace=True` comes from https://stackoverflow.com/a/70344396

hagenw: Ah, yes, sorry, that was also my first solution.

I had a similar problem before with a table. I think as soon as we do `.get()` it does no longer seem to refer to the object belonging to the database object.

frankenjoe: Yes, I'm trying to debug right now what's going on.

frankenjoe: I pushed a possible fix that avoids calling `get()`.

hagenw: I think more important than not using `get()` is that you added a `column._table.df[column._id] = y` afterwards.
But I like your solution and you can continue reviewing the rest of this pull request. Thanks for providing the fix.

PR #137, opened 2022-01-04
WIP: Set dtype when reading index from CSV file

### LINKED ISSUE #33: Speedup by specifying dtype for index columns when reading CSV files
Reading CSV files with `pandas.read_csv()` is faster if you provide data types of the columns, e.g. https://towardsdatascience.com/️-load-the-same-csv-file-10x-times-faster-and-with-10x-less-memory-️-e93b485086c7

At the moment we [specify data types for all the data columns](https://github.com/audeering/audformat/blob/de6073060de76d3d413baff30455c8cd0fbf6b13/audformat/core/table.py#L882-L894),
but [not for the index columns](https://github.com/audeering/audformat/blob/de6073060de76d3d413baff30455c8cd0fbf6b13/audformat/core/table.py#L867-L876).

I'm also wondering  if `dtype = schemes[column.scheme_id].to_pandas_dtype()` is sufficient to detect [categorical](https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html) data types instead of strings.

hagenw:
Closes #33.

In #33 we realized that we don't specify the `dtype` for the index columns when reading from CSV. I think for `start` and `end` we don't need to specify the data type as we use `converters` anyway. The [pandas documentation](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) is not explicit if `dtype` is applied to `index_col` as well.
At least my implementation does not raise an error. But I haven't tested yet that it is faster.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1084,6 +1084,7 @@ class Table(HeaderBase):
 
         # index columns
 
+        dtypes[define.IndexField.FILE] = 'str'
         if self.type == define.IndexType.SEGMENTED:
             converters[define.IndexField.START] = \
                 lambda x: pd.to_timedelta(x)


hagenw:
Closing in favor of https://github.com/audeering/audformat/pull/204

frankenjoe:
I think this was solved in #204

frankenjoe:
Hehe, you were 16 seconds faster than me :)

PR #139, opened 2022-01-05
WIP: Allow index with additional level entries

### LINKED ISSUE #86: Allow for duplicates in index?
At the moment we have the problem, that we don't cover some use cases with our available tables in `audformat`.
If you have as input the comparison between two files, or a file combined with another column as combination and a third with the result, we can not handle it as we do not allow for duplicate index entries.

One example would be a table used for verification experiments:

```
                                                   verification file  same speaker
file                                                                              
wav/id10270/x6uYqmx31kE/00001.wav  wav/id10300/ize_eiCFEg0/00003.wav         False
wav/id10270/x6uYqmx31kE/00001.wav  wav/id10270/GWXujl-xAVM/00017.wav          True
wav/id10270/x6uYqmx31kE/00001.wav  wav/id10273/0OCW1HUxZyg/00001.wav         False
wav/id10270/x6uYqmx31kE/00001.wav  wav/id10270/8jEAjG6SegY/00022.wav          True
```

hagenw:
Closes #86.

This expands our implementation of `audformat` by allowing the `pandas` index objects to contain additional columns. I can thing of several applications where this is needed, e.g. comparing two segmented files. I also do not see any harm we introducing it.

I implemented it the following way:
* the first level is always considered to have the name `'file'`
* if we have at least three levels and the second and third are named `'start'` and `'end'` is the index considered a segmented index
All the checks for the correct data type are still performed.

I also expanded the docstring of `audformat.assert_index()` by stating what we expect from the index and highlighting it with examples

image

In addition, I added two new examples to `audformat.index_type()` including index with extra levels

image

### DIFF
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -44,6 +44,14 @@ def assert_index(
     If you need that check
     use :func:`audformat.assert_no_duplicates` in addition.
 
+    If the index is a :class:`pandas.MultiIndex`
+    the first level is considered
+    to have the name ``'file'``
+    and data type ``string``.
+    If the name of the second level is ``'start'``
+    and the name of the third level ``'end'``
+    both need to be of type ``timedelta64[ns]``.
+
     Args:
         obj: object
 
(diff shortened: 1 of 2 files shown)


frankenjoe:
Are indices with additional columns supposed to work with the utility functions? At the moment this seems not to be the case, e.g.:

```python
index = pd.MultiIndex.from_arrays(  # filewise with extra level
    [
        ['f1', 'f2'],
        [0, 0],
    ],
    names=['file', 'a'],
)
audformat.utils.to_segmented_index(index)
```
```
MultiIndex([(('f1', 0), '0 days', NaT),
            (('f2', 0), '0 days', NaT)],
```

We can see that after converting to a segmented index, the additional level is combined with the file level.

frankenjoe:
Combining a regular index with an index that has an additional level gives the following error:

```python
index1 = audformat.filewise_index(['f1', 'f3'])
index2 = pd.MultiIndex.from_arrays(  # filewise with extra level
    [
        ['f1', 'f2'],
        [0, 0],
    ],
    names=['file', 'a'],
)
audformat.utils.union([index1, index2])
```
```
TypeError: '<' not supported between instances of 'tuple' and 'str'
```

hagenw:
You are right. I checked and it effects the following util functions:

* `audformat.utils.concat()`
* `audformat.utils.expand_file_path()`
* `audformat.utils.intersect()`
* `audformat.utils.replace_file_extension()`
* `audformat.utils.to_filewise_index()`
* `audformat.utils.to_segmented_index()`
* `audformat.utils.union()`

and besides that it might also influence stuff we are not aware of yet. The problem is that a filewise index can now be of type `pd.Index` or `pd.MultiIndex`, which is not very nice.

I will mark the pull request as WIP and see if it is worth the effort fixing all utils functions or if we can find another solution to fix #86.

hagenw:

PR #140, opened 2022-01-12
Release 0.12.4

hagenw:
As I don't think we will be able to make the 1.0.0 during the next days, I would propose to make another bug fix release as it contains one important fix.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,20 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 0.12.4 (2022-01-12)
+---------------------------
+
+* Changed: the API documentation on the ``language`` argument
+  of ``audformat.Database`` is more verbose now
+* Changed: the difference between
+  ``audformat.define.DataType.TIME``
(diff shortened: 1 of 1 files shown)


frankenjoe:
What are the open issues we should fix before `1.0.0`?

hagenw:
https://github.com/audeering/audformat/issues/86, e.g. as it is important to support comparisons of two inputs. I have the feeling that we should fix this before releasing 1.0.0 and writing a paper about `audformat` as this is widely used in some parts of the audio machine learning community.

frankenjoe:
> writing a paper about audformat as this is widely 

Ok, so you propose that we release `1.0.0` together with the paper then?

> as this is widely used in some parts of the audio machine learning community.

Really? :)

hagenw:
> Ok, so you propose that we release 1.0.0 together with the paper then?

It don't have to be at the same time, but I would first like to release 1.0.0 before writing a paper.

frankenjoe:
> It don't have to be at the same time, but I would first like to release 1.0.0 before writing a paper.

Ah ok, then writing the paper is not a blocker for releasing 1.0.0.

PR #142, opened 2022-01-24
Save pickled tables with protocol 4

### LINKED ISSUE #141: Pickled files with Python 3.8 not compatible with older versions
In Python 3.8 a new pickle protocol 5 was introduced, see https://docs.python.org/3/library/pickle.html#data-stream-format

Since it is not understood by older Python versions it might be safer to set `protocol=4` when saving tables to pickle: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_pickle.html

frankenjoe:
Closes #141 

* Force protocol = 4 when storing tables to pickle files.
* Try to recover from CSV file if PKL cannot be loaded.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1171,10 +1171,10 @@ class Table(HeaderBase):
         with open(path, 'w') as fp:
             df.to_csv(fp, encoding='utf-8')
 
-    def _save_pickled(self, path: str):        
+    def _save_pickled(self, path: str):
         self.df.to_pickle(
             path,
-            protocol=4,  # supported by Python >= 3.4 
+            protocol=4,  # supported by Python >= 3.4
         )
 
     def _set_column(self, column_id: str, column: Column) -> Column:
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -647,7 +647,17 @@ class Table(HeaderBase):
             pickled = True
 
         if pickled:
-            self._load_pickled(pkl_file)
+            try:
+                self._load_pickled(pkl_file)
+            except Exception as ex:
+                # if exception is raised (e.g. unsupported pickle protocol)
+                # try to load from CSV and save it again
+                # otherwise raise error
+                if os.path.exists(csv_file):
+                    self._load_csv(csv_file)
+                    self._save_pickled(pkl_file)
+                else:
+                    raise ex
         else:
             self._load_csv(csv_file)
 
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1161,8 +1161,11 @@ class Table(HeaderBase):
         with open(path, 'w') as fp:
             df.to_csv(fp, encoding='utf-8')
 
-    def _save_pickled(self, path: str):
-        self.df.to_pickle(path)
+    def _save_pickled(self, path: str):        
+        self.df.to_pickle(
+            path,
+            protocol=4,  # supported by Python >= 3.4 
+        )
 
     def _set_column(self, column_id: str, column: Column) -> Column:
         if column.scheme_id is not None and \
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -649,7 +649,7 @@ class Table(HeaderBase):
         if pickled:
             try:
                 self._load_pickled(pkl_file)
-            except Exception as ex:
+            except (ValueError, EOFError) as ex:
                 # if exception is raised (e.g. unsupported pickle protocol)
                 # try to load from CSV and save it again
                 # otherwise raise error
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -844,18 +844,45 @@ def test_get_as_segmented():
 
 
 def test_load(tmpdir):
+
+    path_pkl = os.path.join(tmpdir, 'db.table.pkl')
+    path_no_ext = os.path.join(tmpdir, 'db.table')
+
     # Test backward compatibility
-    table_file = os.path.join(tmpdir, 'db.table.pkl')
     table = create_db_table(
         pd.Series(
             [1., 2.],
             index=audformat.filewise_index(['f1', 'f2']),
+            name='column',
         )
     )
-    table.df.to_pickle(table_file, compression='xz')
+    table.df.to_pickle(path_pkl, compression='xz')
     table_loaded = audformat.Table()
-    table_loaded.load(table_file[:-4])
+    table_loaded.load(path_no_ext)
     pd.testing.assert_frame_equal(table.df, table_loaded.df)
+
+    # corrupt pickle file
+    with open(path_pkl, 'wb'):
+        pass
+    with pytest.raises(EOFError):
+        table_loaded.load(path_no_ext)
+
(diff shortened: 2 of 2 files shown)


frankenjoe:
Mhh, it seems we are not compatible with `Python 3.9`. Since the error is raised by `to_csv` I don't think it's related to the changes in this PR.

```

Run python -m pytest
ImportError while loading conftest '/home/runner/work/audformat/audformat/tests/conftest.py'.
tests/conftest.py:14: in <module>
    pytest.DB.save(pytest.DB_ROOT)
audformat/core/database.py:530: in save
    audeer.run_tasks(
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/audeer/core/utils.py:444: in run_tasks
    results[index] = task_func(*param[0], **param[1])
audformat/core/database.py:524: in job
    table.save(
audformat/core/table.py:789: in save
    self._save_csv(csv_file)
audformat/core/table.py:1162: in _save_csv
    df.to_csv(fp, encoding='utf-8')
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/generic.py:3563: in to_csv
    return DataFrameRenderer(formatter).to_csv(
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/io/formats/format.py:1180: in to_csv
    csv_formatter.save()
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/io/formats/csvs.py:261: in save
    self._save()
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/io/formats/csvs.py:266: in _save
    self._save_body()
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/io/formats/csvs.py:304: in _save_body
    self._save_chunk(start_i, end_i)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/io/formats/csvs.py:311: in _save_chunk
    res = df._mgr.to_native_types(**self._number_format)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/internals/managers.py:473: in to_native_types
    return self.apply("to_native_types", **kwargs)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/internals/managers.py:304: in apply
    applied = getattr(b, f)(**kwargs)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/internals/blocks.py:636: in to_native_types
    result = to_native_types(self.values, na_rep=na_rep, quoting=quoting, **kwargs)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/internals/blocks.py:2148: in to_native_types
    values = take_nd(
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/array_algos/take.py:114: in take_nd
    return arr.take(indexer, fill_value=fill_value, allow_fill=allow_fill)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/arrays/masked.py:653: in take
    return type(self)(result, mask, copy=False)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/pandas/core/arrays/integer.py:315: in __init__
    raise TypeError(
E   TypeError: values should be integer numpy array. Use the 'pd.array' function instead
Error: Process completed with exit code 4.
```

frankenjoe:
After disabling`Python 3.9` it now fails with `Python 3.8`. Maybe a problem with the latest pandas / numpy versions...?

hagenw:
> Ok, they released pandas 1.4.0 two days ago. When forcing <1.4.0 tests are again passing. Looking at https://pandas.pydata.org/docs/dev/whatsnew/v1.4.0.html#i-o it seems they changed a lot in the `pandas.read_csv()` function.

Oh dear, what they are thinking?

hagenw:
Seems like they simply implemented `pandas` too complicated and it is very hard to maintain.
But okay, let's see if we can adapt to it in another pull request.

frankenjoe:
Either that or we wait a bit to see if they provide a fix like last time, see https://github.com/audeering/audformat/pull/101#issuecomment-957350325

setup.cfg
```diff
@@ -33,7 +33,7 @@ install_requires =
     oyaml
     pyyaml >=5.4.1
     # https://github.com/audeering/audformat/pull/101
-    pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3
+    pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3,<1.4.0
```
hagenw, line 36: As we are still hoping for a fix, should we maybe use 
```suggestion
    pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3,!=1.4.0
```
instead?

hagenw:
Have you checked what will happen with databases that were stored already using pickle protocol 5?

frankenjoe:
> Have you checked what will happen with databases that were stored already using pickle protocol 5?

We only change the way we write the pickle file here, i.e. it will still fail with Python <3.6. The only thing we could do is fall back to the CSV in case we cannot read the pickle.

setup.cfg
frankenjoe: Since we do not know if it will be fixed with `1.4.1` it might be a bit risky.

hagenw: OK, you are maybe right. The best option would be to fix the error, but here i also agree that it is not worth the effort and we should first wait if `pandas` fixes it.

So let's stay with `<1.4.0`

hagenw:
> The only thing we could do is fall back to the CSV in case we cannot read the pickle.

This might be not a bad idea as maybe a few users have now stored already databases with the new pickle version.
I would propose to load the CSV file and overwrite the PKL file if reading the pickle file fails.

frankenjoe:
> I would propose to load the CSV file and overwrite the PKL file if reading the pickle file fails.

done

audformat/core/table.py
```diff
@@ -647,7 +647,17 @@ def load(
             pickled = True
 
         if pickled:
-            self._load_pickled(pkl_file)
+            try:
+                self._load_pickled(pkl_file)
+            except Exception as ex:
```
hagenw, line 652: Could you replace `Exception` with the actual exception to not catch keyboard interrupts or similar stuff here.

hagenw: Maybe we could support two different kind of exceptions here. One of the non-matching pickle protocol and the other with a corrupted pickle file as you used in the tests?

frankenjoe: Ok, replaced it with `ValueError`, which raised if the pickle protocol is not understood, and `EOFError`, which we use in the test.

hagenw: Interesting that they go with `ValueError` for it. But thanks for changing.

PR #151, opened 2022-02-07
Fix conversion of pickle versions in cache

hagenw:
This also excepts `AttributeError` when a wrong pickle format is found.


frankenjoe:
Let's hope we now cover all possible errors.

PR #155, opened 2022-02-24
Don't allow \ in a portable database

hagenw:
This extends `audformat.Database.is_portable` to not allow for Windows path notation `\`.

image


hagenw:
To fix the Windows test we need to merge https://github.com/audeering/audformat/pull/154 first.

frankenjoe:
> To fix the Windows test we need to merge #154 first.

done

PR #161, opened 2022-03-15
Add utility method to explode overlapping segments

Bahaabrougui:
This MR adds a utility function to add the feature addressed in this [issue](https://gitlab.audeering.com/tools/audbenchmark/-/issues/163#note_179890).  
The tests are failing because of `AttributeError: module 'audformat.utils' has no attribute 'explode_overlapping_segments'`. I wrote the test case in `test_utils.py`, and added the method as import statements in the `__init__.py` file under `utils/`. Is there something else that I should do and I didn't ?

PS: I'm not very proud about the algorithm as it is iterating over the Serie/Dataframe 4 times (4 is just a choice I made, sometimes the overlapping can be resolved after the first iteration, but that depends how many overlapping we have, 4 is a safe choice). Another option to handle this is to use `audbenchmark.ToFrames()` separetely on every class column (when we have classes as a one-hot encoded vector) and then concatenate the generated series, but that will apply the function `ToFrames()` as much times as events we have in the vector (if we have 4 classes, we'll call `ToFrames` on the same files 4 times), and that's also not a very good work around, but it can solve the problem and makes us not need the method I added here. But I think this method is also useful and would be good to have it included.


hagenw:
You need to make sure the function is imported in https://github.com/audeering/audformat/blob/master/audformat/utils/__init__.py

audformat/core/utils.py
```diff
@@ -1074,3 +1074,232 @@ def union(
...
+            file            start                      end
+            audio_file.wav  0 days 00:00:00.861768592  0 days 00:00:09.581768592    (1, 0, 0, 0, 0)  # noqa
+                            0 days 00:00:03.428162853  0 days 00:00:04.488162853    (0, 0, 1, 0, 0)  # noqa
+            >>> obj = explode_overlapping_segments(obj)
+            file            start                      end
+            audio_file.wav  0 days 00:00:00.861768592 0 days 00:00:03.428162853  (1, 0, 0, 0, 0)  # noqa
+                            0 days 00:00:03.428162853 0 days 00:00:04.488162853  (1, 0, 1, 0, 0)  # noqa
+                            0 days 00:00:04.488162853 0 days 00:00:09.581768592  (1, 0, 0, 0, 0)  # noqa
+
+    """
```
hagenw, line 1123: You can specify the `noqa` at the end of the docstring:

```suggestion
            file            start                      end
            audio_file.wav  0 days 00:00:00.861768592  0 days 00:00:09.581768592    (1, 0, 0, 0, 0)
                            0 days 00:00:03.428162853  0 days 00:00:04.488162853    (0, 0, 1, 0, 0)
            >>> obj = explode_overlapping_segments(obj)
            file            start                      end
            audio_file.wav  0 days 00:00:00.861768592 0 days 00:00:03.428162853  (1, 0, 0, 0, 0)
                            0 days 00:00:03.428162853 0 days 00:00:04.488162853  (1, 0, 1, 0, 0)
                            0 days 00:00:04.488162853 0 days 00:00:09.581768592  (1, 0, 0, 0, 0)

    """  noqa: E501
```

Bahaabrougui:
> You need to make sure the function is imported in https://github.com/audeering/audformat/blob/master/audformat/utils/__init__.py
@hagenw it is, I don't why locally it fails and I get that error while here it runs fine (although tests are still failing and I'm looking into it). I'm doing `pytest testing/test_utils.py`, is it not the right way to do it ?

audformat/core/utils.py
```diff
@@ -1074,3 +1074,232 @@ def union(
...
+            >>> obj = pd.Series(index=pd.MultiIndex.from_arrays([
+            ...     ['audio_file.wav',
+            ...     'audio_file.wav'],
+            ...     [pd.to_timedelta('0 days 00:00:00.861768592'),
+            ...     pd.to_timedelta('0 days 00:00:03.428162853')],
+            ...     [pd.to_timedelta('0 days 00:00:09.581768592'),
+            ...     pd.to_timedelta('0 days 00:00:04.488162853')]
+            ...     ], names=('file', 'start', 'end')), data=[(1, 0, 0, 0, 0)
+            ...     , (0, 0, 1, 0, 0)])
+            ...     )
```
hagenw, line 1113: You can use `audformat.segmented_index()` to create the desired index

```suggestion
            >>> obj = pd.Series(
            ...     index=segmented_index(
            ...         ['audio_file.wav', 'audio_file.wav'],
            ...         [0.861768592, 3.428162853],
            ...         [9.581768592, 4.488162853],
            ...     ),
            ...     data=[(1, 0, 0, 0, 0), (0, 0, 1, 0, 0)],
            ... )
```

hagenw (commented):
There is one general question we need to check as well. I'm not sure how support for sequences in `audformat` is at the moment, because we cannot store `(0, 1, 0)` in a meaningful way in the CSV file. I guess it will be stored as a string representation.

@Bahaabrougui did you try to store and load such a table as CSV file?

Bahaabrougui:
> There is one general question we need to check as well. I'm not sure how support for sequences in `audformat` is at the moment, because we cannot store `(0, 1, 0)` in a meaningful way in the CSV file. I guess it will be stored as a string representation.
> 
> @Bahaabrougui did you try to store and load such a table as CSV file?

Yes I'd assume it will be stored as string too, but let me try and do it and get back to you

Bahaabrougui:
> > There is one general question we need to check as well. I'm not sure how support for sequences in `audformat` is at the moment, because we cannot store `(0, 1, 0)` in a meaningful way in the CSV file. I guess it will be stored as a string representation.
> > @Bahaabrougui did you try to store and load such a table as CSV file?
> 
> Yes I'd assume it will be stored as string too, but let me try and do it and get back to you

Update : So I created a series in the desired format with data being a `tuple` and exported it  
```python
serie = pd.Series(
    index=segmented_index(
        ['audio_file.wav', 'audio_file.wav'],
        [0.861768592, 3.428162853],
        [9.581768592, 4.488162853]
    ),
    data=[(1, 0, 0), (0, 0, 1)]
)
serie.to_csv('/data/beddine/test.csv')
```

Then I loaded it and I printed the type of the data, and it is string as expected
```python
                                                                            0
file           start                     end                                 
audio_file.wav 0 days 00:00:00.861768592 0 days 00:00:09.581768592  (1, 0, 0)
               0 days 00:00:03.428162853 0 days 00:00:04.488162853  (0, 0, 1)
<class 'str'>
```

frankenjoe:
I had a brief look at the PR. Two comments:

1. I don't think this function fits into `audformat`. The utility functions we have so far do not make assumptions on the content of a table and here we deal with a very specific one hot encoding. 
2. The algorithms looks quite complicated and in the description you say it needs several iterations to make sure it actually converges (and do we really know that 4 iterations are always enough?). The alternate approach you outlined and where you apply the existing algorithm (which assumes no overlap) on every column and concatenate the results afterwards sounds like a better solution to me. Probably only takes a couple of lines to write it. So I would argue in favor of maintainability over speed here.

hagenw:
> 1. I don't think this function fits into `audformat`. The utility functions we have so far do not make assumptions on the content of a table and here we deal with a very specific one hot encoding.

Do you have any suggestion where to move it instead? We don't have something like `audbenchmark.utils` at the moment, but I guess the `audbenchmark` package should be the best fit.

hagenw:
I also think that the current solution is too complex as in my opinion it should be possible by using a slightly modified version of algorithms discussed at https://codereview.stackexchange.com/questions/69242/merging-overlapping-intervals

hagenw:
In order to proceed here I would suggest we first agree on where to move this pull request to, then we move it, and then @Bahaabrougui tries to find an easier algorithm to solve the problem, and then I will start with the actual code review.

frankenjoe:
> Do you have any suggestion where to move it instead? We don't have something like `audbenchmark.utils` at the moment, but I guess the `audbenchmark` package should be the best fit.

Not sure if it really has to be a utility function. We could directly implement it as `audbechmak.mapper.ToFramesWithOverlap`. In case more classes will follow that are specific to the problem of mixing databases, it could also make sense to create a new package.

hagenw:
> > Do you have any suggestion where to move it instead? We don't have something like `audbenchmark.utils` at the moment, but I guess the `audbenchmark` package should be the best fit.
> 
> Not sure if it really has to be a utility function. We could directly implement it as `audbechmak.mapper.ToFramesWithOverlap`. In case more classes will follow that are specific to the problem of mixing databases, it could also make sense to create a new package.

Makes sense to me. @Bahaabrougui if you agree, please open the corresponding merge request in `audbenchmark`.
BTW, in principle we could also support non hot encoded label, e.g.

``` 
file           start                     end                                 
audio_file.wav 0 days 00:00:00.861768592 0 days 00:00:09.581768592  a
               0 days 00:00:03.428162853 0 days 00:00:04.488162853  b
```

could become

```
file           start                     end
audio_file.wav 0 days 00:00:00 0 days 00:00:01  a
               0 days 00:00:01 0 days 00:00:02  a
               0 days 00:00:02 0 days 00:00:03  a
               0 days 00:00:03 0 days 00:00:04  NaN
               0 days 00:00:04 0 days 00:00:05  a
               0 days 00:00:05 0 days 00:00:06  a
               0 days 00:00:06 0 days 00:00:07  a
               0 days 00:00:07 0 days 00:00:08  a
               0 days 00:00:08 0 days 00:00:09  a
               0 days 00:00:09 0 days 00:00:10  a
```

frankenjoe:
> BTW, in principle we could also support non hot encoded label, e.g.

I.e. you suggest to replace overlapping labels with `NaN`?

hagenw:
> > BTW, in principle we could also support non hot encoded label, e.g.
> 
> I.e. you suggest to replace overlapping labels with `NaN`?

There are two options:
* you take later (or maybe smaller) segment labels as corrections (e.g. showing `b` instead `a` for 0 days 00:00:03 0 days 00:00:04)
* you allow only one label per time, in this case we need to show `NaN` as we don't know if `a` or `b` is correct

hagenw:
But I don't know if we will ever need it. For me it would also be fine to just focus with `audbechmak.mapper.ToFramesWithOverlap` on hot encoded sequences and raise an error for string / integer / float labels.

Bahaabrougui:
@hagenw @frankenjoe then if the algorithm is to complicated, we can in the new method `ToFramesWithOverlap` just call the `ToFrames` method as much time as labels needed maybe? As this algorithm written in this MR does not include mapping to frames in it but just exploding the overlapping segments so we have the right boundaries.  
Btw, I'm on sickness leave this week, I'll get back next week and will move the PR to the appropriate repo

hagenw:
I agree that we should not add the solution here in `audformat` and will close this pull request.

Regarding `ToFramesWithOverlap()` I cannot really judge if this is needed or not. If users are able to create the correct dataframe without it, raising a warning in `ToFrames()` might be sufficient.

PR #162, opened 2022-04-01
Add utils.index_has_overlap()

Bahaabrougui:
image


hagenw:
Could you please split this pull request up into two as it is helpful and good practice to focus on one feature per pull request.

So it seems starting with "Add audformat.utils.iter_index_by_file()" would be a good start.

Bahaabrougui:
> Could you please split this pull request up into two as it is helpful and good practice to focus on one feature per pull request.
> 
> So it seems starting with "Add audformat.utils.iter_index_by_file()" would be a good start.

In this case do I have to implement a separate test for `iter_index_by_file()` too ? As the `has_overlap()` test covered it so far

hagenw:
yes, would be good to have separate tests.

audformat/utils/__init__.py
```diff
@@ -13,5 +14,5 @@
     replace_file_extension,
     to_filewise_index,
     to_segmented_index,
-    union,
+    union
```
frankenjoe, line 17:
```suggestion
    union,
```

audformat/core/utils.py
```diff
@@ -326,6 +326,34 @@ def expand_file_path(
...
+        index: pd.Index,
+) -> bool:
+    r"""Check if one or more segments overlap.
+
+    Args:
+        index: index object conform to
+            :ref:`table specifications <data-tables:Tables>`
+
+    Returns:
+        boolean
```
frankenjoe, line 339:
```suggestion
        ``True`` if overlap is detected, otherwise ``False``
```

audformat/core/utils.py
```diff
@@ -326,6 +326,34 @@ def expand_file_path(
     return index
 
 
+def has_overlap(
```
frankenjoe, line 329: Maybe this is a bit too generic, maybe we should change to `index_has_overlap`? @hagenw what do you think?

tests/test_utils.py
```diff
@@ -624,6 +624,28 @@ def test_expand_file_path(tmpdir, index, root, expected):
     pd.testing.assert_index_equal(expanded_index, expected)
 
 
+@pytest.mark.parametrize(
+    'index, expected',
+    [
+        (
+            audformat.segmented_index(['f1'] * 2, [0, 1], [2, 3]),
```
frankenjoe, line 631: Could you please add tests for an empty index and a segmented index with two files where the segments across files overlap but not within files (to make sure we only detect overlap within a file).

frankenjoe: And maybe also a test without overlap but where the segments touch each other (we expect `False' in that case).

audformat/utils/__init__.py
Bahaabrougui: Any particular reason why we add comma at the end ? Is it also some standardization ?

audformat/core/utils.py
Bahaabrougui: I agree, let's change it to that so it's more clear

audformat/utils/__init__.py
frankenjoe: Yes, if we have multiple arguments / elements etc. over multiple lines we always end with an `,` after the last entry.

audformat/core/utils.py
Bahaabrougui: Or, why don't we keep the name as it is but add support for pd.Series and pd.Dataframe, just like we did with `iter_by_file` ?

tests/test_utils.py
Bahaabrougui: > Could you please add tests for an empty index and a segmented index with two files where the segments across files overlap but not within files (to make sure we only detect overlap within a file).

Added.

> And maybe also a test without overlap but where the segments touch each other (we expect False in that case).

There is already a test case for that

audformat/core/utils.py
frankenjoe: When I see `audformat.utils.has_overlap(df)` it is not intuitive for a user to know what will happen. As a user I would probably assume it checks if values in the columns overlap and not in the index. So I would suggest to rename and only support index.

tests/test_utils.py
frankenjoe: We usually leave it to the reviewer to unresolve the issues.

audformat/core/utils.py
frankenjoe: If we force the user to do `audformat.utils.index_has_overlap(df.index)` the code will be longer, but it's also clear what the statement does.

Bahaabrougui: I added support for `Series` and `Dataframe`, if you think I should revert it to just `pd.Index` let me know

tests/test_utils.py
frankenjoe: > There is already a test case for that

Sorry, overlooked it.

> Added.

Just to be on the safe side, please add also one for an empty `filewise` index.

audformat/core/utils.py
frankenjoe: Ok, as long as we rename to `index_has_overlap()` it is probably ok (and convenient) to support `pd.Series` and `pd.DataFrame` as we do with other functions. Please extend the tests accordingly.

audformat/core/utils.py
```diff
@@ -326,6 +326,36 @@ def expand_file_path(
...
+        obj: typing.Union[pd.Index, pd.DataFrame, pd.Series],
+) -> bool:
+    r"""Check if one or more segments overlap.
```
frankenjoe, line 332:
```suggestion
    r"""Check if one or more segments in the index overlap.
```

Since we now support `Series` and `DataFrame` I think it's better to be more precise here.

docs/api-utils.rst
```diff
@@ -19,6 +19,11 @@ expand_file_path
 
 .. autofunction:: expand_file_path
 
+has_overlap
+___________
```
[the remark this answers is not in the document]
frankenjoe: Otherwise it will be treated as another sub level I think.

tests/test_utils.py
```diff
@@ -624,6 +624,52 @@ def test_expand_file_path(tmpdir, index, root, expected):
...
+            ),
+            False,
+        ),
+        (
+            pd.Series(
+                index=audformat.segmented_index(
+                    ['f1'] * 2,
+                    [0, 2],
+                    [2, 3],
+                )
```
frankenjoe, line 656:
```suggestion
                ),
```

tests/test_utils.py
```diff
@@ -624,6 +624,52 @@ def test_expand_file_path(tmpdir, index, root, expected):
...
+            ),
+            False,
+        ),
+        (
+            pd.DataFrame(
+                index=audformat.filewise_index(['f1'] * 2)
```
frankenjoe, line 662: I would prefer to keep the`filewise` index test as before and add one for `DataFrame` in addition.

tests/test_utils.py
Bahaabrougui: > We usually leave it to the reviewer to unresolve the issues.

Okay didn't know that

> Just to be on the safe side, please add also one for an empty filewise index.
I will

audformat/core/utils.py
Bahaabrougui: Oh the page didn't reload and didn't see your comments on not adding support for Dataframes and Series. But with the last comment I guess should be fine, I'll just rename it so it's more clear

tests/test_utils.py
Bahaabrougui: >  We usually leave it to the reviewer to unresolve the issues.

Just do double check, you mean resolve right ?

Bahaabrougui:
@frankenjoe, all suggestions are applied.

tests/test_utils.py
frankenjoe: Oh of course I mean "resolve" :) This way you give the reviewer the chance to check your proposed changes and if necessary complain again :)

frankenjoe: We usually still leave a comment for the reviewer like "suggestion applied" or something, so the reviewer knows that the issue has been tackled.

Bahaabrougui: > Oh of course I meant "resolve" :) This way you give the reviewer the chance to check your proposed changes and if necessary complain again :)

Haha okay, I get it :)

Bahaabrougui: suggestion applied :)

audformat/core/utils.py
```diff
@@ -353,6 +353,36 @@ def hash(
...
+        ``True`` if overlap is detected, otherwise ``False``
+
+    """
+
+    index = obj if isinstance(obj, pd.Index) else obj.index
+
+    if index_type(index) == define.IndexType.FILEWISE:
+        return False
+
+    index = to_segmented_index(index, allow_nat=False)
```
frankenjoe, line 375: Mhh, I'm not sure if we should force `allow_nat=False` here as it can be costly. Instead we should probably handle `NaT`. E.g.:

```
audformat.segmented_index(
    ['f1'] * 2,
    [0, 2],
    [1, NaT],
)
```

-> False

```
audformat.segmented_index(
    ['f1'] * 2,
    [0, 2],
    [NaT, 3],
)
```

-> True

```
audformat.segmented_index(
    ['f1'] * 2,
    [0, 2],
    [NaT, NaT],
)
```

-> True

Bahaabrougui: Hm why and how costly it is ?

frankenjoe: It will replace `NaT` with the duration of the file, so it has to get the duration for every file that is affected.

frankenjoe: In fact, it's not necessary to replace `NaT` with an exact value anyway. The result will be the same as long as we replace it with a large enough value. So I recommend to remove the line and instead insert:

```
ends = ends.fillna(pd.to_timedelta(sys.maxsize))
```

before the `if` condition within the loop.

Bahaabrougui: Yes I agree this is smarter way to do it. I applied the suggestions and added the adequate tests.

audformat/core/utils.py
```diff
@@ -372,11 +373,12 @@ def index_has_overlap(
     if index_type(index) == define.IndexType.FILEWISE:
         return False
 
-    index = to_segmented_index(index, allow_nat=False)
+    index = to_segmented_index(index)
```
frankenjoe, line 376: Can be removed as we already know that it must be a segmented index.

audformat/core/utils.py
Bahaabrougui: By the way from where can I get the knowledge of what's faster and what's not and these kind of optimizations when handling data structures ? I mean I find things on internet from time to time but maybe you have a better source

audformat/core/utils.py
Bahaabrougui: Yea already did, see latest commit

tests/test_utils.py
```diff
@@ -7,6 +7,8 @@
 import pytest
 
 import audeer
+from pandas._libs.tslibs.nattype import NaT
```
frankenjoe, line 10: I recommend to remove this important and use `pd.NaT` in the tests below.

audformat/core/utils.py
Bahaabrougui: What if the user passes a data frame that index ain't conform to audformat ? This keeps us in the safe side, or is it also a costly operation ?

frankenjoe: Not really, since `to_segmented_index()` also expects an index that is conform to audformat.

tests/test_utils.py
Bahaabrougui: Okay, why the recommendation tho ?

audformat/core/utils.py
frankenjoe: Any operation that needs to access information from disk becomes costly once you scale it up. Otherwise it's not always obvious what will be costly and what not. So usually the only way to find a bottleneck is by running your function on a large number of files / segments. With `audb` we also used a profiler at some point to nail down the expensive operations.

audformat/core/utils.py
Bahaabrougui: Ok, applied.

tests/test_utils.py
frankenjoe: 1. We already import `pandas`, so there is no need for an extra import
2. You are importing from a private module (`_libs`) here, which is not a good idea. They may change it in a future version, but `pd.NaT` will always be there.

audformat/core/utils.py
```diff
@@ -353,6 +354,36 @@ def hash(
...
+    r"""Check if one or more segments in the index overlap.
+
+    Args:
+        obj: object conform to
+            :ref:`table specifications <data-tables:Tables>`
+
+    Returns:
+        ``True`` if overlap is detected, otherwise ``False``
+
+    """
```
frankenjoe, line 369: I think we should add a docstring example.

audformat/core/utils.py
Bahaabrougui: Hm okay, a profiler is a tool that helps you simulate bottlenecks and gives you statistics and stuff right ? And what did you once you figured what are the expensive operations ? Some optimizations or just noting down that they should be avoided on large structures ?

tests/test_utils.py
Bahaabrougui: Okay understandable thanks, applied.

audformat/core/utils.py
frankenjoe: > Hm okay, a profiler is a tool that helps you simulate bottlenecks and gives you statistics and stuff right ?

Yes, a profiler shows you how long each function call takes and helps you to identify the critical calls

> And what did you once you figured what are the expensive operations?

Try to optimize you code by avoiding expensive operations :) E.g. here we now replace `NaT` with a large value which is much faster than accessing the actual file duration. But the result will be the same. These kind of things.

audformat/core/utils.py
Bahaabrougui: Added, and description above updated.

audformat/core/utils.py
Bahaabrougui: Okay perfect, thanks for explaining. Feel free to resolve the thread.

audformat/core/utils.py
frankenjoe: Cool, thanks. I have a slight suggestion though:

```
    Example:
        >>> index = filewise_index(['f1', 'f2'])
        >>> index_has_overlap(index)
        False
        >>> index = segmented_index(
        ...        ['f1', 'f2'],
        ...        [0, 1],
        ...        [2, 3],
        ...    )
        >>> index_has_overlap(index)
        False
        >>> index = segmented_index(
        ...        ['f1'] * 2,
        ...        [0, 1],
        ...        [2, 3],
        ...    )
        >>> index_has_overlap(index)
        True
```

Why? We now start with the simplest example and move on to advanced ones.

audformat/core/utils.py
```diff
@@ -353,6 +354,48 @@ def hash(
...
+        ...        [2, 3],
+        ...    )
+        >>> index_has_overlap(index)
+        True
+        >>> index = filewise_index(['f1', 'f2'])
+        >>> index_has_overlap(index)
+        False
+
+    """
+
```
frankenjoe, line 382: We usually don't start with a blank line after the docstring.

audformat/core/utils.py
Bahaabrougui: Make sense, applied.

frankenjoe: Thanks, could you please update the description then we are good to go I think.

frankenjoe: Please also update the title to `Add utils.index.has_overlap()` (no need to mention that we add tests).

Bahaabrougui: > Thanks, could you please update the description then we are good to go I think.

Already did

> Please also update the title to Add utils.index_has_overlap() (no need to mention that we add tests).

Done now

frankenjoe: > Already did

It still shows the old example, but ok, not so important.

frankenjoe:
@hagenw no more comments from my side

audformat/core/utils.py
Bahaabrougui: Weird, are you talking about the PR description? Because it's updated for me 

image

frankenjoe: Ok, now for me, too. Maybe it was just a caching issue. Looks great now!

hagenw:
If you have an index with the same files, results are not as expected:

```python
>>> index = audformat.filewise_index(['f1', 'f1'])
>>> audformat.utils.index_has_overlap(index)
False
```

frankenjoe:
> If you have an index with the same files, results are not as expected:

Well, one could argue such an index is not conform to audformat as we don't allow duplicates.

hagenw:
> > If you have an index with the same files, results are not as expected:
> 
> Well, one could argue such an index is not conform to `audformat` as we don't allow duplicates.

That's true, but as we don't raise an error it's still risky that it returns the wrong value.

frankenjoe:
> That's true, but as we don't raise an error it's still risky that it returns the wrong value.

We have [assert_no_duplicates()](https://audeering.github.io/audformat/api.html#assert-no-duplicates), but at some point decided it to leave it to the user to call it since it was too expensive to call it in every function. So I don't think we should start calling it again in this function.

hagenw:
> > That's true, but as we don't raise an error it's still risky that it returns the wrong value.
> 
> We have [assert_no_duplicates()](https://audeering.github.io/audformat/api.html#assert-no-duplicates), but at some point decided it to leave it to the user to call it since it was too expensive to use it in every function. So I don't think we should start calling it again in this function.

In this case we only need to check for duplicates in filewise index and can return a `True` as soon as we find one duplicated entry, do you think there is no fast way to check this?
Another solution would be to say that for a filewise index it always returns `False` as there are no segments in a filewise index.

You also state in the docstring at the moment:
"Check if one or more segments in the index overlap."

So maybe just add another sentence, that for a filewise index it always returns `False`.

frankenjoe:
> So maybe just add another sentence, that for a filewise index it always returns `False`.

Yes, let's do that.

Bahaabrougui:
> > > That's true, but as we don't raise an error it's still risky that it returns the wrong value.
> > 
> > 
> > We have [assert_no_duplicates()](https://audeering.github.io/audformat/api.html#assert-no-duplicates), but at some point decided it to leave it to the user to call it since it was too expensive to use it in every function. So I don't think we should start calling it again in this function.
> 
> In this case we only need to check for duplicates in filewise index and can return a `True` as soon as we find one duplicated entry, do you think there is no fast way to check this? Another solution would be to say that for a filewise index it always returns `False` as there are no segments in a filewise index.
> 
> You also state in the docstring at the moment: "Check if one or more segments in the index overlap."
> 
> So maybe just add another sentence, that for a filewise index it always returns `False`.

I've updated the description.

audformat/core/utils.py
```diff
@@ -353,6 +354,56 @@ def hash(
     return str(pd.util.hash_pandas_object(obj).sum())
 
 
+def index_has_overlap(
+        obj: typing.Union[pd.Index, pd.DataFrame, pd.Series],
+) -> bool:
+    r"""Check if one or more segments in the index overlap.
+
+    If the index is filewise, the return value will always be False.
```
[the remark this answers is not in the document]
frankenjoe: > Does it mean I have to unresolve it after it happens automatically ?

If you apply a suggestion by the reviewer directly, there is no need to keep it open.

frankenjoe:
Since it references an internal project, I suggest we remove this from the description:

image

Bahaabrougui:
> Since it references an internal project, I suggest we remove this from the description:
> 
> image

Ah true forget this was supposed to be opensource :D

hagenw:
Great, and thanks for updating this.

BTW, the corresponding segmented index works, which is nice to have:

```python
>>> index = audformat.filewise_index(['f1', 'f1'])
>>> index = audformat.utils.to_segmented_index(index)
>>> index
MultiIndex([('f1', '0 days', NaT),
            ('f1', '0 days', NaT)],
>>> audformat.utils.index_has_overlap(index)
True
>>> index = audformat.filewise_index(['f1', 'f2'])
>>> index = audformat.utils.to_segmented_index(index)
>>> index
MultiIndex([('f1', '0 days', NaT),
            ('f2', '0 days', NaT)],
>>> audformat.utils.index_has_overlap(index)
False
```

PR #164, opened 2022-04-01
Add iter_index_by_file() as utility function.

Bahaabrougui:
Adds the function iter_index_by_file. Doesn't implement separate test for it.
If we merge this first, merging this https://github.com/audeering/audformat/pull/163 next should be fine. Or should I close the previous PR and only wait until this one is merged, and open a new PR from the new Master ?

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1074,3 +1074,24 @@ def union(
     index = index.drop_duplicates()
 
     return index
+
+
+def iter_index_by_file(
+        index: pd.Index
+) -> (str, typing.Iterator[pd.Index]):
+    r"""Iterate over index by file.
+
+    In each iteration return sub-index for one file.
+
+    Args:
+        index: index object conform to
+            :ref:`table specifications <data-tables:Tables>`
+
+    Returns:
+        iterator in form of (file, sub_index)
+    """
+    index = to_segmented_index(index, allow_nat=False)
+    files = index.get_level_values('file').drop_duplicates()
+    for file in files:
+        select = index[index.get_loc(file)]
+        yield file, select


audformat/core/utils.py
```diff
@@ -1074,3 +1074,24 @@ def union(
...
+    r"""Iterate over index by file.
+
+    In each iteration return sub-index for one file.
+
+    Args:
+        index: index object conform to
+            :ref:`table specifications <data-tables:Tables>`
+
+    Returns:
+        iterator in form of (file, sub_index)
```
hagenw, line 1091: Could you please add an docstring example for this.
That might then even cover the code already.

hagenw: Or in code https://github.com/audeering/audformat/blob/eeb55de67a1d6fd06e448117f40e8ecda9cbc029/audformat/core/utils.py#L258-L267

Bahaabrougui: I've added the docstring example, and the code is covered as you said

hagenw:
> > Adds the function iter_index_by_file. Doesn't implement separate test for it.
> 
> I'd prefer we add tests for just this function in this PR.

Yes, that should be in general the goal for pull requests. Add a single feature (e.g. class/method/function) including tests and documentation.

audformat/core/utils.py
```diff
@@ -1074,3 +1074,24 @@ def union(
...
+    """
+    index = to_segmented_index(index, allow_nat=False)
```
frankenjoe, line 1100: First of all, I don't think we should force `allow_nat=False` here by default. It can be a costly. But I'm actually not sure if we should convert to a segmented index at all here, bust instead add support for `filewise` indices. It does not make so much sense at the moment, since we currently don't allow duplicates, so it will return exactly one file in each iteration, but maybe we will allow duplicates in the future again. Generally, I'd like to leave it to the user to convert to a segmented index if needed.

hagenw:
You need to include the added function to the API, which means listing it inside `audformat/utils/__init__.py` and `docs/api-utils.rst` in an alphabetically ordered way.

Bahaabrougui:
> You need to include the added function to the API, which means listing it inside `audformat/utils/__init__.py` and `docs/api-utils.rst` in an alphabetically ordered way.

My bad forgot about  `audformat/utils/__init__.py`. But `docs/api-utils.rst`, I didn't knew about it

audformat/core/utils.py
```diff
@@ -1074,3 +1074,31 @@ def union(
...
+        >>> print(next(iter_index_by_file(i)))
+        ('f1', MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:02'),
+            ('f1', '0 days 00:00:01', '0 days 00:00:03')],
+           names=['file', 'start', 'end']))
+    """
+    index = to_segmented_index(index, allow_nat=False)
+    files = index.get_level_values('file').drop_duplicates()
+    for file in files:
+        select = index[index.get_loc(file)]
+        yield file, select
```
frankenjoe, line 1104: I'm not so sure if we should really return `file` and `select` as separate values. I know, this is my own implementation, but we know deal with the official API and not just some custom function used in a specific context. What is not so nice here is that `select' is not confirm to `audformat` since it has only `start` and `end`, which we do not officially support, i.e. other utility function will not work. @hagenw what is your take here?

hagenw:
> > You need to include the added function to the API, which means listing it inside `audformat/utils/__init__.py` and `docs/api-utils.rst` in an alphabetically ordered way.
> 
> My bad forgot about `audformat/utils/__init__.py`. But `docs/api-utils.rst`, I didn't knew about it

Compare http://devops.pp.audeering.com/python/documentation.html#api-documentation

audformat/core/utils.py
hagenw: I would prefer that it simply returns a segmented index containing a single file and all segments.
For filewise, I'm not sure if it should return a filewise index, or also a segmented index.

Can you provide possible applications of this utility function?

hagenw: As the name of the function and its description only indicates the iteration, I would expect that it returns a filewise index with a single file when feeding it with a filewise index.

Bahaabrougui: When I was testing the function offline trying to see what it does, select is actually a segmented index that is conform to `audformat` (see example in the docstring). Isn't that intended ?

audformat/core/utils.py
hagenw: Thanks.

audformat/core/utils.py
frankenjoe: One is the overlap scenario, of course, when we look for overlaps we only want to consider segments belonging to the same file. Another one might be an evaluation per file:

```python
for index_per_file in iter_index_by_file(segmented_index):
    pred = model.process_index(index_per_file)
    truth = truths.loc(index_per_file)
    ...
```

If we ever allow duplicates in indices again, it would be handy to select all labels that belong to the same files, e.g.

```python
for files in index_per_file(filewise_index):
   labels_belonging_to_the_same_file = df.loc[index_per_file]
```

> For filewise, I'm not sure if it should return a filewise index, or also a segmented index.

My motivation to not convert to a segmented index here is that then we don't have to care about `allow_nat` argument and `num_workers` argument, which we otherwise have to pass on I guess.

frankenjoe: Mhh, yes probably we should, as we already have it. Otherwise the user has to do 'index.get_level_values('file')[0]` to get it. So I recommend we return the file name and an index, but the index should be conform to `audformat`, i.e. include the `file` level.

frankenjoe: But according to https://github.com/audeering/audformat/pull/164#discussion_r840581711 this is already the case :)

Bahaabrougui: Then if we want to consider all cases, let's return a Filewise index per file if the index object is a filewise index as @hagenw said. If not we add the allow_nat as an argument, and based on that we either apply the changes and return (file, segmented_index) or we return (file, segmented_index) as it is with nat in the indexes

hagenw: Agree, in the example the `file` level is included in the returned index, so that looks fine to me.

To also avoid the `aloow_nat` problem I would then simply return a filewise index for filewise input.

frankenjoe: > But adding allow_nat as an argument to iter_index_by_file isn't the same as leaving it to the user ? I mean we're just allowing more flexibility in case the user forgets that he needs it :x

I'm in favor of having as less arguments as possible. And especially avoid arguments that only apply in some cases. Cause if the input is filewise, then `allow_nat` has no effect at all. + if we add `allow_nat` then would also need to add `num_workers`.

Bahaabrougui: Mhm okay then, let's keep it simple

frankenjoe: The only argument for adding it, would be if the algorithm requires that we do not have `NaT` values. But this is not the case here.

Bahaabrougui:
Have a look at the new docstring, and tell me if this is the expected behaviour :  
```python
Examples
        >>> index = segmented_index(['f1', 'f1', 'f2'], [0, 1, 0], [2, 3, 1])
        >>> print(next(iter_index_by_file(index)))
        ('f1', MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:02'),
            ('f1', '0 days 00:00:01', '0 days 00:00:03')],
           names=['file', 'start', 'end']))
        >>> index = filewise_index(['f1', 'f1', 'f2'])
        >>> print(next(iter_index_by_file(index)))
        ('f1', Index(['f1'], dtype='object', name='file'))
```

frankenjoe:
I suggest we rename to `iter_by_file()` and add support for `pd.Series` and `pd.DataFrame`, e.g.:

```python
def iter_by_file(
        obj: typing.Union[
            pd.Index,
            pd.Series,
            pd.DataFrame,
        ],
) -> (str, typing.Iterator[typing.Union[pd.Index, pd.Series, pd.DataFrame]]):
    
    is_index = isinstance(obj, pd.Index)
    index = obj if is_index else obj.index

    files = index.get_level_values('file').drop_duplicates()
    if index_type(index) == define.IndexType.FILEWISE:
        for file in files:
            sub_index = filewise_index(file)
            sub_obj = sub_index if is_index else obj.loc[sub_index]
            yield file, sub_obj
    else:
        for file in files:
            sub_index = index[index.get_loc(file)]
            sub_obj = sub_index if is_index else obj.loc[sub_index]
            yield file, sub_obj

**Comment by Bahaabrougui** [comment]
> I suggest we rename to `iter_by_file()` and add support for `pd.Series` and `pd.DataFrame`, e.g.:
> 
> ```python
> def iter_by_file(
>         obj: typing.Union[
>             pd.Index,
>             pd.Series,
>             pd.DataFrame,
>         ],
> ) -> (str, typing.Iterator[typing.Union[pd.Index, pd.Series, pd.DataFrame]]):
>     
>     is_index = isinstance(obj, pd.Index)
>     index = obj if is_index else obj.index
> 
>     files = index.get_level_values('file').drop_duplicates()
>     if index_type(index) == define.IndexType.FILEWISE:
>         for file in files:
>             sub_index = filewise_index(file)
>             sub_obj = sub_index if is_index else obj.loc[sub_index]
>             yield file, sub_obj
>     else:
>         for file in files:
>             sub_index = index[index.get_loc(file)]
>             sub_obj = sub_index if is_index else obj.loc[sub_index]
>             yield file, sub_obj
> ```



> I suggest we rename to `iter_by_file()` and add support for `pd.Series` and `pd.DataFrame`, e.g.:
> 
> ```python
> def iter_by_file(
>         obj: typing.Union[
>             pd.Index,
>             pd.Series,
>             pd.DataFrame,
>         ],
> ) -> (str, typing.Iterator[typing.Union[pd.Index, pd.Series, pd.DataFrame]]):
>     
>     is_index = isinstance(obj, pd.Index)
>     index = obj if is_index else obj.index
> 
>     files = index.get_level_values('file').drop_duplicates()
>     if index_type(index) == define.IndexType.FILEWISE:
>         for file in files:
>             sub_index = filewise_index(file)
>             sub_obj = sub_index if is_index else obj.loc[sub_index]
>             yield file, sub_obj
>     else:
>         for file in files:
>             sub_index = index[index.get_loc(file)]
>             sub_obj = sub_index if is_index else obj.loc[sub_index]
>             yield file, sub_obj
> ```

@frankenjoe which example/situation made you think that sometimes it's better to have  a generator returning (file, sub_obj) and not (file, sub_index) ?

**Comment by frankenjoe** [comment]
> @frankenjoe which example/situation made you think that sometimes it's better to have a generator returning (file, sub_obj) and not (file, sub_index) ?

If the input is an index we return an index, but if we get a series / frame, we return a series / frame instead.

**Comment by frankenjoe** [comment]
Could you please add unit tests for the new function?

**Comment by Bahaabrougui** [comment]
> > @frankenjoe which example/situation made you think that sometimes it's better to have a generator returning (file, sub_obj) and not (file, sub_index) ?
> 
> If the input is an index we return an index, but if we get a series / frame, we return a series / frame instead.

Yes yes I understand that, I was curious which specific use case to your mind to think we should add it, but if it's just to be more general and flexible then never mind

**Comment by Bahaabrougui** [comment]
> Could you please add unit tests for the new function?

Ok, although doesn't the examples in docstring count as "tests" and cover for it as I understood from Hagen or it's not enough ?

**Comment by frankenjoe** [comment]
> cover for it as I understood from Hagen or it's not enough ?

If it's a simple function, then the docstring example can be enough. But in this case we should add unit tests to to make sure we cover all corner cases. In particular, input is empty and input is a index, series or frame (the docstring only covers index at the moment).

**Comment by Bahaabrougui** [comment]
I've wrote this test samples (haven't wrote the function code yet)
```python

@pytest.mark.parametrize(
    'obj, expected',
    [
        (
            
        ),
        (
            audformat.segmented_index(['f1', 'f1', 'f2'], [0, 1, 0], [2, 3, 1]),
            ('f1', audformat.segmented_index(['f1', 'f1'], [0, 1], [2, 3]))
        ),
        (
            pd.Series(
                index=audformat.segmented_index(['f1', 'f1', 'f2'], [0, 1, 0], [2, 3, 1]),
            ),
            ('f1', pd.Series(
                index=audformat.segmented_index(['f1', 'f1'], [0, 1], [2, 3]))
             )
        ),
        (
            pd.DataFrame(
                index=audformat.filewise_index(['f1', 'f1', 'f2'])
            ),
            ('f1', pd.DataFrame(
                index=audformat.filewise_index(['f1', 'f1']))
             )
        ),
    ]
)
def test_iter_by_file(obj, expected):
    pass
```
As for the first example where the input is empty, empty input can mean either input is `None` or input is an empty index e.g `audformat.filewise_index()`; and in both case the `iter_by_file()` func will throw an exception, in the first case it will be `None object has no attribute index`; and in the next case it will be `StopIteration` as there is nothing to iterate on. In such scenarios, should we treat these cases explicitly in the function `iter_by_file()`, or the exception can actually be the expected behaviour ?

frankenjoe:
> As for the first example where the input is empty, empty input can mean either input is `None` or input is an empty index e.g `audformat.filewise_index()`

You don't have to test for `None` since the typestring does not allow it (otherwise we would set it to `typing.Optional[...]`). But an empty index / series / frame is a valid input and we don't want to raise an exception. i guess there most be some way to return an empty iterator or something.

Bahaabrougui:
> > As for the first example where the input is empty, empty input can mean either input is `None` or input is an empty index e.g `audformat.filewise_index()`
> 
> You don't have to test for `None` as it is not supported by the type string (otherwise we would set it to `typing.Optional[...]`). But an empty index / series / frame is a valid input and we don't want to raise an exception. i guess there most be some way to return an empty iterator or something.

But as far as I know, the typing is just for documentation, and the function will accept a `None` object if you pass to it, or are you just referring to documentation wise ? Meaning if someone read the documentation will know that `None` is not supported ?

audformat/core/utils.py
```diff
@@ -1088,17 +1088,24 @@ def iter_index_by_file(
             :ref:`table specifications <data-tables:Tables>`
 
     Returns:
-        iterator in form of (file, sub_index)
+        iterator in form of (file, sub_index) if the index is segmentwise
+        otherwise (file, filewise_index(file))
```
frankenjoe: I would not make a distinction between segmented and filewise index here, saying `iterator in form of (file, sub_index)` should be fine

audformat/core/utils.py
```diff
@@ -1074,3 +1074,52 @@ def union(
...
+    Args:
+        obj: Can be a DataFrame or Series that index is conform to
+            :ref:`table specifications <data-tables:Tables>`, or just
+            the index object itself.
+
+    Returns:
+        If obj is an index, the return value will be an iterator in form of
+        (file, sub_index) if the index is segmentwise otherwise
+        (file, filewise_index(file)). If obj is a Series or a Dataframe, the
+        return value will be an iterator in form of (file, sub_obj).
```
frankenjoe, line 1099: Or actually I think we can simplify to: 

```
Returns:
    iterator in form of (file, sub_obj)
```

audformat/utils/__init__.py
```diff
@@ -13,4 +13,5 @@
     to_filewise_index,
     to_segmented_index,
     union,
+    iter_by_file
```
frankenjoe, line 16: Please keep the alphabetic order.

tests/test_utils.py
```diff
@@ -1597,3 +1597,61 @@ def test_union(objs, expected):
...
         expected,
     )
+
+
+@pytest.mark.parametrize(
+    'obj, expected',
+    [
+        (
+            pd.Series(),
+            []
```
frankenjoe, line 1607:
```suggestion
            [],
```

We always end multi-line arguments with a comma after the last entry. Please also add a comma in the other cases.

frankenjoe: Actually, to match how we usually format our code, please change to:

```python
@pytest.mark.parametrize(
    'obj, expected',
    [
        (
            pd.Series(),
            [],
        ),
        (
            audformat.filewise_index(),
            [],
        ),
        (
            audformat.segmented_index(
                ['f1', 'f1', 'f2'],
                [0, 1, 0],
                [2, 3, 1],
            ),
            [
                (
                    'f1', 
                    audformat.segmented_index(
                        ['f1', 'f1'], 
                        [0, 1],
                        [2, 3],
                    ),
                ),
                (
                    'f2', 
                    audformat.segmented_index(['f2'], [0], [1]),
                ),
            ],
        ),
        (
            pd.Series(
                index=audformat.segmented_index(
                    ['f1', 'f1', 'f2'],
                    [0, 1, 0], 
                    [2, 3, 1],
                ),
            ),
            [
                (
                    'f1', 
                    pd.Series(
                        index=audformat.segmented_index(
                            ['f1', 'f1'],
                            [0, 1], 
                            [2, 3],
                        ),
                    ),
                ),
                (
                    'f2', 
                    pd.Series(
                        index=audformat.segmented_index(
                            ['f2'],
                            [0], 
                            [1],
                        ),
                    ),
                ),
            ],
        ),
        (
            pd.DataFrame(
                index=audformat.filewise_index(['f1', 'f1', 'f2']),
            ),
            [
                (
                    'f1', 
                    pd.DataFrame(
                        index=audformat.filewise_index(['f1', 'f1']),
                    ),
                ),
                (
                    'f2', 
                    pd.DataFrame(
                        index=audformat.filewise_index(['f2']),
                    ),
                ),
            ],
        ),
    ]
)
```

tests/test_utils.py
```diff
@@ -1597,3 +1597,61 @@ def test_union(objs, expected):
         audformat.utils.union(objs),
         expected,
     )
+
+
+@pytest.mark.parametrize(
+    'obj, expected',
+    [
+        (
+            pd.Series(),
```
frankenjoe, line 1606: Please also add a test for an empty `pd.DataFrame`.

audformat/core/utils.py
```diff
@@ -1074,3 +1074,52 @@ def union(
...
+def iter_by_file(
+        obj: typing.Union[
+            pd.Index,
+            pd.Series,
+            pd.DataFrame,
+        ],
+) -> (str, typing.Iterator[typing.Union[pd.Index, pd.Series, pd.DataFrame]]):
+    r"""Iterate over index by file.
+
+    In each iteration return sub-index for one file.
```
frankenjoe, line 1088:
```suggestion
    In each iteration return sub-object for one file.
```

audformat/core/utils.py
```diff
@@ -1074,3 +1074,52 @@ def union(
...
+
+    Args:
+        obj: Can be a DataFrame or Series that index is conform to
```
frankenjoe, line 1091:
```suggestion
        obj: object conform to
```

frankenjoe: And I would remove the "or just the index object itself."

docs/api-utils.rst
```diff
@@ -29,6 +29,10 @@ intersect
 
 .. autofunction:: intersect
 
+iter_by_file
+__________________
```
frankenjoe, line 33:
```suggestion
------------
```

audformat/core/utils.py
```diff
@@ -1074,3 +1074,52 @@ def union(
...
     return index
+
+
+def iter_by_file(
+        obj: typing.Union[
+            pd.Index,
+            pd.Series,
+            pd.DataFrame,
+        ],
+) -> (str, typing.Iterator[typing.Union[pd.Index, pd.Series, pd.DataFrame]]):
```
frankenjoe, line 1085:
```suggestion
) -> typing.Iterator[
    typing.Tuple[
        str,
        typing.Union[pd.Index, pd.Series, pd.DataFrame],
    ],
]:
```

audformat/core/utils.py
```diff
@@ -1074,3 +1074,52 @@ def union(
...
+            :ref:`table specifications <data-tables:Tables>`, or just
+            the index object itself.
+
+    Returns:
+        If obj is an index, the return value will be an iterator in form of
+        (file, sub_index) if the index is segmentwise otherwise
+        (file, filewise_index(file)). If obj is a Series or a Dataframe, the
+        return value will be an iterator in form of (file, sub_obj).
+
+    Examples
```
frankenjoe, line 1101:
```suggestion
    Example:
```

audformat/core/utils.py
```diff
@@ -1074,3 +1074,52 @@ def union(
...
+        >>> index = segmented_index(['f1', 'f1', 'f2'], [0, 1, 0], [2, 3, 1])
```
frankenjoe, line 1102: I would suggest start with the filewise index and show the segmented index afterwards. And maybe also add an example for a non-index input, e.g.:

```
        >>> index = filewise_index(['f1', 'f1', 'f2'])
        >>> next(iter_by_file(index))
        ('f1', Index(['f1'], dtype='object', name='file'))
        >>> index = segmented_index(['f1', 'f1', 'f2'], [0, 1, 0], [2, 3, 1])
        >>> next(iter_by_file(index))
        ('f1', MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:02'),
            ('f1', '0 days 00:00:01', '0 days 00:00:03')],
           names=['file', 'start', 'end']))
        >>> obj = pd.Series(['a', 'b', 'b'], index)
        >>> next(iter_by_file(obj))
        ('f1', file  start            end
        f1    0 days 00:00:00  0 days 00:00:02    a
              0 days 00:00:01  0 days 00:00:03    b
        dtype: object)
```

Bahaabrougui:
> Sorry for the spam of suggestions :) Most of them are just formatting issues, but it would be good if you could adapt to our style of coding that we use across all our packages.
> 
> Btw: I saw the Windows test failing in other projects, too. It's related to `sox` and we need to find out what's going on. You can ignore it for the moment.

Okay I addressed the comments. And no worries, I wouldn't learn and adapt if you guys don't spam :)

audformat/core/utils.py
```diff
@@ -1074,3 +1074,61 @@ def union(
     index = index.drop_duplicates()
 
     return index
+
+
+def iter_by_file(
```
frankenjoe, line 1079: Could you please move this up to fit the alphabetical order?

audformat/core/utils.py
```diff
@@ -1074,3 +1074,61 @@ def union(
...
+        ],
+) -> typing.Iterator[
+    typing.Tuple[
+        str,
+        typing.Union[pd.Index, pd.Series, pd.DataFrame],
+    ],
+]:
+    r"""Iterate over object by file.
+
+    Each iteration returns a sub-object per file.
```
frankenjoe, line 1093:
```suggestion
    Each iteration returns a file and the according sub-object.
```

frankenjoe:
> Okay I addressed the comments. And no worries, I wouldn't learn and adapt if you guys don't spam :)

Thanks. I had two more comments, but then we are good to go I think.

audformat/core/utils.py
```diff
@@ -438,6 +438,64 @@ def intersect(
...
+        typing.Union[pd.Index, pd.Series, pd.DataFrame],
+    ],
+]:
+    r"""Iterate over object by file.
+
+    Each iteration returns a file and the according sub-object.
+
+    Args:
+        obj: object conform to
+            :ref:`table specifications <data-tables:Tables>`.
```
frankenjoe, line 459:
```suggestion
            :ref:`table specifications <data-tables:Tables>`
```

We usually don't end with a dot here

frankenjoe:
Thanks, looks good now. After applying my (really) last comment, could you please update name / description of the PR. Best practice: post a screenshot of the new function in the API.

Bahaabrougui:
> Thanks, looks good now. After applying my (really) last comment, could you please update name / description of the PR. Best practice: post a screenshot of the new function in the API.

I'm having some package problems compiling the documentation (coming from Graphviz not found in path). I'll post a screenshot of the API as soon as I find a fix

Bahaabrougui:
> > Thanks, looks good now. After applying my (really) last comment, could you please update name / description of the PR. Best practice: post a screenshot of the new function in the API.
> 
> I'm having some package problems compiling the documentation (coming from Graphviz not found in path). I'll post a screenshot of the API as soon as I find a fix

Okay should be fine now, the screenshot is in the description

frankenjoe:
@hagenw tests are passing again, please have a look if you agree with the PR. I don't have any more comments at the moment.

hagenw:
Cool, thanks for the effort. Looks all good from my side.

PR #165, opened 2022-04-12
WIP: CI: use conda to install sox

hagenw:
This uses conda to install sox on all platforms as it was failing with choco under Windows.

As an alternative we can also still wait for a new release of `audiofile` and remove the installation of `sox` altogether.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,19 +35,24 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Set up conda
+      uses: s-weigand/setup-conda@v1
+      with:
+        activate-conda: false  # don't switch Python version
+
+    - name: Install sox using conda
+      # MP3 support under Linux and macOS, see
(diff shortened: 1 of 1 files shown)


hagenw:
At the moment we only need `sox` to avoid a failing test when building the documentation (compare https://github.com/audeering/audformat/pull/166) as `audiofile` throws a warning when running `import sox` without an available binary. I think this could be fixed inside `audiofile` by moving the `import sox` statement inside the `try` statement.

frankenjoe:
> I think this could be fixed inside `audiofile` by moving the `import sox` statement inside the `try` statement.

Yes, I already to suggest this, too. Would be nice if we get rid of the message.

frankenjoe:
> As an alternative we can also still wait for a new release of `audiofile` and remove the installation of `sox` altogether.

Yes, I think that makes sense, otherwise we will have to add the `conda` workaround in almost all our packages now.

hagenw:
Closing in favor of #167

PR #166, opened 2022-04-12
WIP: CI: don't install sox

hagenw:
Just a test to see what happens if we don't install `sox`.


hagenw:
Closing in favor of #167

PR #168, opened 2022-04-14
Allow pandas>=1.4.x

### LINKED ISSUE #169: Figure out why `pd.read_csv()` fails with `pandas >= 1.4.0`.
See https://github.com/audeering/audformat/pull/142#issuecomment-1020175759 and https://github.com/audeering/audformat/pull/168

hagenw:
Closes #169

With #171 we can add support for pandas >= 1.4.0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ install_requires =
     pyyaml >=5.4.1
     # https://github.com/audeering/audformat/pull/101
     # https://github.com/audeering/audformat/pull/142
-    pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3,!=1.4.0
+    pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3
 setup_requires =
     setuptools_scm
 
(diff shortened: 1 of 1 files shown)


hagenw:
Seems like our code does still not work with `pandas>=1.4.0` (it installed 1.4.2 in the CI job). So maybe we should open an issue here pointing out, that we need to fix this maybe ourselves?

frankenjoe:
Do you know where we initially discussed why it is failing. I think it was related to the `DataFrame.to_csv()` function.

hagenw:
> Do you know where we initially discussed why it is failing. I think it was related to the `DataFrame.to_csv()` function.

https://github.com/audeering/audformat/pull/142#issuecomment-1020186163

frankenjoe:
Thanks, so it's actually the `read_csv()` function causing the issues. So yes we probably have to take a closer look why it happens, maybe they changed a default argument or something like this.

hagenw:
> Thanks, so it's actually the `read_csv()` function causing the issues. So yes we probably have to take a closer look why it happens, maybe they changed a default argument or something like this.

Can you open an issue so we don't forget it.

frankenjoe:
https://github.com/audeering/audformat/issues/169

frankenjoe:
A fix is provided in #170

frankenjoe:
@hagenw tests are passing! Please assign to me and I will merge.

hagenw:
Great

PR #170, opened 2022-04-14
WIP: Support pandas 1.4.x

### LINKED ISSUE #169: Figure out why `pd.read_csv()` fails with `pandas >= 1.4.0`.

frankenjoe:
Closes #169 

Since pandas 1.4.0 `DataFrame.to_csv()` no longer works for categories with dtype `Int64` so we have to temporarily convert to plain `Int64`. With older pandas versions we don't need the workaround so we have to exclude it from code coverage for the moment.

At least they also fixed a bug where categories of type `Int64` were converted to `int64` when saving to pickle (maybe removing this bug is related to the introduction of the new bug). Since we still need to support older pandas versions because of Python 3.6, I removed our workaround from code coverage.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1147,7 +1147,10 @@ class Table(HeaderBase):
         except pickle.UnpicklingError:
             df = pd.read_pickle(path, compression='xz')
 
-        for column_id in df:
+        for column_id in df:  # pragma: no cover
+            # TODO: Fixed with pandas 1.4.0 and can be removed
+            #   once we drop support for Python 3.6
+
             # Categories of type Int64 are somehow converted to int64.
             # We have to change back to Int64 to make column nullable.
             if isinstance(df[column_id].dtype,
@@ -1169,7 +1172,18 @@ class Table(HeaderBase):
         # that is newer than the PKL file
         df = self.df
         with open(path, 'w') as fp:
+            # Since pandas 1.4.0 DataFrame.to_csv()
+            # no longer works for categories with dtype Int64
+            # so we have to convert column to plain Int64
+            tmp_cols = {}
+            for column, dtype in zip(df, df.dtypes):
+                if dtype.name == 'category' and \
+                        dtype.categories.dtype.name == 'Int64':
+                    tmp_cols[column] = df[column]
+                    df[column] = df[column].astype('Int64')
             df.to_csv(fp, encoding='utf-8')
+            for column, y in tmp_cols.items():
+                df[column] = y
 
     def _save_pickled(self, path: str):
         self.df.to_pickle(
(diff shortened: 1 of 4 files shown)


frankenjoe:
@hagenw It's not very nice we have to exclude two spots from code coverage now, do you see a better solution?

audformat/core/table.py
```diff
@@ -1169,8 +1172,25 @@ def _save_csv(self, path: str):
...
         df = self.df
         with open(path, 'w') as fp:
+            # TODO: Include into code coverage
+            #   once we drop support for < Python 3.8
+
+            tmp_cols = {}
+            for column, dtype in zip(df, df.dtypes):  # pragma: no cover
+                # Since pandas 1.4.0 DataFrame.to_csv()
+                # no longer works for categories with dtype Int64
+                # we have to temporarily convert column to plain Int64
```
hagenw, line 1182: What's the difference between `dtype Int64` and `plain Int64`?

docs/data-example.rst
```diff
@@ -21,7 +21,12 @@ and as CSV:
 .. jupyter-execute::
     :hide-code:
 
-    print(db['files'].get().to_csv())
+    df = db['files'].get()
+    # Since pandas 1.4.0 DataFrame.to_csv()
+    # no longer works for categories with dtype Int64
+    # we have to convert column to plain Int64
+    df['label_map_int'] = df['label_map_int'].astype('Int64')
```
hagenw, line 28: This doesn't look to me like something a user should have to care about.

audformat/core/table.py
frankenjoe: We convert a column with dtype "categories with dtype Int64" to a column with "dtype Int64".

docs/data-example.rst
hagenw: Maybe we should first open an issue at `pandas` and ask if the behavior is desired or a bug?

frankenjoe: A user will call `db.save()` to save the whole database or `db['table'].save()` to save a particular table.

hagenw: OK, I now realized that this code is not visible for the user, and we just need it to show the CSV output:

image

So, it's still nasty that we need this workaround, but it should be fine as the user never has to call this.

hagenw:
Regarding the code coverage problem: it should be possible to collect reports over different Python versions and combine them, e.g. https://coverage.readthedocs.io/en/stable/cmd.html#cmd-combine
This would be one solution to reach 100% without removing parts of the code from coverage.
What I don't know yet is how this would affect running it locally.

frankenjoe:
> Maybe we should first open an issue at pandas and ask if the behavior is desired or a bug?

Yes, independent of this PR we should do that. It's definitely a bug. I'll prepare a minimal example that fails first and post it here.

frankenjoe:
While preparing a minimal example that fails with `Int64` I realized we don't need `Int64` but can simply store integer categories as `int64`. Created an alternative PR: https://github.com/audeering/audformat/pull/171

frankenjoe:
Closing in favor of #171

PR #171, opened 2022-04-19
Store category with integers as int64 instead of Int64

frankenjoe:
In #170 we propose a workaround for storing categories with integer values as dtype `Int64` (it's currently not working when saving to CSV). However, as the following example shows, the type categories is nullable per se, i.e. we actually don't have to use `Int64` but can stick with a regular (not-nullable) `int64` instead:

```python
pd.Series(
  [1, 2, 1, None],
  dtype='category',  
)
```
```
0      1
1      2
2      1
3    NaN
dtype: category
Categories (2, int64): [1, 2]
```

I guess that's also the reason why they don't care to support `Int64` with categories in pandas. The only limitation I could think of is that you cannot have None as a category item, but since we anyway implicitly support None, there is not need for that anyway.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1147,20 +1147,6 @@ class Table(HeaderBase):
         except pickle.UnpicklingError:
             df = pd.read_pickle(path, compression='xz')
 
-        for column_id in df:
-            # Categories of type Int64 are somehow converted to int64.
-            # We have to change back to Int64 to make column nullable.
-            if isinstance(df[column_id].dtype,
-                          pd.core.dtypes.dtypes.CategoricalDtype):
-                if isinstance(df[column_id].dtype.categories,
-                              pd.core.indexes.numeric.Int64Index):
-                    labels = df[column_id].dtype.categories.values
-                    labels = pd.array(labels, dtype=pd.Int64Dtype())
-                    dtype = pd.api.types.CategoricalDtype(
-                        categories=labels,
-                        ordered=False)
-                    df[column_id] = df[column_id].astype(dtype)
-
         self._df = df
 
     def _save_csv(self, path: str):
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -34,7 +34,7 @@ install_requires =
     pyyaml >=5.4.1
     # https://github.com/audeering/audformat/pull/101
     # https://github.com/audeering/audformat/pull/142
-    pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3, <1.4.0
+    pandas >=1.1.5,!=1.3.0,!=1.3.1,!=1.3.2,!=1.3.3
```
frankenjoe, line 37: We should probably do this in a separate PR.

audformat/core/table.py
```diff
@@ -1147,20 +1147,6 @@ def _load_pickled(self, path: str):
         except pickle.UnpicklingError:
             df = pd.read_pickle(path, compression='xz')
 
-        for column_id in df:
```
frankenjoe, line 1150: We skip this as we don't need to convert to `int64` any longer.

setup.cfg
hagenw: Yes, makes sense.

frankenjoe: Ok, will revert once tests are passing.

frankenjoe: done

hagenw:
Great that we can get rid of the `categories` `Int64` combination, that is anyway not supported by `pandas`.

Bahaabrougui:
Is there a particular reason why we're using `int64` instead of `int8` or other small variants of integer, as here we're modeling categories with `int_x` types and those values are expected to be in a small range that `int8` or `int16` can cover. Or is it just to be on the safe side ?

PR #172, opened 2022-04-19
Release 0.14.2

frankenjoe:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,10 +7,14 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
-Version 0.14.2 (2022-04-19)
+Version 0.14.2 (2022-04-29)
 ---------------------------
 
-* Added: support ``pandas>=1.4.0``
+* Added: ``audformat.utils.index_has_overlap()``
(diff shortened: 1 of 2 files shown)


hagenw:
You forgot that we also had https://github.com/audeering/audformat/pull/164. And I guess there is also work to be done in https://github.com/audeering/audformat/pull/162, or is everything integrated now that @Bahaabrougui needs for `audbenchmark`?

frankenjoe:
Oh yes we should add #164 of course. But don't you think it's important we get a version out that supports latest pandas right away?

hagenw:
We are missing support for `pandas` 1.4.0 now since some time, so I think we can also wait a few more days.

frankenjoe:
Ok, I'll put WIP.

Bahaabrougui:
> You forgot that we also had #164. And I guess there is also work to be done in #162, or is everything integrated now that @Bahaabrougui needs for `audbenchmark`?

#162 is no longer WIP and can be merged too

frankenjoe:
@Bahaabrougui any other function you wanted to integrate or are we good to go?

Bahaabrougui:
> @Bahaabrougui any other function you wanted to integrate or are we good to go?

Nope, all good

PR #175, opened 2022-05-02
Improve speed of drop_files() on segmented index

### LINKED ISSUE #174: audformat.Table.drop_files() might result in a warning
I used `audformat.Table.drop_files(files, inplace=True)` on a segmented table, where `files` is a list of files and got the following warning:

```
.../python3.8/site-packages/audformat/core/table.py:374: PerformanceWarning: dropping on a non-lexsorted multi-index without a level parameter may impact performance.
  self.df.drop(index, inplace=True)
```

hagenw:
Closes #174 

This checks if we have a filewise or segmented index and sets the `level` argument accordingly when calling `self.df.drop()` inside `audformat.Table.drop_files()`.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -371,7 +371,11 @@ class Table(HeaderBase):
         else:
             index = self.files.intersection(files)
             index.name = define.IndexField.FILE
-            self.df.drop(index, inplace=True)
+            if self.is_segmented:
+                level = 'file'
+            else:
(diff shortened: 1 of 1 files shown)


frankenjoe:
Cool, did you also try to measure if it actually improves speed?

frankenjoe:
Hehe, ok, let's assume it's at least not slower :)

PR #177, opened 2022-05-06
WIP: Allow audformat.Column.set() to set duplicated index

hagenw:
This provides the possibility that `audformat.Column.set(values)` will not result in an error when the underlying index of the table has duplicated entries. This will only work if a user does not provide the `index` argument.


frankenjoe:
> This will only work if a user does not provide the `index` argument.

If we officially support duplicates, I think our target should be to support it everywhere.

hagenw:

Issue #178, opened 2022-05-06
Introduce an index-less table


hagenw:
As discussed in https://github.com/audeering/audformat/issues/86#issuecomment-1119581276 it might be a good idea to introduce an index-less table to store labels that we cannot fit into a filewise or segmented table.

To do so, we need to tackle the following points:

1. Find a name for the table in line with `filewise` and `segmented`, e.g. `index_less` or `general`
2. Update the table specifications (https://audeering.github.io/audformat/data-tables.html)
3. Add the Python implementation

As we show lots of examples in the table specifications I think it might be the easiest to tackle all three points in a single pull request,
or to start first with the Python implementation as the specification shouldn't be too complicated.

---

Having such a general table one could easily solve pairwise comparisons of files or segments without the need of duplicated index entries as requested in https://github.com/audeering/audformat/issues/86, e.g. consider the following filewise table defining index entries for each file

```
        idx
file    
f1      0
f2      1
f3      2
```

A comparison of those files can then be defined in an index-less table as

```
idx-0   idx-1
0       0
0       1
0       2
1       1
1       1
2       2
```

hagenw:
This is solved by the newly introduced `MiscTable` class, which can contain any kind of index (https://github.com/audeering/audformat/pull/189).

PR #179, opened 2022-05-07
WIP: Support for miscellaneous tables

frankenjoe:
Proposes a solution for #86 and #178 

This adds a new field `db.misc` to store miscellaneous tables. In contrast to the `Table` class the index of a `Misc` object must not be conform to `audformat` and can contain data of any type. The data of a misc object can be accessed with `db.misc['id'].get()` or `db.misc['id'].df.

image
image

### Example

Create database with a miscellaneous object:

```python
  [
    ('a', 0),
    ('b', 1),
    ('c', 2),
  ],
  names=['idx1', 'idx2'],
)
df = pd.DataFrame(
  {
    'c1': [.1, .2, .3],
    'c2': [-1, -2, -3],
  },
  index,
)

db.misc['misc'] = audformat.Misc(df, description='I am a miscellaneous table')
db
```
```
name: unittest
source: internal
usage: unrestricted
languages: [deu, eng]
misc:
  misc:
    description: I am a miscellaneous table
    levels: [idx1, idx2]
    columns: [c1, c2]
['db.misc.misc.csv', 'db.yaml']
```

Access object:

```python
db.misc['misc'].get()
```
```
            c1  c2
idx1 idx2         
a    0     0.1  -1
b    1     0.2  -2
c    2     0.3  -3
```

Save database:

```python
db.save('db')
audeer.list_file_names(
  'db', 
  recursive=True, 
  basenames=True,
)
```
```
['db.misc.df.csv', 'db.misc.s.csv', 'db.yaml']
```

### Discussion

* In principle, we could also support other types than `pd.Series` and `pd.DataFrame`, but we are limited to objects that can be stored as CSV, so I cannot think of anything else that make sense. -> **we decided to store only `pd.DataFrame`**
* Maybe we should actually drop  support for `pd.Series` to make it more similar to `Table` (e.g. we could then rename `Misc.obj` to `Misc.df`, which we also have for `Table`. And maybe rename to `MiscTable` then. -> **we decided not to support `pd.Series`**
* Currently, `Misc` objects are accessed through `db.misc` to separate them from `Table` objects. I principle we could also access them through `db.tables`. -> **we decided to keep it under `db.misc`**


frankenjoe:
@hagenw I did not add tests yet, but first wanted to get your feedback if this is the right direction to go.

frankenjoe:
For some reasons I suddenly encountered circulate import problems with `audformat/core/typing.py`, so I had to rename it to `typehints.py`.

hagenw:
Cool, thanks for trying to solve #178.

`audformat.Misc` is an interesting approach I haven't thought of before. It seems to be a mixture between `audformat.Table` and a none existing `audformat.Attachment` which we might need for storing additional header information like images (compare https://github.com/audeering/audb/issues/136).

I would have implemented the index-less table simply as `audformat.Table` to get the benefits that it is listed in `db.tables` and we can automatically use `set()` and `get()` and in general can access it as the other tables. To make that possible I would not allow any kind of index, but just no index.
The downside would be that the following methods would make not too much sense and we might need to add one breaking change:

```
drop_files() -> do nothing
drop_index() -> do nothing
ends -> []
extend_index() -> do nothing
files -> []
get() -> ignore `index` argument
index -> None
is_filewise -> False
is_segmented -> False
pick_files() -> do nothing
pick_index() -> do nothing
set() -> ignore `index` argument
starts() -> []
type -> `'index_less'`

Table.__init__(index=None) -> index-less table (BREAKING: at the moment a filewise index is generated)
```

As I haven't thought about adding a different type (`audformat.Misc`) before, I'm not sure yet what might be the better approach. Could you maybe list a few advantages you see in `audformat.Misc`?

hagenw:
Another advantage we would get for free when implementing it as a `Table` is `Scheme`, which I think would be nice to have for index-less tables as well.

frankenjoe:
> I would have implemented the index-less table

The first problem with an index-less table is that it is not supported by pandas:

```python
df = pd.DataFrame(
  {
    'c1': [1, 2, 3],
    'c2': [-1, -2, -3],
  },
  index=None,
)
df
```
```
   c1  c2
0   1  -1
1   2  -2
2   3  -3
```

Trying to force `None` results in an error:

```
df.index = None
```
```
TypeError: Index(...) must be called with a collection of some kind, None was passed
```

That means an index-less table would still have an index. Not really nice :)

The other big problem I see is that mot of our utility functions will not work with an index-less table, e.g. ` to_segmented_index()`, `concat()`, `iter_by_file()`, etc. Currently we have implemented those functions to work with filewise and segmented tables / indices, but I don't see them work with index-less tables. That means for most functions we can no longer say: hey, this function works with any table conform to audformat, but we have to exclude index-less tables, add additonal checks, raise additional error message etc.

> Could you maybe list a few advantages you see in `audformat.Misc`?

I see two main advantages:

1. It supports any kind of DataFrame / Series, with any kind of index, though not required by the use-case described in #86 I'm pretty sure it will be useful in the future
2. We don't have to change the specification of an audformat table, i.e. no need to touch utility functions etc.

The only downside might be that `Misc` tables do not share the same interface as `Table`, which is the reason I propose to group them under `.misc` and not `tables` (though it would be possible to have them also under `tables` if we want that). However, the interface of `Misc' is actually much easier, since we don't need control over index, data types, etc. Basically, it's just a pointer to any kind of DataFrame / Series and can be directly manipulated. Very simple and powerful I think.

frankenjoe:
> Another advantage we would get for free when implementing it as a `Table` is `Scheme`, which I think would be nice to have for index-less tables as well.

True, but on the other side it also offers a great advantage with respect to schemes. Currently, we store schemes completely in the header, which can lead to very long headers. The main problem are labels with many fields, which we currently store like this:

```python
labels = {
    's1': {'age', 30, 'gender': 'male' , ....},
    's2': {'age', 30, 'gender': 'male' , ....},
}
db.schemes['speaker'] = Scheme(labels=labels)
```

This is ok for a few fields, but imagine we have a 100 attributes we want to store with a each speaker. Now, which `Misc` we could do:

```
df = pd.DataFrame(labels)
db.misc['speaker'] = Misc(df)
db.schemes['speaker'] = Scheme(misc_id='speaker')
```

I.e. we link a column of `Table` to a scheme which we define as a `Misc` table. In the the header we would only list the speaker keys and the field names, but the actual values are stored in a separate file.

hagenw:
```python
df = pd.DataFrame(labels)
db.misc['speaker'] = Misc(df)
db.schemes['speaker'] = Scheme(misc_id='speaker')
```
that's indeed nice.

But it would also amplify a problem we have at the moment with the `labels` we allow for schemes: they don't come with schemes. E.g. if you have 
```python
labels = {
    's1': {'age', 30, 'gender': 'male' , ....},
    's2': {'age', 30, 'gender': 'male' , ....},
   ...,
}
```
you cannot have a scheme for `gender` in your database (or at least you cannot connect it to those labels).
This has two disadvantages:
* when looking at the schemes of a database, you will not see that it comes with gender labels
* you can set values with typos like `males` and `audformat` will not complain

I guess it will not be easy to solve this and being backwards compatible, but I think it would be one argument for still allowing to use schemes with `audformat.Misc`.

Otherwise, I totally agree with your observation that it will be a lot of work to extend all our helper functions to support an index-less table, even if `pandas` does not support that.
So, I guess you might be right and we should add another object.
Would you then restrict `audformat.Misc` to holding only dataframes/series and add another `audformat.Attachment` or would you combine those and allow to store everything in `audformat.Misc`?

If we go only for dataframe/series, the big question is how to best combine it with `list(db.tables)`. I think it does not make much sense to simply include it there, as the underlying object is not a table. Maybe we can add a helper function instead that lists all tables and misc objects.

frankenjoe:
> Would you then restrict audformat.Misc to holding only dataframes/series and add another audformat.Attachment or would you combine those and allow to store everything in audformat.Misc?

Currently we still have a `save()` / `load()` with `Misc`, I don't think that makes sense for images, which you mentioned as an example you would like to store under `audformat.Attachment`.

frankenjoe:
So maybe we should rename to `db.MiscTable` and `db.misc_tables` and maybe also restrict it to hold `pd.DataFrame` objects?

hagenw:
Yes, I would restrict it to pd.DataFrame, but I would stick with `audformat.Misc` and `db.misc`.

hagenw:
> can contain data of any type

Here, I'm not so sure as we don't want that the dataframe is different before saving to CSV and after loading from CSV.

hagenw:
As the `Misc` object does not support `Column` at the moment, there is no way to use a scheme with it. 
We could argue that a scheme should go only with labels that belong to files/segments,
but if we allow to use a `Misc` object to define labels of another scheme, it would still be nice to have in mind how to solve the issue described in https://github.com/audeering/audformat/pull/179#issuecomment-1123340559

frankenjoe:
> Here, I'm not so sure as we don't want that the dataframe is different before saving to CSV and after loading from CSV.

At least for some types we can solve it by storing the `dtype` in the header. But if a user stores e.g. tuples or complex objects it will become very tricky (probably impossible). So either we again have some restrictions on the data type or we simply add a warning to the docstring that data types may change when saving to CSV.

> but if we allow to use a `Misc` object to define labels of another scheme, it would still be nice to have in mind how to solve the issue described in [#179 (comment)](https://github.com/audeering/audformat/pull/179#issuecomment-1123340559)

``` python
gender = audformat.Scheme(labels=['female', 'male'])
labels = [
  {'speaker': 's1', 'age': 30, 'gender': 'female'},
  {'speaker': 's2', 'age': 50, 'gender': 'Male'},
]
df = pd.DataFrame(labels).set_index('speaker')
df.gender.astype(gender.to_pandas_dtype())
```
```
         age  gender
speaker             
s1        30  female
s2        50     NaN
```

So maybe we could add scheme support with `Misc` like this:

```
db.schemes['gender'] = audformat.Scheme(labels=['female', 'male'])
db.misc['speaker'] = Misc(df, scheme_ids={'gender: 'gender'})
```

frankenjoe:
So maybe we should not officially support the following use-case from the example in the description:

```python
db.misc['df'].obj['c'] = s
```
```
columns: [c]
index: [idx]
type: frame
```

Just like we also don't officially support `db['table'].df['c'] = s`.

So probably we should remove `.obj` all together and instead add `Misc.get()` method as we have with `Table`.

hagenw:
I created https://github.com/audeering/audformat/issues/181 to track the general scheme problem we have with `labels`.

```python
db.schemes['gender'] = audformat.Scheme(labels=['female', 'male'])
db.misc['speaker'] = Misc(df, scheme_ids={'gender: 'gender'})
```

would indeed be one solution to it.

hagenw:
> So maybe we should not officially support the following use-case from the example in the description:
> 
> ```python
> db.misc['df'].obj['c'] = s
> ```
> 
> ```
> columns: [c]
> index: [idx]
> type: frame
> ```
> 
> Just like we also don't officially support `db['table'].df['c'] = s`.
> 
> So probably we should remove `.obj` all together and instead add `Misc.get()` method as we have with `Table`.

Agree, remove `.obj` and do not officially allow for `db['table'].df['c'] = s`

frankenjoe:
> Agree, remove `.obj` and do not officially allow for `db['table'].df['c'] = s`

Ok, I also created https://github.com/audeering/audformat/issues/182 since we should probably also get rid of `Table.df`.

frankenjoe:
I applied the following changes:

* remove support for `pd.Series`
* renamed property `Misc.obj` to `Misc.df`
* added property `Misc.index`
* added methods `Misc.copy()` and `Misc.get()`

I updated the usage example in the description. Once we agree on the new interface, I will add tests and docstring examples. Regarding the scheme stuff (assigning schemes to columns and using `Misc` as a scheme) I suggest to work on it in separate PRs.

frankenjoe:
> Here, I'm not so sure as we don't want that the dataframe is different before saving to CSV and after loading from CSV.

Regarding this issue. Turns out that also with `Table` you can lose the data type if you store as CSV:

```python
import audformat


db['table'] = audformat.Table(index)
db['table']['column'] = audformat.Column()
values = pd.to_timedelta([1, 2, 3], unit='s')
db['table']['column'].set(values)
value = db['table']['column'].get()[0]
type(value)
```
<class 'pandas._libs.tslibs.timedeltas.Timedelta'>
```

But when we save and load again:

```
db.save('db')
db = audformat.Database.load('db')
value = db['table']['column'].get()[0]
type(value)
```
```
<class 'str'>
```

This can be avoided by assigning a scheme:

```python
...
db.schemes['time'] = audformat.Scheme('time')
...
db['table']['column'] = audformat.Column(scheme_id='time')
...
db.save('db')
db = audformat.Database.load('db')
value = db['table']['column'].get()[0]
type(value)
```
```
<class 'pandas._libs.tslibs.timedeltas.Timedelta'>
```

Once we add scheme support to `Misc', we have the option to apply the same workaround. But generally, I guess we will never be able to offer full data type support with CSV.

hagenw:

PR #183, opened 2022-05-12
WIP: Support for miscellaneous tables #2

frankenjoe:
Alternative implementation to #179 

What is not so nice about #179 is the different interfacing for `Table` and `Misc`. This is an alternative proposal that tries to make `Misc` more `Table`-like. It also has the advantage that we get support for schemes for free. Probably the most tricky part will be  how to maximize the code shared between `Misc` and `Table`. Currently not all methods of `Misc` will work properly as they still expect a filewise or segmented index, so the following example only demonstrates the basic `get()` and `set()` methods:

### Example

```python
  [
    ('a', 0),
    ('b', 1),
    ('c', 2),
  ],
  names=['idx1', 'idx2'],
)

db.schemes['gender'] = audformat.Scheme(labels=['female', 'male'])
db['misc']['column'] = audformat.Column(scheme_id='gender')
db['misc']['column'].set(['male', 'female', 'male'])
db
```
```
name: unittest
source: internal
usage: unrestricted
languages: [deu, eng]
schemes:
  gender:
    dtype: str
    labels: [female, male]
misc_tables:
  misc:
    columns:
      column: {scheme_id: gender}
    levels: [idx1, idx2]
```

Access data:

```python
db['misc'].get()
```
```
idx1 idx2        
a    0       male
b    1     female
c    2       male
```

Save and load:

```python
db.save('db')
db = audformat.Database.load('db')
db['misc'].get()
```
```
idx1 idx2        
a    0       male
b    1     female
c    2       male
```

Check for valid ID:

```python
db['misc'] = audformat.Table()
```
```
There is already a miscellaneous table with ID 'misc'.
```


hagenw:
I haven't looked at your implementation, but this is exactly what I had hoped for. You get an object that basically behaves like a table, but has no restrictions on the index.

The name `Misc` is maybe a little bit too generic now. I would still prefer a single word instead of `MiscTable`, and would propose one of the following:

* `audformat.Record` + `db.records`
* `audformat.Entry` + `db.entries`
* `audformat.Data` + `db.data`
* `audformat.Bag` + `db.bags`

frankenjoe:
> The name `Misc` is maybe a little bit too generic now. I would still prefer a single word instead of `MiscTable`, and would propose one of the following:
> 
>     * `audformat.Record` + `db.records`
> 
>     * `audformat.Entry` + `db.entries`
> 
>     * `audformat.Data` + `db.data`
> 
>     * `audformat.Bag` + `db.bags`

I agree that `Misc` is too generic, but I think also holds for your suggestions :) I guess it will be hard to find a intuitive name without `Table` in the class name.  So far I would still prefer `MiscTable` at the moment.

hagenw:
> > The name `Misc` is maybe a little bit too generic now. I would still prefer a single word instead of `MiscTable`, and would propose one of the following:
> > ```
> > * `audformat.Record` + `db.records`
> > 
> > * `audformat.Entry` + `db.entries`
> > 
> > * `audformat.Data` + `db.data`
> > 
> > * `audformat.Bag` + `db.bags`
> > ```
> 
> I agree that `Misc` is too generic, but I think also holds for your suggestions :) I guess it will be hard to find a intuitive name without `Table` in the class name. At the moment I would still prefer `MiscTable`.

The problem is that you need to type the name to access the table, e.g.

```python
db.misc_table['misc_table_id']
```
which means it will be much handier if we have a short name.

frankenjoe:
I think it#s ok to go with `db.misc_tables` since I don't expect we will have many of those tables.

frankenjoe:
> I think it's ok to go with `db.misc_tables` since I don't expect we will have many of those tables.

At least we should not encourage users to use them too frequently as such tables will not benefit from the advantages you get with an `audformat` index, e.g. you cannot use them with `audinterface'.

hagenw:
> > I think it's ok to go with `db.misc_tables` since I don't expect we will have many of those tables.
> 
> At least we should not encourage users to use them too frequently as such tables will not benefit from the advantages you get with an `audformat` index, e.g. you cannot use them with `audinterface'.

Yes and no. You cannot use them with `audinterface`, but if your only job is to work with speaker ID you will most likely use them a lot in addition to the normal tables.

frankenjoe:
> Yes and no. You cannot use them with `audinterface`, but if your only job is to work with speaker ID you will most likely use them a lot in addition to the normal tables.

Of course there are use-cases. Otherwise such a feature would not be needed :)

What we could do as a compromise is stick with the longer but at least expressive names `MiscTable` and `db.misc_tables', but like `Table` also make them available through the shortcut: `db['misc-table-id']`.

hagenw:
> make them available through the shortcut: db['misc-table-id']

I thought about this, but I think it will not work as `misc-table-id` and `table-id` could be the same, or do you plan to include a check that they have to have different names? At least in your first implementation you stored them with `db.misc.*.csv`, so it should be possible to use the same name.

frankenjoe:
> do you plan to include a check that they have to have different names

Currently we forbid that the ID of a `Table` starts with `misc.`. But if we make both available through `db['id']` then simply have to make sure that `Table` /  `MiscTable` IDs are unique.  Then we also don't have to store misc tables as `db.misc.*.csv`.

hagenw:
> > do you plan to include a check that they have to have different names
> 
> Currently we forbid that the ID of a `Table` starts with `misc.`. But if we make both available through `db['id']` we have to make sure that `Table` / `MiscTable` IDs are unique instead. Then we also don't have to store misc tables as `db.misc.*.csv`.

It seems indeed cleaner to not add an additional `misc` to the filename, but then I would vote for updating `audformat.utils.read_csv()` to not raise an error if the index is not conform as a user cannot see from the filename if the table is conform to `audformat`.

frankenjoe:
I have updated the example in the description. `MiscTable` objects can now be accessed via `db['table-id']` and it's not possible to have a `Table` and  `MiscTable` with the same ID.

> but then I would vote for updating `audformat.utils.read_csv()` to not raise an error if the index is not conform as a user cannot see from the filename if the table is conform to `audformat`.

Yes, makes sense, but maybe we can do that in a follow up PR. I would suggest to only introduce the new class with basic `get()` and `set()` functionality in this PR (so basically what I have implemented so far). We can then step by step add the other methods like `drop_index()`, `extend_index()` etc. in separate PRs.

hagenw:
> Yes, makes sense, but maybe we can do that in a follow up PR. I would suggest to only introduce the new class with basic `get()` and `set()` functionality in this PR (so basically what I have implemented so far). We can then step by step add the other methods like `drop_index()`, `extend_index()` etc. in separate PRs.

Yes, it will in general be better to make a few more pull requests as we need to introduce a few more changes here.

hagenw:
A few more points that come to my mind:
* even when we allow for `db[misc_table_id]`, I think we should have to separate `db.tables` and `db.misc_tables` entries. In addition, it would maybe be nice to have a utility method that lists both combined, e.g. `db.labels`
* the downside of allowing some of the methods to work with misc tables (e.g. `audformat.Database.__getitem__`) and others not (e.g. `audformat.Database.tables`) means it might be slightly confusing for the user and makes it also harder for us to write the documentation. But it might still make sense, as otherwise we would need to repeat a lot of those methods.
* we also need to think about how to handle `db.update()`.

frankenjoe:
> even when we allow for `db[misc_table_id]`, I think we should have to separate `db.tables` and `db.misc_tables` entries.

that's what we do at the moment

> means it might be slightly confusing for the user and makes it also harder for us to write the documentation

Yes, maybge `audformat.Database.__getitem__` and ``audformat.Database.__setitem__` should be the only methods that work with both as they provide a sort of shortcut. For other methods like `drop_tables()` I suggest to add according methods `drop_misc_tables()`. It means more code, but otherwise as you pointed out the interface will become too confusing.

> we also need to think about how to handle db.update().

Right, we will have to adapt it to also update `MiscTable`s.

hagenw:
The following methods or attributes affect tables inside `audformat.Database`:

* `__contains__(table_id)`
* `__getitem__(table_id)`
* `__setitem__(table_id, table)`
* `drop_tables(table_ids)`
* `pick_tables(table_ids)`
* `tables`

There are three options to handle them for `MiscTable`:
* use the same method
* re-implement with different name
* do not implement

The two cleanest solutions to me seem to be:
* use the same methods/attributes in all cases
* do not re-use any of it, but re-implement them (if needed)

So maybe we are better off if we don't allow for `db[misc_table_id]`, but require instead `db.misc_table[misc_table_id]`.

frankenjoe:
> * `__contains__(table_id)`
> 
> * `__getitem__(table_id)`
> 
> * `__setitem__(table_id, table)`

These three methods are now implemented to work with both, `Table` and `MiscTable`.

> * `drop_tables(table_ids)`
> 
> * `pick_tables(table_ids)`

As I mentioned in my previous comment I would introduce `drop_misc_tables()` and `pick_misc_tables()` in addition.

> * `tables`

We already have `mics_tables`.

frankenjoe:
I would argue that it makes sense this way as `__contains__(table_id)` is a shortcut for:

`table_id in db.tables or table_id in db.misc_tables`

and `__getitem__` / `__setitem__` are shortcuts for:

`db.tables[table_id]` / `db.misc_tables[table_id]` or `db.tables[table_id] = Table(...)` / `db.misc_tables[table_id] = MiscTable()`

hagenw:
> > * `drop_tables(table_ids)`
> > * `pick_tables(table_ids)`
> 
> As I mentioned in my previous comment I would introduce `drop_misc_tables()` and `pick_misc_tables()` in addition.

I would propose to simply skip those methods for `misc_tables`. We also don't have `drop_schemes()`.

frankenjoe:
> I would propose to simply skip those methods for `misc_tables`. We also don't have `drop_schemes()`.

Mhh, but then it's less obvious that `drop_tables()` only works for `Table`.

hagenw:
> > I would propose to simply skip those methods for `misc_tables`. We also don't have `drop_schemes()`.
> 
> Mhh, but then it's less obvious that `drop_tables()` only works for `Table`.

If we need to make that obvious, why not also use it for `misc_tables` then?

frankenjoe:
> If we need to make that obvious, why not also use it for `misc_tables` then?

Sorry, don't get it.

frankenjoe:
However, since we plan to add the option to store labels in a `MiscTable` it might in fact not be a good idea to add `drop_misc_tables()` / `pick_misc_tables()`.

hagenw:
> > If we need to make that obvious, why not also use it for `misc_tables` then?
> 
> Sorry, don't get it.

Doesn't matter I would not vote for it, and go with your suggestion to simply not add `drop_misc_tables()` / `pick_misc_tables()`

frankenjoe:
Ok! 

It's probably a bit early but we also have to keep in mind how to deal with `MiscTable`s in `audb`. If we allow to pick certain `MiscTable`s, then such functions would be useful. On the other hand, we could also keep it simple and only add an all-or-nothing option. What will be tricky is if a `MiscTable` is used to store scheme labels, in that case we always have to load the table, even with `only_metadata=True`.

hagenw:
> Ok!
> 
> It's probably a bit early but we also have to keep in mind how to deal with `MiscTable`s in `audb`. If we allow to pick certain `MiscTable`s, then such functions would be useful. On the other hand, we could also keep it simple and only add an all-or-nothing option. What will be tricky is if a `MiscTable` is used to store scheme labels, in that case we always have to load the table, even with `only_metadata=True`.

Good point. So if we allow `MiscTable` to specify parts of a header, I guess we always have to load all `MiscTable`s and we cannot filter them by `tables` in `audb`

frankenjoe:
> I guess we always have to load all `MiscTable`s and we cannot filter them by `tables` in `audb`

Well, at least load the tables that are actually used in scheme (should be easy to find out).

hagenw:

PR #184, opened 2022-05-13
Rename typing.py to typehints.py

frankenjoe:
As mentioned in #179 I had circular import problems with `typing.py` and it's generally probably not a good idea to import a module in a file with the same name. So I suggest we rename `typing.py` to `typehints.py`.


hagenw:
I still don't understand this error. We are never doing `from audformat.core import typing`, but always `from audformat.core.typing import ...`, which should not be of any problem. And I also think that the name `audformat.core.typing` makes totally sense and it is good to mimic the name `typing`

frankenjoe:
It occurs inside `PyCharm`, which I am using. So I would prefer if we could rename.

frankenjoe:
> It occurs inside `PyCharm`, which I am using. So I would prefer if we could rename.

Ok, found a way to avoid the problem. Will close.

PR #185, opened 2022-05-18
MiscTable: add table base class

frankenjoe:
We decided to split up #183 in smaller PRs. Here is the first one. It introduces a table base class to share code between `Table` and `MiscTable`, which we will add in another PR. Currently, the base class implements the following methods:

* `copy()`
* `drop_columns()`
* `get()`
* `load()`
* `pick_columns()`
* `save()`
* `set()`
 
Two abstract functions `_get_by_index()` and `_index_levels_and_converters()` have been added to inject code specific to filewise / segmented tables.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -148,6 +148,36 @@ class Base(HeaderBase):
         if self.split_id is not None and self.db is not None:
             return self.db.splits[self.split_id]
 
+    def drop_columns(
+            self,
+            column_ids: typing.Union[str, typing.Sequence[str]],
+            *,
+            inplace: bool = False,
+    ) -> 'Table':
+        r"""Drop columns by ID.
+
+        Args:
+            column_ids: column IDs
+            inplace: drop columns in place
+
+        Returns:
+            new ``Table`` if ``inplace=False``, otherwise ``self``
+
+        """
+        if not inplace:
+            return self.copy().drop_columns(column_ids, inplace=True)
+
+        if isinstance(column_ids, str):
+            column_ids = [column_ids]
+        column_ids_ = set()
+        for column_id in column_ids:
+            column_ids_.add(column_id)
+        self.df.drop(column_ids_, inplace=True, axis='columns')
+        for column_id in column_ids_:
+            self.columns.pop(column_id)
+
+        return self
+
     def get(
             self,
             index: pd.Index = None,
@@ -288,6 +318,32 @@ class Base(HeaderBase):
                     raise ex
         else:
             self._load_csv(csv_file)
+
+    def pick_columns(
+            self,
+            column_ids: typing.Union[str, typing.Sequence[str]],
+            *,
+            inplace: bool = False,
+    ) -> 'Table':
+        r"""Pick columns by ID.
+
+        All other columns will be dropped.
+
+        Args:
+            column_ids: column IDs
+            inplace: pick columns in place
+
+        Returns:
+            new ``Table`` if ``inplace=False``, otherwise ``self``
+
+        """
+        if isinstance(column_ids, str):
+            column_ids = [column_ids]
+        drop_ids = set()
+        for column_id in list(self.columns):
+            if column_id not in column_ids:
+                drop_ids.add(column_id)
+        return self.drop_columns(list(drop_ids), inplace=inplace)
 
     def save(
             self,
@@ -714,36 +770,6 @@ class Table(Base):
         table._df = self.df.copy()
         return table
 
-    def drop_columns(
-            self,
-            column_ids: typing.Union[str, typing.Sequence[str]],
-            *,
-            inplace: bool = False,
-    ) -> 'Table':
-        r"""Drop columns by ID.
-
-        Args:
-            column_ids: column IDs
-            inplace: drop columns in place
-
-        Returns:
-            new ``Table`` if ``inplace=False``, otherwise ``self``
-
-        """
-        if not inplace:
-            return self.copy().drop_columns(column_ids, inplace=True)
-
-        if isinstance(column_ids, str):
-            column_ids = [column_ids]
-        column_ids_ = set()
-        for column_id in column_ids:
-            column_ids_.add(column_id)
-        self.df.drop(column_ids_, inplace=True, axis='columns')
-        for column_id in column_ids_:
-            self.columns.pop(column_id)
-
-        return self
-
     def drop_files(
             self,
             files: typing.Union[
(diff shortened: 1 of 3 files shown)


frankenjoe:
> Entries missing in the docs of `audformat.Table`:

They should be back after I forced `Table` to show inherited members. 

The following looks weird, though:

image

hagenw:
This is a know bug and was fixed by https://github.com/audeering/sphinx-audeering-theme/pull/58, but I need to make a new `sphinx-audeering-theme` release.

hagenw:
> This is a know bug and was fixed by [audeering/sphinx-audeering-theme#58](https://github.com/audeering/sphinx-audeering-theme/pull/58), but I need to make a new `sphinx-audeering-theme` release.

Done (v1.1.9)

frankenjoe:
Great, looks good now:

image

hagenw:
There is still one minor difference in the docs, but this is due to a bug in sphinx, see https://github.com/sphinx-doc/sphinx/issues/2196

For the inherited methods the link to the sources changes from

image

to

image

But it's not a big issue and we can simply ignore it.

hagenw:
You even added something new to the docs:

image

This attribute does indeed exist, but maybe you can change the docstring text to "Table description"?

frankenjoe:
> This attribute does indeed exist, but maybe you can change the docstring text to "Table description"?

No really, since it is inherited from `HeaderBase`. So maybe we hide it by selecting the inherited members it should show and ignore those from `HeaderBase` as we do for all other classes at the moment?

hagenw:
Another problem is that we now document also all the `audobject` related methods like
* `dump()`
* `from_dict()`
* `to_dict()`

This we should hide from the user.

docs/api.rst
```diff
@@ -84,4 +84,5 @@ There are two types of tables:
 
 .. autoclass:: Table
     :members:
-    :special-members:
+    :inherited-members:
```
hagenw, line 87:
```suggestion
    :inherited-members: HeaderBase
```

This would hide the unwanted methods, but `description` we also want to have. Don't know yet how to integrate that.

hagenw: One solution would be to remove `description` from `HeaderBase` and implement in the deriving classes instead, then we could also adjust the docstring accordingly.

hagenw:
At the moment the following methods are only part of `audformat.core.table.Table`, but not `audformat.core.table.Base`.

### Column related methods

* `copy()`
* `drop_columns()`
* `pick_columns()`

I think the goal was to use `Column` and `Scheme` with `MiscTable` (and possible other tables),
so I guess they could go into `audformat.core.table.Base` as well. 
At least we have `audformat.core.table.Base.columns` and `audformat.core.table.Base._set_column()`.

### Index related methods

* `drop_index()`
* `extend_index()`
* `pick_index()`

Here the question is what kind of index we allow for `MiscTable`. I think we need some kind of convention, otherwise we can also not implement `_get_by_index()` and `_index_levels_and_converters()`.
But if we have a convention -- like the first column is the index, or all columns that don't have a `Column` object are the index --, we could also implement the methods listed above.

### Index and Column related methods

* `update()`

We will need `update()` in `MiscTable`. If we decide to also integrate all the methods listed above in `audformat.core.table.Base`, we could also add `update()` then.

audformat/core/table.py
```diff
@@ -28,7 +28,439 @@
...
+
+        """
+        for idx, data in values.items():
+            self.columns[idx].set(data, index=index)
+
+    def _get_by_index(
+            self,
+            index: pd.Index,
+    ) -> (pd.DataFrame, bool):  # pragma: no cover
+        raise NotImplementedError()
```
hagenw, line 365: As this is not self-explanatory, I would add a small comment, e.g.

```suggestion
        # Executed when calling `self.get(index=index)`.
        # Returns `df, df_is_copy`
        raise NotImplementedError()
```

audformat/core/table.py
```diff
@@ -28,7 +28,439 @@
...
+
+    def _index_levels_and_converters(self) -> typing.Tuple[
+            typing.Sequence[str],
+            typing.Dict[str, typing.Callable],
+    ]:  # pragma: no cover
+        raise NotImplementedError()
```
hagenw, line 371: Here also some comments would be nice, e.g.

```suggestion
        # Executed when loading table from CSV file.
        # Returns: `index_levels, index_converters`
        raise NotImplementedError()
```

frankenjoe:
> Another problem is that we now document also all the `audobject` related methods like

We are not using `audobject` in this project. The methods you mentioned are also inherited from `HeaderBase` and should not be included anymore.

hagenw:
> > Another problem is that we now document also all the `audobject` related methods like
> 
> We are not using `audobject` in this project. The methods you mentioned are also inherited from `HeaderBase` and should not be included anymore.

Yes, I also realized this after submitting my comment ;)

hagenw:
> > This attribute does indeed exist, but maybe you can change the docstring text to "Table description"?
> 
> Not really since it is inherited from `HeaderBase`. So maybe we hide it by selecting the inherited members it should show and ignore those from `HeaderBase` as we do for all other classes at the moment?

But hiding is also not a nice option, as `Table.description` is something a user should be aware of.

docs/api.rst
frankenjoe: But there is also `meta`, which we cannot easily remove from `HeaderBase` since we have some custom code there how to handle it. I think it's more elegant to keep both in the base class and find a generic docstring instead.

frankenjoe:
> At the moment the following methods are only part of `audformat.core.table.Table`, but not `audformat.core.table.Base`.

As I wrote in the description I only moved `get(), load(), save() and set()` to the base class so far. But I will have a second look which are the easy ones. The more complicated like `update()` I would prefer to handle in separate PRs once we have actually added `MiscTable`.

docs/api.rst
frankenjoe: But since this also relates to the other classes I suggest to tackle this issue in a separate PR and stay with the current solution of hiding those for now.

hagenw:
> > At the moment the following methods are only part of `audformat.core.table.Table`, but not `audformat.core.table.Base`.
> 
> As I wrote in the description I only moved `get()`, `load()`, `save()` and `set()` to the base class so far. But I will have a second look which are the easy ones. The more complicated like `update()` I would prefer to handle in separate PRs once we have actually added `MiscTable`.

Ah, ok, sorry forgot this.
It's fine with me to move other methods at a later stage.

docs/api.rst
hagenw: I crated https://github.com/audeering/audformat/issues/186 for `description`

frankenjoe:
> It's fine with me to move other methods at a later stage.

I added `drop_columns()` and `pick_columns()` now, since those are no-brainers. I updated the description accordingly.

audformat/core/table.py
```diff
@@ -28,7 +28,499 @@
...
+        Args:
+            column_ids: column IDs
+            inplace: drop columns in place
+
+        Returns:
+            new ``Table`` if ``inplace=False``, otherwise ``self``
+
+        """
+        if not inplace:
+            return self.copy().drop_columns(column_ids, inplace=True)
```
hagenw, line 168: Does this work even though you don't implement `self.copy()`?

hagenw: It works for `audformat.core.table.Table` as you implement `copy()` there, but a developer does not know that she has to implement it, so I think we have to move it to `Base` as well when moving `drop_columns()`

frankenjoe: Good catch. The problem is that I'm not sure if we can implement it in the base class since we want to create a copy of the derived class. So probably we'll have to make it an abstract method to force the developer to implement it.

frankenjoe: Turns out it's simply, just use `self.__class__(...)` to construct the object in the base class. Note that I override `copy()` in the child class, but just to get the correct return type in the docs.

frankenjoe: done

PR #187, opened 2022-05-20
WIP: Replace iso-639 with pycountry

hagenw:
As I would like to add `audformat.utils.map_country()` besides `audformat.utils.map_language()` in order to support ISO 3166 for countries, I would propose to switch to `pycountry` instead which will then also allow us to use ISO 3166 for countries.

I guess the main reason we used https://github.com/noumar/iso639 instead of `pycountry` was that it is smaller as it only contains language codes, correct?

* `pycountry` wheel size: 11MB
* `iso-639` wheel size: 165K

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -4,9 +4,9 @@ import re
 import sys
 import typing as typing
 
-import iso639
 import numpy as np
 import pandas as pd
+import pycountry
 
 import audeer
 import audiofile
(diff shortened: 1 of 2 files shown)


hagenw:
This pull request will most likely not be needed as we can use https://github.com/deactivated/python-iso3166 to map countries.

frankenjoe:
> I guess the main reason we used https://github.com/noumar/iso639 instead of `pycountry` was that it is smaller as it only contains language codes, correct?

yes

hagenw:
Closing in favor of https://github.com/audeering/audformat/pull/188

Issue #181, opened 2022-05-11
Add support for using schemes in labels


hagenw:
Currently we have the problem that we cannot define schemes for attributes defined inside labels, e.g.

```python
labels = {
  's1': {'age': 30, 'gender': 'female'},
  's2': {'age': 50, 'gender': 'male'},
}
db.schemes['speaker'] = audformat.Scheme(labels=labels)
```
In the above example it would be nice to have the schemes `age` and `gender` as well in order to
* get a correct overview of what schemes are in the database
* avoid typos like `'Male'`

Possible changes need to be done in a backward compatible way.

---

One solution might be to introduce an additional `audformat.Labels` class that can be handed as `labels` argument to `Schemes()` in addition to a list or a dict. This class could then assign schemes to the different keys.

frankenjoe:
We should be able to solve this once we have integrated the new `MiscTable` class, see https://github.com/audeering/audformat/pull/179#issuecomment-1123296715

hagenw:
Done in https://github.com/audeering/audformat/pull/201

PR #189, opened 2022-05-23
Misc table class

frankenjoe:
Relates to #86 #178 #181 

Adds `MiscTable` class:

image

In addition adds two new helper functions:

image

and a new error class `TableExistsError`:

image

which may be raised by `Database.__setitem__()`:

image

### DIFF
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -235,7 +235,7 @@ def is_filewise_index(
         True
         >>> is_filewise_index(pd.Index([]))
         False
-        
+
     """
     if not isinstance(obj, pd.Index):
         obj = obj.index
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -257,6 +257,12 @@ def is_segmented_index(
     Returns:
         ``True`` if index is segmented, otherwise ``False``
 
+    Example:
+        >>> is_segmented_index(segmented_index())
+        True
+        >>> is_segmented_index(pd.Index([]))
+        False
+        
     """
     if not isinstance(obj, pd.Index):
         obj = obj.index
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -230,6 +230,12 @@ def is_filewise_index(
     Returns:
         ``True`` if index is filewise, otherwise ``False``
 
+    Example:
+        >>> is_filewise_index(filewise_index())
+        True
+        >>> is_filewise_index(pd.Index([]))
+        False
+        
     """
     if not isinstance(obj, pd.Index):
         obj = obj.index
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -183,7 +183,7 @@ def filewise_index(
 
 
 def index_type(
-        obj: typing.Union[pd.Index, pd.Series, pd.DataFrame]
+        obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
 ) -> define.IndexType:
     r"""Derive index type.
 
@@ -214,6 +214,53 @@ def index_type(
         return define.IndexType.FILEWISE
     else:
         return define.IndexType.SEGMENTED
+
+
+def is_filewise_index(
+        obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
+) -> bool:
+    r"""Check if object has a filewise index.
+
+    Returns ``True`` if index is a filewise index conform to
+    :ref:`table specifications <data-tables:Tables>`.
+
+    Args:
+        obj: object
+
+    Returns:
+        ``True`` if index is filewise, otherwise ``False``
+
+    """
+    if not isinstance(obj, pd.Index):
+        obj = obj.index
+
+    return len(obj.names) == 1 and obj.names[0] == define.IndexField.FILE
+
+
+def is_segmented_index(
+        obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
+) -> bool:
+    r"""Check if object has a segmented index.
+
+    Returns ``True`` if index is a segmented index conform to
+    :ref:`table specifications <data-tables:Tables>`.
+
+    Args:
+        obj: object
+
+    Returns:
+        ``True`` if index is segmented, otherwise ``False``
+
+    """
+    if not isinstance(obj, pd.Index):
+        obj = obj.index
+
+    return (
+            len(obj.names) == 3 and
+            obj.names[0] == define.IndexField.FILE and
+            obj.names[1] == define.IndexField.START and
+            obj.names[2] == define.IndexField.END
+    )
 
 
 def segmented_index(
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -256,10 +256,10 @@ def is_segmented_index(
         obj = obj.index
 
     return (
-            len(obj.names) == 3 and
-            obj.names[0] == define.IndexField.FILE and
-            obj.names[1] == define.IndexField.START and
-            obj.names[2] == define.IndexField.END
+        len(obj.names) == 3
+        and obj.names[0] == define.IndexField.FILE
+        and obj.names[1] == define.IndexField.START
+        and obj.names[2] == define.IndexField.END
     )
 
 
--- a/audformat/core/index.py
+++ b/audformat/core/index.py
@@ -262,7 +262,7 @@ def is_segmented_index(
         True
         >>> is_segmented_index(pd.Index([]))
         False
-        
+
     """
     if not isinstance(obj, pd.Index):
         obj = obj.index
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -20,6 +20,7 @@ from audformat.core.errors import (
 from audformat.core.index import (
     filewise_index,
     index_type,
+    is_segmented_index,
 )
 from audformat.core.media import Media
 from audformat.core.split import Split
@@ -1082,7 +1083,7 @@ class Table(Base):
         result = super().get(index, map=map, copy=copy)
 
         # if necessary, convert to segmented index and replace NaT
-        is_segmented = index_type(result.index) == define.IndexType.SEGMENTED
+        is_segmented = is_segmented_index(result.index)
         if (
                 (not is_segmented and as_segmented)
                 or (is_segmented and not allow_nat)
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -543,13 +543,72 @@ class Base(HeaderBase):
         return column
 
 
+class MiscTable(Base):
+    r"""Miscellaneous table.
+
+    ..note:: Only use for tables
+        with an index that is not conform to
+        :ref:`table specifications <data-tables:Tables>`.
+        Otherwise, it is preferable to use
+        :class:`audformat.Table`.
+
+    To fill a table with labels,
+    add one or more :class:`audformat.Column`
+    and use :meth:`audformat.MiscTable.set` to set the values.
+
+    Args:
+        index: table index,
+            if ``None`` creates an empty table
+        split_id: split identifier (must exist)
+        media_id: media identifier (must exist)
+        description: database description
+        meta: additional meta fields
+
+    """
+    def __init__(
+            self,
+            index: pd.Index = None,
+            *,
+            split_id: str = None,
+            media_id: str = None,
+            description: str = None,
+            meta: dict = None,
+    ):
+        super().__init__(
+            index,
+            split_id=split_id,
+            media_id=media_id,
+            description=description,
+            meta=meta,
+        )
+
+        self.levels = None
+        r"""Index levels."""
+
+        if index is not None:
+            if isinstance(index, pd.MultiIndex):
+                levels = list(index.names)
+            else:
+                levels = [index.name]
+            self.levels = levels
+
+    def _get_by_index(self, index: pd.Index) -> (pd.DataFrame, bool):
+        return self.df.loc[index]
+
+    def _index_levels_and_converters(self) -> typing.Tuple[
+        typing.Sequence[str],
+        typing.Dict[str, typing.Callable],
+    ]:
+        return self.levels, {}
+
+
 class Table(Base):
-    r"""Table with annotation data.
+    r"""Table conform to :ref:`table specifications <data-tables:Tables>`.
 
     Consists of a list of file names to which it assigns
     numerical values or labels.
     To fill a table with labels,
-    add one ore more :class:`audformat.Column`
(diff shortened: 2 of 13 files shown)


audformat/core/table.py
```diff
@@ -543,13 +543,117 @@ def _set_column(self, column_id: str, column: Column) -> Column:
...
+        return super().copy()
+
+    def _get_by_index(self, index: pd.Index) -> (pd.DataFrame, bool):
+        return self.df.loc[index], False
+
+    def _index_levels_and_converters(self) -> typing.Tuple[
+        typing.Sequence[str],
+        typing.Dict[str, typing.Callable],
+    ]:
+        return self.levels, {}
```
frankenjoe, line 647: We currently do not store type information for the levels. But since we cannot link schemes to index columns it might be useful to do so. @hagenw what's your opinion? In any case, I'd prefer to work on it in a separate PR.

frankenjoe: https://github.com/audeering/audformat/issues/190

hagenw:
When having `audformat.is_filewise_index()` and `audformat.is_segmented_index()` does it then still make sense to also have `audformat.index_type()`?

audformat/core/index.py
```diff
@@ -216,6 +216,53 @@ def index_type(
...
+    Returns ``True`` if index is a filewise index conform to
+    :ref:`table specifications <data-tables:Tables>`.
+
+    Args:
+        obj: object
+
+    Returns:
+        ``True`` if index is filewise, otherwise ``False``
+
+    """
```
hagenw, line 233: c

audformat/core/index.py
```diff
@@ -216,6 +216,53 @@ def index_type(
...
+    Returns ``True`` if index is a segmented index conform to
+    :ref:`table specifications <data-tables:Tables>`.
+
+    Args:
+        obj: object
+
+    Returns:
+        ``True`` if index is segmented, otherwise ``False``
+
+    """
```
hagenw, line 254:
```suggestion
    Example:
        >>> is_segmented_index(segmented_index())
        True
        >>> is_segmented_index(pd.Index([]))
        False
        
    """
```

frankenjoe:
> When having `audformat.is_filewise_index()` and `audformat.is_segmented_index()` does it then still make sense to also have `audformat.index_type()`?

I guess we could. But actually my first idea was to extend `index_type()` to return `'miscellaneous'` if it is neither filewise nor segmented. But unfortunately we currently raise an error in this case, i.e. this would be a breaking change. So as an act of necessity I introduce these two new functions. Another option would be to add an argument to `index_type()` that it should not raise or add it again with a new name. Or do you think `is_filewise_index()` and `is_segmented_index()` is a good solution? And yes, we should probably not have all three function in parallel.

tests/test_errors.py
```diff
@@ -52,3 +53,27 @@ def test_errors():
...
+        db.splits['bad']
+
+    with pytest.raises(audformat.errors.BadKeyError):
+        db.tables['bad']
+
+    with pytest.raises(ValueError):
+        db['misc'] = audformat.Table()
+
+    with pytest.raises(ValueError):
+        db['files'] = audformat.MiscTable(pd.Index([]))
```
hagenw, line 79: Why does this raise a `ValueError`?

There is also no `Raises` section in the docstring of `MiscTable`

audformat/core/index.py
hagenw, line 233:
```suggestion
    Example:
        >>> is_filewise_index(filewise_index())
        True
        >>> is_filewise_index(pd.Index([]))
        False
        
    """
```

hagenw:
> > When having `audformat.is_filewise_index()` and `audformat.is_segmented_index()` does it then still make sense to also have `audformat.index_type()`?
> 
> I guess we could. But actually my first idea was to extend `index_type()` to return `'miscellaneous'` if it is neither filewise nor segmented. But unfortunately we currently raise an error in this case, i.e. this would be a breaking change. So as an act of necessity I introduced two new functions. Another option would be to add an argument to `index_type()` that it should not raise or add it again with a new name. Or do you think `is_filewise_index()` and `is_segmented_index()` is a good solution? And yes, we should probably not have all three function in parallel.

There might be still good use-cases for `audformat.index_type()`, e.g.
```python
if audformat.index_type(idx1) != audformat.index_type(idx2):
    ...
```
This would be more complicated to write with `audformat.is_filewise_index()` and `audformat.is_segmented_index()`.

On the other hand we are also doing stuff like
```python
is_segmented = audformat.index_type(idx) == audformat.define.IndexType.SEGMENTED
```
which is easier using `audformat.is_segmented_index()`.
So maybe there is even space for all three functions.

I also thought about that `audformat.index_type()` might return `'miscellaneous'` or `'other'`, but this would somehow then lead to the question why we are not adding a `'miscellaneous'` index to `Table` and extend all our functions to support `'miscellaneous'` index as well.

frankenjoe:
> why we are not adding a `'miscellaneous'` index to `Table` and extend all our functions to support `'miscellaneous'` index as well.

I think there is a good reason to not do that - `filewise` and `segmented` indices can be combined, so that `db.tables['a'] + db.tables['b']` always works. But once we allow `Table` with an `miscellaneous` this is not longer true. And also other functions like `process_index()` do not make sense with `miscellaneous` indices. So I think it's better to keep a clear separation.

hagenw:
> > why we are not adding a `'miscellaneous'` index to `Table` and extend all our functions to support `'miscellaneous'` index as well.
> 
> I think there is a good reason to not do that - `filewise` and `segmented` indices can be combined, so that `db.tables['a'] + db.tables['b']` always works. But once we allow `Table` with an `miscellaneous` this is not longer true. And also other functions like `process_index()` do not make sense with `miscellaneous` indices. So I think it's better to keep a clear separation.

Good, but then I think we can also simply stay with the current behavior of `audformat.index_type()`.

But I would propose that you update parts of the code where we do stuff like
```python
is_segmented = index_type(result.index) == define.IndexType.SEGMENTED
```
with the new functions.

frankenjoe:
> But I would propose that you update parts of the code where we do stuff like

Good idea, will do!

frankenjoe:
> But I would propose that you update parts of the code where we do stuff like

done

tests/test_errors.py
frankenjoe: I replaced `ValueError` with a custom class `TableExistsError` like we do for other errors and added the missing raised section in `Database.__setitem()`. I also added a comment to the test and updated the description of the PR accordingly.

hagenw: What happens if you try to set a filewise table with the same ID as another already existing filewise table?

frankenjoe: Then the old table is simply replaced. Same when you replace a filewise with a segmented table or vice versa. Or when you replace a misc table with a misc table.

hagenw:
> > why we are not adding a `'miscellaneous'` index to `Table` and extend all our functions to support `'miscellaneous'` index as well.
> 
> I think there is a good reason to not do that - `filewise` and `segmented` indices can be combined, so that `db.tables['a'] + db.tables['b']` always works. But once we allow `Table` with an `miscellaneous` this is not longer true. And also other functions like `process_index()` do not make sense with `miscellaneous` indices. So I think it's better to keep a clear separation.

I did take a look at our current `audformat` specification in order to see how to integrate `MiscTable` (https://github.com/audeering/audformat/issues/192), and after that I would again vote for introducing a new table type instead of a new table class.
Functions that only support filewise or segmented index can simply raise an error or skip the table.

frankenjoe:
> I did take a look at our current `audformat` specification in order to see how to integrate `MiscTable` (#192), and after that I would again vote for introducing a new table type instead of a new table class.
> Functions that only support filewise or segmented index can simply raise an error or skip the table.

I would keep a clear cut. If we allow `Table` to take any index, it can happen that a user accidentally creates a table not conform to `audformat` when she actually wanted to create a filewise or segmented index. She we will probably not notice until someone complains that `process_index()` is not working with that table.

hagenw:
image

Showing the `levels` here is the desired output, right?

tests/test_errors.py
hagenw: As we share the same table IDs, I think it would make more sense if we also allow replacing a filewise with a misc table, etc.

hagenw:
When trying to access a column in a misc table

```python
    import audformat
    import audformat.testing
    import pandas as pd


        [
            ('f1', 'f2'),
            ('f1', 'f3'),
            ('f2', 'f3'),
        ],
        names=['file', 'other'],
    )
    db['misc']['values'] = audformat.Column()
    values_list = [0, 1, 0]
    values_dict = {'values': values_list}
    db['misc'].set(values_dict)
    db['misc']['values'].get()
```

I get

```
TypeError                                 Traceback (most recent call last)
Input In [18], in <cell line: 20>()
     18 values_dict = {'values': values_list}
     19 db['misc'].set(values_dict)
---> 20 db['misc']['values'].get()

File ~/git/audeering/audformat/audformat/core/column.py:192, in Column.get(self, index, map, copy, as_segmented, allow_nat, root, num_workers, verbose)
    187 if self._table is None:
    188     raise RuntimeError(
    189         'Column is not assigned to a table.'
    190     )
--> 192 result = self._table.get(
    193     index,
    194     copy=False,
    195     as_segmented=as_segmented,
    196     allow_nat=allow_nat,
    197     root=root,
    198     num_workers=num_workers,
    199     verbose=verbose,
    200 )
    201 result = result[self._id]
    203 if map is not None:

TypeError: get() got an unexpected keyword argument 'as_segmented'
```

frankenjoe:
> Showing the `levels` here is the desired output, right?

Yes, in order to restore the index we need to store the levels. I first wanted to call it `index`, but we already have a `MiscTable.index` property so we have to choose a different name. You think `levels` is ok?

tests/test_errors.py
frankenjoe: Mhh ok, but since we plan to reference `MiscTable`s in schemes it can then happen that we lose the labels.

frankenjoe:
> When trying to access a column in a misc table

Oh I overlooked that we actually have a `as_segmented` argument in `Column.get()` which obviously doesn't make sense in combination with `MiscTable`. I wanted to avoid it, but I guess we'll have to introduce a `MiscColumn` :-(

frankenjoe:
The other option we have is to ignore `as_segmented` when called on a `Column` that belongs to a `MiscTable`. @hagenw what's your opinion here?

hagenw:
> > I did take a look at our current `audformat` specification in order to see how to integrate `MiscTable` (#192), and after that I would again vote for introducing a new table type instead of a new table class.
> > Functions that only support filewise or segmented index can simply raise an error or skip the table.
> 
> I would keep a clear separation. If we allow `Table` to take any index, it can happen that a user accidentally creates a table not conform to `audformat` when she actually wanted to create a filewise or segmented index. She we will probably not notice until someone complains that `process_index()` is not working with that table.

I tried to incorporate the MiscTable specifications in #195, and it looks quite ok.
I used this branch as the base, so we could integrate it here.

frankenjoe:
> The other option we have is to ignore `as_segmented` when called on a `Column` that belongs to a `MiscTable`. @hagenw what's your opinion here?

I pushed a solution (your example should work now).

hagenw:
> > Showing the `levels` here is the desired output, right?
> 
> Yes, in order to restore the index we need to store the levels. I first wanted to call it `index`, but we already have a `MiscTable.index` property so we have to choose a different name. You think `levels` is ok?

Yes, `levels` makes sense, and it's also nice to have it accessibly to the user as you did (or would it be better as a property?).
I was more wondering if it should be part of the table YAML representaton (as it is at the moment) or not.
Can we freely choose if we want to display it there, or not?

frankenjoe:
> Can we freely choose if we want to display it there, or not?

No, we have to store this information in YAML. Otherwise the information is lost when we load the database from disk.

tests/test_errors.py
hagenw: That is a valid point, but to a user it feels strange that they are more protected than a filewise table, which I can replace by a segmented table.

Is there maybe a way to check if such a table is used in a scheme, and only then raise the error (when we implement that feature)

hagenw:
> > The other option we have is to ignore `as_segmented` when called on a `Column` that belongs to a `MiscTable`. @hagenw what's your opinion here?
> 
> I pushed an according solution (your example should work now).

The solution looks good, and my example does now indeed work.

hagenw:
> > I was more wondering if it should be part of the table YAML representaton (as it is at the moment) or not.
> > Can we freely choose if we want to display it there, or not?
> 
> We have to store this information in YAML. Otherwise the information is lost when we load the database from disk.

It's fine with me to stay with the current solution. We need the information anyway to know what columns in the CSV file belong to the index. And way not make this also available to the user.

tests/test_errors.py
hagenw: > The other option we have is to generally forbid that a user can replace an existing table.

Which will also not really help as the user can use `drop_tables()` or `del` to first remove the table and then assign a different one under the same name.

I would propose that for now we allow for replacing the tables and cover raising an error when we tackle `MiscTable` as a scheme.

frankenjoe:
> The solution looks good, and my example does now indeed work.

I changed the docstring of `Column.get()` to:

image

You think this is enough or do we need to highlight it more?

hagenw:
Should be sufficient.

The users can anyway do no harm by setting the arguments.

tests/test_errors.py
frankenjoe: > Which will also not really help as the user can use drop_tables() or del to first remove the table and then assign a different one under the same name.

This currently works only for `Table` objects and not `MiscTable`. And it still has the advantage that the user has to explicitly drop the tables first. Currently, it can happen that a user accidentally uses the same table ID twice.

frankenjoe: On the other hand, you are right the user can always use `del`, also with schemes etc. So maybe we simply assume that a user takes care of not overwriting it by accident. Will change it accordingly.

frankenjoe: Oh wait, there is of course one problem: the user can now do: `db.tables['table'] = Table()` and `db.misc_tables['table'] = MiscTable()`.

frankenjoe: > This is how the current implementation works, right?

yes

hagenw: Good, then I would now also be in favor of using this and will resolve this.

hagenw:
I have now finished the review and would suggest that we first merge this pull request, and only afterwards tackle https://github.com/audeering/audformat/pull/195.

Do you agree?

PR #195, opened 2022-06-03
DOC: misc table specifications

### LINKED ISSUE #192: Add MiscTable to the audformat specifications
There is the option to not add `MiscTable` (compare #189) to the official specifications, but as it stores tables under the same name as a normal table, adds entries to the YAML file and will be tightly integrated to our Python implementation I would vote to include it.

At the moment the structure of the specification looks like this

image

We could integrate the specifications under "Tables", but at the moment "Tables" is very much focused on filewise or segmented tables, e.g.

---

image

---

image

---

image

---

After reading the specification, I think it would make more sense (and will be much easier to understand) to introduce a `miscellaneous` table type and don't bother the user with different definitions and classes.
Methods like `__add__` or `process_index()` can then check for the index type and raise an error or skip the table if the index is not filewise or segmented.

hagenw:
Closes #192 

Updates the database specification by a Misc Table. I only used the term "Miscellaneous Table" at two places and introduced the short name Misc Table instead as I think it will be easier to talk about those tables if the name and class name do match.

Overall, it looks quite ok and I think we can integrate the misc table in a meaningful way to the specification.

~~Note, due to the error reported in https://github.com/audeering/audformat/pull/189#issuecomment-1145859656, I have commented out one of the code examples.~~


docs/data-format.rst
```diff
@@ -1,42 +1,55 @@
...
-audformat is implemented in the :class:`audformat.Database`.
-Written to hard disk it is converted to a YAML file (*Header*),
-which contains information about the raters,
-annotation schemes and meta information.
-Actual annotations are stored
-across (possibly) multiple CSV files (*Tables*).
+An audformat database consists of a **header**,
+several **tables**,
+and **media** files.
+On hard disc all of them are stored inside a single folder.
```
frankenjoe, line 7:
```suggestion
On hard disk all of them are stored inside a single folder.
```

I think this is the more common spelling.

docs/data-misc-tables.rst
```diff
@@ -0,0 +1,84 @@
...
+    import audformat
+
+
+    audformat.core.common.format_series_as_html()
+
+
+A misc table or miscellaneous table links labels to an index.
+The index has no restrictions,
+but should not contain duplicated entries.
+The arbitrary number of index columns of a misc table
```
frankenjoe, line 18:
```suggestion
An arbitrary number of index columns of a misc table
```

docs/data-misc-tables.rst
```diff
@@ -0,0 +1,84 @@
...
+are followed by an arbitrary number of label columns.
```
frankenjoe, line 19:
```suggestion
is followed by an arbitrary number of label columns.
```

frankenjoe:
Looks good, I left some minor comments.

Regarding the issue with level names we can also merge and change after solving #196

docs/data-format.rst
```diff
@@ -1,42 +1,55 @@
...
+The header is stored as a YAML file,
+the tables contain labels stored in (possibly) multiple CSV files,
```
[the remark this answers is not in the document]
hagenw: I would not put PKL files as part of the specification as those are only used in the implementation of `audb` to speed up processing, but if you want to implement a C-library for `audformat` you don't have to support PKL files.

docs/data-misc-tables.rst
hagenw: If we change to "An arbitrary" I'm afraid that the meaning changes. Now it could also mean that after some index columns we store some label columns. But we need to emphasize, that we first store all index columns and then the label columns.

So, maybe rephrasing to:

```rst
The index has no restrictions
and can contain an arbitrary number of columns (called levels),
but should not contain duplicated entries.
In the CSV file of the misc table,
the index columns are stored in front of the label columns.
```

docs/data-format.rst
```diff
@@ -1,42 +1,55 @@
...
+    ``db.yaml``                 Meta information, schemes, list of raters
+    ``db.<table_id>.csv``       Table with files or file segments as index
+                                and columns holding annotations
+    ``db.<misc_table_id>.csv``  Misc table with unspecified index
+                                and columns holding annotations
+    ``<folder(s)/file(s)>``     Audio/Video files referenced in the tables
+    ==========================  ==========================================
 
-The connection between the header and the tables
+The connection between the header, media file a table
```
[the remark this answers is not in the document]
frankenjoe: No, I don't think we need to update the sketch. I changed the suggestion accordingly.

docs/data-misc-tables.rst
frankenjoe: Yes, your suggestion sounds good.

hagenw: Changed accordingly in https://github.com/audeering/audformat/pull/195/commits/6e8000f3f24a376943bed7b7fd003b555447c528

frankenjoe:
Copying from #196 

image

hagenw:
Added the example:

image

frankenjoe:
> Added the example:

Great, thanks!

PR #198, opened 2022-06-08
List all tables with Database.__iter__

### LINKED ISSUE #191: Introduce property to audformat.Database to list all tables
Usually the first thing I do after loading a database is calling

```python
list(db.tables)
```

After https://github.com/audeering/audformat/pull/189 is merged and we added `MiscTable` this can become even more complicated:

```python
list(db.tables) + list(db.misc_tables)
```

The question is if we should introduce a property as a short cut for this. Implementation is straightforward, but finding a name and deciding if we should always show misc tables as well isn't.

hagenw:
Closes #191.

This implements `audformat.Database.__iter__` in order to use `list(db)` to get a sorted list of all table and misc table IDs.

I extended the docstring example to also contain a misc table and use the `list(db)` statement at the end:

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -118,6 +118,8 @@ class Database(HeaderBase):
             media_id: audio
             columns:
               column: {scheme_id: emotion, rater_id: rater}
+        >>> list(db)
+        ['table']
 
     """
     def __init__(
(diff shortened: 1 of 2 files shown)


frankenjoe:
> I wasn't sure if I should include in the the doc example, but have add it now:

Cool! You think we should add also a `MIscTable` to show the user that those are also returned?

hagenw:
OK, I have extended the docstring example by adding a misc table to the database.

PR #201, opened 2022-06-24
Support MiscTable as labels in audformat.Scheme

hagenw:
Add support for using `MiscTable` as a scheme with labels. This has no the big advantage that we can define schemes for entries in schemes.

~~Not sure if I overlooked something, but it seems relatively easy to add.~~

image

image

I added the `speaker` scheme as example to `audformat.testing.create_db()`:

```python
import audformat.testing

audformat.testing.create_db()
```
results in

```yaml
name: unittest    
description: A database for unit testing.
source: internal
[...]
schemes:
[...]
  speaker:
    {dtype: str, labels: misc}
[...]
misc_tables:
  misc:
    levels: [speaker]
    columns:
      age: {scheme_id: age}
      gender: {scheme_id: gender}
```

### DIFF
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -224,7 +224,8 @@ class Column(HeaderBase):
                     f"Column '{self._id}' is not assigned to a scheme."
                 )
 
-            labels = self._table._db.schemes[self.scheme_id].labels
+            scheme = self._table._db.schemes[self.scheme_id]
+            labels = scheme._get_labels()
 
             if labels is None:
                 raise ValueError(
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -226,14 +226,14 @@ class Column(HeaderBase):
                 )
 
             scheme = self._table._db.schemes[self.scheme_id]
-            labels = scheme._get_labels()
+            labels = scheme._labels_to_dict()
 
             if labels is None:
                 raise ValueError(
                     f"Scheme '{self.scheme_id}' has no labels."
                 )
 
-            if isinstance(labels, list):
+            if not any(labels.values()):
                 raise ValueError(
                     f"Scheme '{self.scheme_id}' provides no mapping "
                     "for its labels."
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -1,3 +1,5 @@
+from __future__ import annotations  # allow typing without string
+
 import typing
 
 import numpy as np
@@ -11,8 +13,13 @@ from audformat.core.index import (
     to_array,
 )
 from audformat.core.rater import Rater
-from audformat.core.scheme import Scheme
 from audformat.core.typing import Values
+
+if typing.TYPE_CHECKING:
+    # Fix to make mypy work without circular imports,
+    # compare
+    # https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/
+    from audformat.core.scheme import Scheme  # pragma: nocover
 
 
 def assert_values(
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -188,9 +188,10 @@ class Column(HeaderBase):
         Raises:
             FileNotFoundError: if file is not found
             RuntimeError: if column is not assigned to a table
-            ValueError: if trying to map without a scheme
-            ValueError: if trying to map from a scheme that has no labels
-            ValueError: if trying to map to a non-existing field
+            ValueError: if trying to map without a scheme,
+                or from a scheme that has no labels,
+                or from a scheme that has only a list of labels,
+                or to a non-existing field
 
         """
         if self._table is None:
@@ -230,6 +231,12 @@ class Column(HeaderBase):
             if labels is None:
                 raise ValueError(
                     f"Scheme '{self.scheme_id}' has no labels."
+                )
+
+            if isinstance(labels, list):
+                raise ValueError(
+                    f"Scheme '{self.scheme_id}' has no mappings "
+                    "for its labels."
                 )
 
             mapping = {}
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -235,7 +235,7 @@ class Column(HeaderBase):
 
             if isinstance(labels, list):
                 raise ValueError(
-                    f"Scheme '{self.scheme_id}' has no mappings "
+                    f"Scheme '{self.scheme_id}' provides no mapping "
                     "for its labels."
                 )
 
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -71,7 +71,7 @@ class Scheme(HeaderBase):
         >>> db['speaker'] = audformat.MiscTable(
         ...     pd.Index(['spk1', 'spk2'], name='speaker')
         ... )
-        >>> Scheme(str, labels='speaker')
+        >>> Scheme('str', labels='speaker')
         {dtype: str, labels: speaker}
 
     """
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -45,7 +45,7 @@ class Scheme(HeaderBase):
         ValueError: ``dtype`` does not match type of ``labels``
             if ``labels`` is a list or dictionary
         ValueError: when assigning a scheme,
-            that contains a table ID as ``lables``,
+            that contains a table ID as ``labels``,
             to a database,
             but the corresponding misc table is not part of the database,
             or the given table ID is not a misc table,
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -91,6 +91,7 @@ class Scheme(HeaderBase):
         pd.Timedelta: define.DataType.TIME,
         'date': define.DataType.DATE,
         datetime.datetime: define.DataType.DATE,
+        pd.Timestamp: define.DataType.DATE,
     }
 
     def __init__(
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -48,6 +48,7 @@ class Scheme(HeaderBase):
             that contains a table ID as ``lables``,
             to a database,
             but the corresponding misc table is not part of the database,
+            or the given table ID is not a misc table,
             or its index is multi-dimensional,
             or its index contains duplicates,
             or ``dtype`` doe snot match type of labels
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -35,9 +35,10 @@ class Scheme(HeaderBase):
         BadValueError: if an invalid ``dtype`` is passed
         ValueError: if ``labels`` are not passed as list or dictionary
         ValueError: if ``labels`` are not of same data type
-        ValueError: if ``labels`` is a misc table not assigned to any database
-        ValueError: if ``labels`` is a misc table contain duplicates
-        ValueError: if ``labels`` is a misc table with a multi-dimensional index
+        ValueError: if ``labels`` is a misc table,
+            but the misc table is not assigned to any database,
+            or its index is multi-dimensional,
+            or its index contains duplicates
         ValueError: ``dtype`` does not match type of ``labels``
 
     Example:
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -7,7 +7,6 @@ import pandas as pd
 
 from audformat.core import define
 from audformat.core.common import HeaderBase
-from audformat.core.table import MiscTable
 
 
 class Scheme(HeaderBase):
@@ -17,15 +16,21 @@ class Scheme(HeaderBase):
     ``'bool'``, ``'str'``, ``'int'``, ``'float'``, ``'time'``, and ``'date'``
     (see :class:`audformat.define.DataType`).
     Values can be restricted to a set of labels
-    provided by a list, dictionary or a :class:`audformat.MiscTable`,
+    provided by a list,
+    dictionary
+    or a table ID of a :class:`audformat.MiscTable`,
     for which the labels are given by the index.
     A continuous range can be limited by a minimum and
     maximum value.
 
     Args:
-        dtype: if ``None`` derived from ``labels``, otherwise set to ``'str'``
-        labels: list, dictionary or :class:`audformat.MiscTable`
-            with valid labels
+        dtype: if ``None`` derived from ``labels``,
+            otherwise set to ``'str'``
+        labels: list, dictionary or table ID
+            of a corresponding :class:`audformat.MiscTable`
+            containing labels as index.
+            If a table ID is provided,
+            ``dtype`` has to be specified
         minimum: minimum value
         maximum: maximum value
         description: scheme description
@@ -33,13 +38,20 @@ class Scheme(HeaderBase):
 
     Raises:
         BadValueError: if an invalid ``dtype`` is passed
-        ValueError: if ``labels`` are not passed as list or dictionary
+        ValueError: if ``labels`` are not passed as string, list, or dictionary
+        ValueError: if ``labels`` is a table ID,
+            but ``dtype`` is not specified
         ValueError: if ``labels`` are not of same data type
-        ValueError: if ``labels`` is a misc table,
-            but the misc table is not assigned to any database,
-            or its index is multi-dimensional,
-            or its index contains duplicates
         ValueError: ``dtype`` does not match type of ``labels``
+            if ``labels`` is a list or dictionary
+        ValueError: when assigning a scheme,
+            that contains a table ID as ``lables``,
+            to a database,
+            but the corresponding misc table is not part of the database,
+            or its index is multi-dimensional,
+            or its index contains duplicates,
+            or ``dtype`` doe snot match type of labels
+            from misc table
 
     Example:
         >>> Scheme()
@@ -60,12 +72,8 @@ class Scheme(HeaderBase):
         ... )
         >>> db['speaker']['age'] = audformat.Column(scheme_id='age')
         >>> db['speaker']['age'].set([31, 46])
-        >>> Scheme(labels=db['speaker'])
-        dtype: str
-        labels:
-          levels: [speaker]
-          columns:
-            age: {scheme_id: age}
+        >>> Scheme(labels='speaker', dtype='str')
+        {dtype: str, labels: speaker}
 
     """
     _dtypes = {
@@ -87,7 +95,7 @@ class Scheme(HeaderBase):
             self,
             dtype: typing.Union[typing.Type, define.DataType] = None,
             *,
-            labels: typing.Union[dict, list, MiscTable] = None,
+            labels: typing.Union[dict, list, str] = None,
             minimum: typing.Union[int, float] = None,
             maximum: typing.Union[int, float] = None,
             description: str = None,
@@ -105,15 +113,27 @@ class Scheme(HeaderBase):
 
         if labels is not None:
             self._check_labels(labels)
-            dtype_labels = self._dtype_from_labels(labels)
-            if dtype is not None and dtype != dtype_labels:
-                raise ValueError(
-                    "Data type is set to "
-                    f"'{dtype}', "
-                    "but data type of labels is "
-                    f"'{dtype_labels}'."
-                )
-            dtype = dtype_labels
+
+            if isinstance(labels, str):
+
+                # Labels from misc table
+                if dtype is None:
+                    raise ValueError(
+                        "'dtype' has to be provided "
+                        "when using a misc table as labels."
+                    )
+            else:
+
+                # Labels from list or dictionary
+                dtype_labels = self._dtype_from_labels(labels)
+                if dtype is not None and dtype != dtype_labels:
+                    raise ValueError(
+                        "Data type is set to "
+                        f"'{dtype}', "
+                        "but data type of labels is "
+                        f"'{dtype_labels}'."
+                    )
+                dtype = dtype_labels
 
         self.dtype = dtype
         r"""Data type"""
@@ -229,7 +249,7 @@ class Scheme(HeaderBase):
 
     def replace_labels(
             self,
-            labels: typing.Union[dict, list, MiscTable],
+            labels: typing.Union[dict, list, str],
     ):
         r"""Replace labels.
 
@@ -276,7 +296,7 @@ class Scheme(HeaderBase):
                 'Cannot replace labels when '
                 'scheme does not define labels.'
             )
-        self._check_labels(self.labels)
+        self._check_labels(labels)
 
         dtype_labels = self._dtype_from_labels(labels)
         if dtype_labels != self.dtype:
@@ -290,12 +310,13 @@ class Scheme(HeaderBase):
         self.labels = labels
 
         if self._db is not None and self._id is not None:
+            labels = self._labels_to_list(labels)
             for table in self._db.tables.values():
                 for column in table.columns.values():
                     if column.scheme_id == self._id:
                         y = column._table.df[column._id]
                         y = y.cat.set_categories(
-                            new_categories=self.labels,
+                            new_categories=labels,
                             ordered=False,
                         )
                         column._table.df[column._id] = y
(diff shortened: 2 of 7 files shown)


audformat/core/column.py
```diff
@@ -11,9 +13,14 @@
...
 )
 from audformat.core.rater import Rater
-from audformat.core.scheme import Scheme
 from audformat.core.typing import Values
 
+if typing.TYPE_CHECKING:
+    # Fix to make mypy work without circular imports,
+    # compare
+    # https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/
+    from audformat.core.scheme import Scheme  # pragma: nocover
```
frankenjoe, line 22: nice!

audformat/core/scheme.py
```diff
@@ -65,7 +72,7 @@ def __init__(
             self,
             dtype: typing.Union[typing.Type, define.DataType] = None,
             *,
-            labels: typing.Union[dict, list] = None,
+            labels: typing.Union[dict, list, MiscTable] = None,
```
frankenjoe, line 90: I was expecting we would pass a table ID and not the object itself. Otherwise we cannot know if the table is actually part of the database.

hagenw: We can know, as it contains `self._db`.
I also wanted to start with using the table ID, but for some reason/problem I cannot remember I switched then ;)

But maybe this problem is gone and we can switch back?

frankenjoe: Ok, but since everywhere else we are using IDs when we refer to objects, I think we should stick to this rule here, too.

hagenw: Access to `self._db` we only have after we assigned the scheme to the database. But to assign it, we first need to create it.

frankenjoe: Then we should add `Scheme.db` as with `Table.db`.

hagenw: We have access to `db` inside `Scheme`, but only after we have assigned the scheme to a database.
The only way around would be to first create an empty scheme, assign it to a database and then update the labels with a misc table. Or we could add a `db` argument to `Scheme` which is not very nice.

frankenjoe: > We have access to db inside Scheme, but only after we have assigned the scheme to a database.

I see, we need to access `MiscTable` already in the constructor to derive `dtype`....

frankenjoe: We could add a new `dtype` for `MiscTable`.

frankenjoe: E.g. `dtype='table'`

frankenjoe:
I'm getting an error when I try to assign values to a column that I linked to a scheme that references a `MiscTable`:

```python
index = pd.Index(['spk1', 'spk2', 'spk3'], name='speaker')
db.schemes['scheme'] = audformat.Scheme(labels=db['misc'])
db['table'] = audformat.Table(audformat.filewise_index(['f1', 'f2', 'f3']))
db['table']['speaker'] = audformat.Column(scheme_id='scheme')
db['table']['speaker'].set(['spk1', 'spk3', 'spk3'])
```
```
audformat.core.errors.BadKeyError: "Bad key '0', expected one of []"
```

It seems that the `Scheme.__contains__` is not properly handling `MiscTable` yet.

audformat/core/scheme.py
hagenw: We would also need to remove all checks that the given table is valid and that the table ID exists at all in the database.
In addition, we could not output the YAML of the misc table until the scheme is assigned to a database.

frankenjoe: > We would also need to remove all checks that the given table is valid and that the table ID exists at all in the database.

We can those checks here:  https://github.com/audeering/audformat/blob/ee93219b24c3b428d5ab0ded5f56ab72bb866101/audformat/core/database.py#L1010

> In addition, we could not output the YAML of the misc table until the scheme is assigned to a database.

Which is fine I think, since this is anyway duplicated information.

hagenw:
> I'm getting an error when I try to assign values to a column that I linked to a scheme that references a `MiscTable`:
> 
> ```python
> db = audformat.testing.create_db(minimal=True)
> index = pd.Index(['spk1', 'spk2', 'spk3'], name='speaker')
> db['misc'] = audformat.MiscTable(index)
> db.schemes['scheme'] = audformat.Scheme(labels=db['misc'])
> db['table'] = audformat.Table(audformat.filewise_index(['f1', 'f2', 'f3']))
> db['table']['speaker'] = audformat.Column(scheme_id='scheme')
> db['table']['speaker'].set(['spk1', 'spk3', 'spk3'])
> ```
> 
> ```
> audformat.core.errors.BadKeyError: "Bad key '0', expected one of []"
> ```
> 
> It seems that the `Scheme.__contains__` is not properly handling `MiscTable` yet.

Thanks for spotting this. I was already surprised that all the tests are working, but the problem was that I did not added enough tests targeting the new implementation ;)

Will try to update the tests.

audformat/core/scheme.py
hagenw: Makes indeed sense.

How would you then handle checking for dtypes, and handling of defined dtypes, e.g.
```python
define.DataType.assert_has_attribute_value(dtype)
```
Should we then check for the right dtypes only when the scheme is assigned to the database, or should we simply don't care and go with dtype `table`?

I guess for the YAML representation you would then go simply with
```python
>>> Scheme(labels='table_id')
dtype: table
labels: table_id
```

hagenw: an alternative to `dtype='table'` would be to force a user to provide a `dtype` when assigning a table ID and then check for the correct dtypes in the index when the scheme is assigned to the database.

frankenjoe: > an alternative to dtype='table' would be to force a user to provide a dtype when assigning a table ID and then check for the correct dtypes in the index when the scheme is assigned to the database.

Yes, that also sounds like a good solution.

hagenw: I switched to using a table ID to assign a misc table as scheme and require now indeed that `dtype` is provided in this case.

hagenw:
I guess we should also add a recommendation when to use a misc table as scheme, but I'm not sure yet where to best put it and would not include it into this pull request.

audformat/core/scheme.py
```diff
@@ -156,8 +200,9 @@ def draw(
                 seq = string.ascii_letters + string.digits
                 x = [''.join([random.choice(seq) for _ in range(str_len)])
                      for _ in range(n)]
-        elif type(self.labels) in (list, dict):
-            x = [random.choice(list(self.labels)) for _ in range(n)]
+        else:
+            labels = self._labels_to_list(self.labels)
+            x = [random.choice(labels) for _ in range(n)]
```
hagenw, line 205: BTW, if a misc table scheme is not assigned to a database `draw()` will return an empty list as `self._labels_to_list(self.labels)` returns `[]` as long as the scheme is not assigned to a database.

tests/test_scheme.py
```diff
@@ -20,33 +34,130 @@ def test_scheme():
     assert 'label1' in db.schemes['label_map_str']
     assert 1 in db.schemes['label_map_int']
 
+    # Misc table
+    # assigned scheme
+    assert 'spk1' in db.schemes['speaker']
+    # remove table
+    # db.drop_tables(['misc'])
+    db.misc_tables.pop('misc')
```
hagenw, line 42: `db.drop_tables()` is not working for misc tables, is this on purpose?

I created https://github.com/audeering/audformat/pull/203 to change it.

tests/test_scheme.py
```diff
@@ -5,7 +5,21 @@
...
+    speakers = ['spk1', 'spk2', 'spk3']
+    index = pd.Index(speakers, name='speaker')
+    db['misc'] = audformat.MiscTable(index)
+    db.schemes['scheme'] = audformat.Scheme(labels='misc', dtype='str')
+    db['table'] = audformat.Table(audformat.filewise_index(['f1', 'f2', 'f3']))
+    db['table']['speaker'] = audformat.Column(scheme_id='scheme')
+    db['table']['speaker'].set(speakers)
+
+    assert list(db['table']['speaker'].get()) == speakers
+
```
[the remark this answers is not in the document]
hagenw: OK, so then I would suggest we do nothing here, but add a test and a custom error message in #203 (or maybe another pull request thereafter).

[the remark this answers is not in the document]
frankenjoe: Ok, let's first finish this one and then we add it in #203

audformat/core/testing.py
```diff
@@ -392,4 +395,10 @@ def to_scheme_type(series: pd.Series) -> str:
     db['misc']['gender'] = Column(scheme_id='gender')
     db['misc']['gender'].set(db.schemes['gender'].draw(len(index)))
 
+    ############################
+    # Schemes from Misc Tables #
+    ############################
+
+    db.schemes['speaker'] = Scheme(labels='misc', dtype='str')
```
frankenjoe, line 402: I think we should add a column to `db['files']` and `db['segments']` where we actually define this scheme.

frankenjoe:
The following does not work:

```python

index = pd.Index(['spk1', 'spk2', 'spk3'], name='speaker')
db['misc']['age'] = audformat.Column()
db['misc']['age'].set([33, 44, 55])

db.schemes['scheme'] = audformat.Scheme(
    dtype=audformat.define.DataType.STRING,
    labels='misc',
)
db['table'] = audformat.Table(index)
db['table']['speaker'] = audformat.Column(scheme_id='scheme')
db['table']['speaker'].set(['spk1', 'spk3', 'spk1'])
```
```
AttributeError: 'str' object has no attribute 'items'
```

As far as I saw, the problems is that `Scheme.labels` just returns the ID of the misc table and not the labels.

audformat/core/testing.py
```diff
@@ -320,11 +320,42 @@ def to_scheme_type(series: pd.Series) -> str:
...
+    db['misc']['age'] = Column(scheme_id='age')
+    db['misc']['age'].set(db.schemes['age'].draw(len(index)))
+    db['misc']['gender'] = Column(scheme_id='gender')
+    db['misc']['gender'].set(db.schemes['gender'].draw(len(index)))
+
+    ############################
+    # Schemes from Misc Tables #
+    ############################
+
+    db.schemes['speaker'] = Scheme(labels='misc', dtype='str')
```
frankenjoe, line 358: The other schemes we call after the data type, so maybe we should rename to `label_from_misc`?

frankenjoe: On the other hand, this makes sense since we again map the columns of `map` to `age` and `gender`....

hagenw:
> > For this I implemented `_labels_to_list()` as I thought that `Scheme.labels` should contain what you assigned to the `labels` argument and should not depend on if you have the scheme assigned to a table or not.
> 
> Yes, so I guess you have to look for all places in the code where currently `Scheme.labels` is used and possibly replace it with the helper function.

Turns out that `_labels_to_list()` is not sufficient as we need the labels as a dictinary with `map`. I added `_get_labels()` which returns the labels as a dictionary for misc table and dictionary, or as  a list for list.

BTW, inside `map` the case of list as labels was not covered I added that case and raise an `ValueError` now:

image

audformat/core/testing.py
hagenw: I renamed to `label_map_misc` and use `label` and `int` inside the corresponding misc table.

As we need `label` also for another test as a column in the table, we still assign it to the tables, also you would not expect it for a misc scheme as you would use mapping to get it.

hagenw:
> > I'm not sure if I need to add a test that it works for all dtype and a misc table as data types are differently defined in an index of a table, but maybe this is covered already by the to_pandas_dtype() method?
> 
> I run the following code:
> 
> ```python
> for dtype, values in [
>   (audformat.define.DataType.BOOL, [True, False]),
>   # (
>   #     audformat.define.DataType.DATE,
>   #     pd.to_datetime(['3/11/2000', '3/12/2000', '3/13/2000']),
>   # ),
>   (audformat.define.DataType.INTEGER, [1, 2, 3]),
>   (audformat.define.DataType.FLOAT, [1.0, 2.0, 3.0]),
>   (audformat.define.DataType.STRING, ['a', 'b', 'c']),
>   (audformat.define.DataType.TIME, pd.to_timedelta(['1s', '2s', '3s'])),
> ]:
>     index = pd.Index(values, name='speaker')
>     db['misc'] = audformat.MiscTable(index)
>     index = audformat.filewise_index([f'f{idx}' for idx in range(len(values))])
>     db.schemes['scheme'] = audformat.Scheme(dtype=dtype, labels='misc')
>     db['table'] = audformat.Table(index)
>     db['table']['speaker'] = audformat.Column(scheme_id='scheme')
>     db['table']['speaker'].set(values)
>     print(db['table'].get())
> ```
> 
> ```
>      speaker
> file        
> f0       NaN
> f1       NaN
>      speaker
> file        
> f0         1
> f1         2
> f2         3
>      speaker
> file        
> f0       1.0
> f1       2.0
> f2       3.0
>      speaker
> file        
> f0         a
> f1         b
> f2         c
>              speaker
> file                
> f0   0 days 00:00:01
> f1   0 days 00:00:02
> f2   0 days 00:00:03
> ```
> 
> Except for the boolean case it works as expected. The problem here is that it gets converted to categories `[0, 1]`. But I'm not sure if we really have to support this case since it is not very likely a user will ever create a scheme with boolean as index (it would only allow for two labels anyway).
> 
> More problematic is the date case in comments. It produces:
> 
> ```
> audformat.core.errors.BadValueError: Bad value '<class 'pandas._libs.tslibs.timestamps.Timestamp'>', expected one of ['bool', 'date', 'float', 'int', 'str', 'time']
> ```
> 
> The problem is that currently we only support `datetime.datetime` and not `pd.Timestamp`. The problem is that even when I tried to pass `datetime.datetime` objects it was still converting to a `pd.Timestamp` in the index. So I think we have to add support for `pd.Timestamp`, too. But I suggest we ignore it for the moment and work on it in a separate PR.
> 
> Generally, I would say it's not a bad idea to add tests for all types as it is indeed not straightforward.

Sorry, I overlooked that I should not add it here ;)

I added tests for all cases and implemented also the solution you suggested for `dtype='date'`.
For `dtype='bool'` I now raise a `ValueError` and state that this is not allowed together with a misc table.

frankenjoe:
> Sorry, I overlooked that I should not add it here ;)

No it's fine, when I wrote this I wasn't aware that adding support `pd.Timestamp` was so easy.

audformat/core/column.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations  # allow typing without string
```
frankenjoe, line 1: Why do we nedd this? Is it also required for the trick with the circular imports? Don't really understand the comment, though.

audformat/core/scheme.py
```diff
@@ -15,24 +15,45 @@ class Scheme(HeaderBase):
...
         BadValueError: if an invalid ``dtype`` is passed
-        ValueError: if ``labels`` are not passed as list or dictionary
+        ValueError: if ``labels`` are not passed as string, list, or dictionary
+        ValueError: if ``labels`` is a table ID,
+            but ``dtype`` is not specified
         ValueError: if ``labels`` are not of same data type
         ValueError: ``dtype`` does not match type of ``labels``
+            if ``labels`` is a list or dictionary
+        ValueError: when assigning a scheme,
+            that contains a table ID as ``lables``,
```
frankenjoe, line 48:
```suggestion
            that contains a table ID as ``labels``,
```

audformat/core/scheme.py
```diff
@@ -44,6 +65,17 @@ class Scheme(HeaderBase):
...
         >>> Scheme(float, minimum=0, maximum=1)
         {dtype: float, minimum: 0, maximum: 1}
+        >>> # Misc Table as Scheme
+        >>> import audformat
+        >>> db = audformat.Database('mydb')
+        >>> db.schemes['age'] = Scheme('int')
+        >>> db['speaker'] = audformat.MiscTable(
+        ...     pd.Index(['spk1', 'spk2'], name='speaker')
+        ... )
+        >>> db['speaker']['age'] = audformat.Column(scheme_id='age')
```
frankenjoe, line 75: I think would simplify the example and use a `MiscTable` without a column.

audformat/core/scheme.py
```diff
@@ -15,24 +15,45 @@ class Scheme(HeaderBase):
     Allowed values for ``dtype`` are:
     ``'bool'``, ``'str'``, ``'int'``, ``'float'``, ``'time'``, and ``'date'``
     (see :class:`audformat.define.DataType`).
-    Values can be restricted to a set of labels provided by a
-    list or a dictionary.
+    Values can be restricted to a set of labels
+    provided by a list,
+    dictionary
+    or a table ID of a :class:`audformat.MiscTable`,
+    for which the labels are given by the index.
```
frankenjoe, line 22: Maybe:

```suggestion
    where the values of the index are used as labels.
```

audformat/core/scheme.py
```diff
@@ -44,6 +65,17 @@ class Scheme(HeaderBase):
         {dtype: int}
         >>> Scheme(float, minimum=0, maximum=1)
         {dtype: float, minimum: 0, maximum: 1}
+        >>> # Misc Table as Scheme
```
frankenjoe, line 68:
```suggestion
        >>> # Use index of misc table as labels
```

audformat/core/column.py
frankenjoe: But here we define the `Column` class so why do we have to write 'Scheme'?

audformat/core/scheme.py
frankenjoe: Yes, I think this is something we should cover in the usage section, for this example I find it too complicated.

audformat/core/column.py
hagenw: So we need it together with

```python
if typing.TYPE_CHECKING:
    # Fix to make mypy work without circular imports,
    # compare
    # https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/
    from audformat.core.scheme import Scheme  # pragma: nocover
```

So maybe I add a remark in the comment that it belongs to this part?

frankenjoe: Ah ok, I think I get it. Because we import `Scheme` later to avoid circulate imports, we would need to put it in strings. And that we avoid with the future import.

frankenjoe: Now that I understand how it works, I'd say it's fine :)

audformat/core/scheme.py
hagenw: OK, I removed the column and the `age` scheme and updated the description of this pull request.

tests/test_scheme.py
```diff
@@ -188,3 +346,24 @@ def test_replace_labels(values, labels, new_labels, expected):
...
+    db['table']['columns'] = audformat.Column(scheme_id='scheme')
+    db['table']['columns'].set(['spk1', 'spk2'])
+
+    db['misc-new'] = audformat.MiscTable(
+        pd.Index(['spk1', 'spk2', 'spk3'], name='speaker')
+    )
+    db.schemes['scheme'].replace_labels('misc-new')
+
+    # Using not assigned scheme
+    scheme.replace_labels('misc-not-assigned')
```
frankenjoe, line 369: What exactly are we testing here?

hagenw: We are testing replacing the scheme with a non-existing misc table.
I updated the table name and added several `assert` statements to the test.

tests/test_scheme.py
```diff
@@ -188,3 +346,24 @@ def test_replace_labels(values, labels, new_labels, expected):
...
+    scheme = audformat.Scheme(labels='misc', dtype='str')
+    db.schemes['scheme'] = scheme
+    db['table'] = audformat.Table(index=audformat.filewise_index(['f1', 'f2']))
+    db['table']['columns'] = audformat.Column(scheme_id='scheme')
+    db['table']['columns'].set(['spk1', 'spk2'])
+
+    db['misc-new'] = audformat.MiscTable(
+        pd.Index(['spk1', 'spk2', 'spk3'], name='speaker')
+    )
+    db.schemes['scheme'].replace_labels('misc-new')
```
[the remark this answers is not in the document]
hagenw: I added an additional test for replacing by a dictionary and also by a list

tests/test_scheme.py
```diff
@@ -5,7 +5,25 @@
...
+    db['misc'] = audformat.MiscTable(index)
+    db['misc']['age'] = audformat.Column()
+    db['misc']['age'].set(ages)
+    db.schemes['scheme'] = audformat.Scheme(labels='misc', dtype='str')
+    db['table'] = audformat.Table(audformat.filewise_index(['f1', 'f2', 'f3']))
+    db['table']['speaker'] = audformat.Column(scheme_id='scheme')
+    db['table']['speaker'].set(speakers)
+
+    assert list(db['table']['speaker'].get()) == speakers
+    assert list(db['table']['speaker'].get(map='age')) == ages
```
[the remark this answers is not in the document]
hagenw: Good idea, I added that test as well.

audformat/core/scheme.py
```diff
@@ -301,6 +361,33 @@ def _dtype_from_labels(
 
         return dtype
 
+    def _get_labels(
```
frankenjoe, line 364: Why do we need two functions: `_get_labels()` and `_labels_to_list()`? And why is `_get_labels()` sometimes returning a dictionary and sometimes a list?

hagenw: `_get_labels()` is trying to copy the behavior of `Scheme.labels` that we had before, which was returning a list for list labels and a dict for dict labels. I now added the case of a misc table, which is returned as a dict, which we need for applying `map` in `Column`.

At several places we used `list(Scheme.labels)` before. This could in principle be replaced by `list(_get_labels())`, the problem is that inside `replace_labels()` we also need it for labels that are not assigned to the scheme yet, hence `_labels_to_list(labels)` solves this issue.

One solution might be to change `_get_labels(self)` to `_get_labels(self, labels)`, but this also looks strange.

Any other ideas how to make this nicer?

tests/test_scheme.py
```diff
@@ -188,3 +348,55 @@ def test_replace_labels(values, labels, new_labels, expected):
...
+    assert scheme.labels == 'misc'
+    assert scheme._get_labels() == {'spk1': {}, 'spk2': {}}
+
+    # replace with list
+    scheme.replace_labels(['spk1', 'spk2'])
+    assert scheme.labels == ['spk1', 'spk2']
+    assert scheme._get_labels() == ['spk1', 'spk2']
+
+    # replace with non-existing misc table scheme
+    scheme.replace_labels('misc-non-existing')
```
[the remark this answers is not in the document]
hagenw: In principle we would need to apply all the checks we do when assigning a scheme with misc table to a database.

The other problem is that this method also has to work when the scheme was not yet assigned to any database, that's the reason why I selected the current behavior that works in both cases, but has indeed some unwanted consequences when replacing misc table labels in a scheme that is already assigned to a database.

tests/test_scheme.py
```diff
@@ -188,3 +348,55 @@ def test_replace_labels(values, labels, new_labels, expected):
...
+    assert scheme.labels == 'misc-new'
+    assert scheme._get_labels() == {'spk1': {}, 'spk2': {}, 'spk3': {}}
+
+    # replace with dictionary
+    scheme.replace_labels({'spk1': {}})
+    assert scheme.labels == {'spk1': {}}
+    assert scheme._get_labels() == {'spk1': {}}
+    assert list(db['table']['columns'].get().values) == ['spk1', np.NaN]
+
+    # replace again with mis table
```
frankenjoe, line 389:
```suggestion
    # replace again with misc table
```

tests/test_scheme.py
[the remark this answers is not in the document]
frankenjoe: And what speaks again applying the checks in case the scheme is assigned to a database?

audformat/core/scheme.py
frankenjoe: > _get_labels() is trying to copy the behavior of Scheme.labels that we had before, which was returning a list for list labels and a dict for dict labels. I now added the case of a misc table, which is returned as a dict, which we need for applying map in Column.

But if I look at the following two tests, wouldn't it be more consistent if we return also a list if the misc table does not have columns?

```
    # replace again with mis table
    scheme.replace_labels('misc')
    assert scheme.labels == 'misc'
    assert scheme._get_labels() == {'spk1': {}, 'spk2': {}}

    # replace with list
    scheme.replace_labels(['spk1', 'spk2'])
    assert scheme.labels == ['spk1', 'spk2']
    assert scheme._get_labels() == ['spk1', 'spk2']
```

frankenjoe: > Any other ideas how to make this nicer?

Not really.

frankenjoe: >   But if I look at the following two tests, wouldn't it be more consistent if we return also a list if the misc table does not have columns?

Or as an alternative, always a dictionary. Might be easier to work with if the return value is always of the same type.

frankenjoe: Ok, cool. I think it's fine to stay with both functions.

audformat/core/scheme.py
```diff
@@ -301,6 +403,29 @@ def _dtype_from_labels(
 
         return dtype
 
+    def _labels_to_dict(
+            self,
+            labels: typing.Union[dict, list, str] = None,
+    ) -> typing.Dict:
+        r"""Return actual labels as list or dict."""
```
frankenjoe, line 410:
```suggestion
        r"""Return actual labels as dict."""
```

PR #203, opened 2022-07-07
Extend pick and drop from Database to misc tables

hagenw:
Make `Database.drop_tables()` and `Database.pick_tables()` work with misc tables as well.

Before we did not really had a test for what should happen when a non-existing table ID is provided.
It fails with a `KeyError` in the current master:
```
>>> db = audformat.testing.create_db()
>>> db.drop_tables('non')
...                                        
--> 326     self.tables.pop(table_id)   
                                        
KeyError: 'non'                          
```

At the moment I changed it to silently ignore missing table IDs. If you don't like that approach we should readd the `KeyError`, but with a custom error message or raise a custom `audformat.errors.BadIdError`.

image

image


frankenjoe:
> At the moment I changed it to silently ignore missing table IDs.

I think we should raise an error as it most likely means that the user made a typo in the code.

frankenjoe:
In https://github.com/audeering/audformat/pull/183#issuecomment-1127318584 we had a discussion if `drop_tables() / pick_tables()` should support misc tables. I think back then you voted they should not. So I guess you changed your mind then :)

Once we have #201 should we check if a misc table is used in a scheme and in that case raise an error it cannot be dropped?

frankenjoe:
I was expecting we can remove a misc table again after assign the scheme to a different misc table. But I'm getting:

```python

# misc table with gender labels
index = pd.Index(['male', 'female'], name='gender')
db['gender'] = audformat.MiscTable(index)

# gender scheme
db.schemes['gender'] = audformat.Scheme(
    define.DataType.STRING,
    labels='gender',
)

# create table with column that references gender scheme
db['files'] = audformat.Table(audformat.filewise_index(['f1', 'f2']))
db['files']['gender'] = audformat.Column(scheme_id='gender')
db['files']['gender'].set(['male', 'female'])

# misc table with new gender labels
index = pd.Index(['male', 'female', 'child'], name='gender')
db['gender-with-child'] = audformat.MiscTable(index)

# assign new gender labels
db.schemes['gender'].replace_labels('gender-with-child')

# drop misc table with old gender labels
db.drop_tables('gender')
```
```
RuntimeError: Misc table 'gender' is used as scheme(s): 'gender', and cannot be removed.
```

audformat/core/database.py
```diff
@@ -314,16 +314,36 @@ def drop_tables(
...
         """
-        if isinstance(table_ids, str):
-            table_ids = [table_ids]
+        table_ids = audeer.to_list(table_ids)
         for table_id in table_ids:
-            self.tables.pop(table_id)
+            if table_id in self.tables:
+                self.tables.pop(table_id)
+            elif table_id in self.misc_tables:
+                schemes = self.misc_tables[table_id]._used_in_schemes()
```
frankenjoe, line 334: Or can't we simply do:

```python
                schemes = [
                    scheme._id for scheme in self.schemes.values()
                    if scheme.labels == table_id
                ]
```

frankenjoe:
> I was expecting we can remove a misc table again after assign the scheme to a different misc table.

See the two comment I left as possible solutions. We should probably extend the test to check if a table can be removed after labels were updated.

audformat/core/database.py
hagenw: Very good idea, I removed `_used_in_schemes()` and use your solution now.

hagenw:
I extended the tests by assigning a different misc table as scheme and remove the original misc table afterwards.

audformat/core/database.py
```diff
@@ -487,18 +510,25 @@ def pick_tables(
...
 
         Args:
             table_ids: table IDs to pick
 
+        Raises:
+            audformat.errors.BadIdError: if a table with provided ID
+                does not exist in the database
+            RuntimeError: if a misc table
+                that is used as scheme(s)
+                is part of the non-picked tables
```
frankenjoe, line 523: Sounds a bit strange. Maybe: "if a misc table that is used in a scheme would be removed?"

audformat/core/database.py
```diff
@@ -314,16 +314,39 @@ def drop_tables(
...
+        r"""Drop (miscellaneous) tables by ID.
 
         Args:
             table_ids: table IDs to drop
 
+        Raises:
+            audformat.errors.BadIdError: if a table with provided ID
+                does not exist in the database
+            RuntimeError: if a misc table with provided ID
+                is used as scheme(s)
```
frankenjoe, line 326:
```suggestion
                is used in a scheme
```

frankenjoe:
I left two minor comments, otherwise this looks good now!

audformat/core/database.py
hagenw: Changed accordingly.

frankenjoe:
Could you please update the description, then we can merge I think.

PR #205, opened 2022-07-15
Add utils.is_index_alike()

frankenjoe:
To implement the remaining methods like `MiscTable.update()` etc. we will need some new utility functions. This adds a first function `utils.is_index_alike()` that checks if the indices have the same level names and dtypes. Surprisingly, it seems no such function exists in `pandas`. 

~~I think we should keep them separate from the utility functions that expect objects conform to `audformat`, so I added the function under `utils.misc`~~

image


frankenjoe:
We should probably rename to `is_index_compatible()` or `is_index_alike()` to express that the function returns a boolean value.

audformat/core/utils_misc.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    objs = [obj if isinstance(obj, pd.Index) else obj.index for obj in objs]
+
+    # check names
+    levels = set([obj.names for obj in objs])
+    if len(levels) > 1:
+        return False
+
+    # check dtypes
+    dtypes = set()
```
frankenjoe, line 50: We should probably wait for #204 and convert the dtype with `to_audformat_dtype()` before doing the comparison.

frankenjoe: done, also added a test to ensure that `int` and `Int64` are not treated as different `dtype`.

hagenw:
> We should probably rename to `is_index_compatible()` or `is_index_alike()` to express that the function returns a boolean value.

Maybe `is_index_alike()` as compatible sounds a little bit like compatible with audformat definition to me.

I would also not go for `audformat.utils.misc`. If you think the user should have access to the function and it should not be placed under `audformat.utils`, I would then create another top-level sub-group like `audformat.misc`

hagenw:
Something seems to be broken on Github.
If I look for changes I see stuff like

image

but this was removed already in the master

frankenjoe:
> but this was removed already in the master

I rebased once again, seems to be fixed now.

frankenjoe:
> I would also not go for `audformat.utils.misc`. If you think the user should have access to the function and it should not be placed under `audformat.utils`, I would then create another top-level sub-group like `audformat.misc`

I agree that maybe we don't have to expose this function to the user. But we will definitely need according fucntions to `utils.concat()` and `utils.union()` and I though it might be a good idea to still have them under `utils`.

frankenjoe:
> Maybe `is_index_alike()` as compatible sounds a little bit like compatible with audformat definition to me.

I renamed it accordingly.

frankenjoe:
The other option I see is adding 

> I agree that maybe we don't have to expose this function to the user. But we will definitely need according fucntions to `utils.concat()` and `utils.union()` and I though it might be a good idea to still have them under `utils`.

Another option would be `utils.concat_misc()` and ``utils.union_misc()`.

frankenjoe:
I now moved the function to `utils.is_index_alike()`.

audformat/core/utils.py
```diff
@@ -12,11 +12,12 @@
 import audeer
 import audiofile
 
+from audformat.core import common
```
hagenw, line 15: Would also be fine with me if you do
```python
from audformat.core.common import to_audformat_dtype
```
as we usually do.

The main reason I did not do it inside `audformat.core.scheme` was that there we define a method that is also called `to_pandas_dtype()`, and I thought `common.to_pandas_dtype()` will then be more obvious to differentiate them.

hagenw (commented):
Great to see that this can be implemented in a straightforward way.

audformat/core/utils.py
frankenjoe: Ok, changed it accordingly.

audformat/core/utils.py
```diff
@@ -490,6 +491,63 @@ def intersect(
...
 
 
+def is_index_alike(
+        objs: typing.Sequence[typing.Union[pd.Index, pd.Series, pd.DataFrame]],
+) -> bool:
+    r"""Check if index objects are alike.
+
+    Two index objects are alike
+    if they have the same number of levels
+    and share the same level names and dtypes.
```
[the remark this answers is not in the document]
frankenjoe: Ah good point. I already changed the docstring example to show that there is no difference between different `int` representation, but it's indeed a good idea to also explicitly mention it in the docstring.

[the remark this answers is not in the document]
hagenw: Cool, can you also update the screenshot in the description of the pull request, thanks.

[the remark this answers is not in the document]
frankenjoe: done

PR #208, opened 2022-07-20
Add a code comment for dtype argument in Schemes

### LINKED ISSUE #206: Restrict input of dtype for audformat.Scheme?
In the documentation for `audformat.Scheme` we state:

image

But in the code we allow a lot more:

```python
    _dtypes = { 
        'bool': define.DataType.BOOL,
        bool: define.DataType.BOOL,
        'str': define.DataType.STRING,
        str: define.DataType.STRING,
        'int': define.DataType.INTEGER,
        int: define.DataType.INTEGER,
        'float': define.DataType.FLOAT,
        float: define.DataType.FLOAT,
        'time': define.DataType.TIME,
        pd.Timedelta: define.DataType.TIME,
        'date': define.DataType.DATE,
        datetime.datetime: define.DataType.DATE,
        pd.Timestamp: define.DataType.DATE,
    }
    dtype = self._dtypes[dtype]
```

It might be convenient to have it, but I'm wondering if we then should document it?

hagenw:
Closes #206 

This adds a comment to highlight that the `_dtypes` variable in the code allows to support more than the documented input types for the `dtype` argument of `audformat.Schemes`

### DIFF
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -75,6 +75,10 @@ class Scheme(HeaderBase):
         {dtype: str, labels: speaker}
 
     """
+    # Mapping for dtype input argument,
+    # e.g. to allow `str` besides `'str'`.
+    # This behavior is only for convinience
+    # and not mentioned in the docstring
     _dtypes = {
         'bool': define.DataType.BOOL,
         bool: define.DataType.BOOL,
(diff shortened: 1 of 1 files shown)


audformat/core/scheme.py
```diff
@@ -75,6 +75,10 @@ class Scheme(HeaderBase):
         {dtype: str, labels: speaker}
 
     """
+    # Mapping for dtype input argument,
+    # e.g. to allow `str` besides `'str'`.
+    # This behavior is only for convinience
```
frankenjoe, line 80:
```suggestion
    # This behavior is only for convenience
```

PR #212, opened 2022-07-20
Add audformat.define.DataType.OBJECT

### LINKED ISSUE #210: Support for pd.StringDType
In 1.3 Pandas has introduced [pd.StringDType](https://pandas.pydata.org/docs/reference/api/pandas.StringDtype.html) as dtype for string data. It's still experimental, but could be very useful in the context of annotation where we usually have short and repeating strings:

image

and also in the [documentation](https://pandas.pydata.org/docs/user_guide/text.html) they already suggest to use it:

image

Though it's still marked as experimental features we can expect that it will become the standard to store string data int the future and think about ways to support it.

### LINKED ISSUE #211: Add define.DataType.Object
As mentioned in https://github.com/audeering/audformat/pull/204#issuecomment-1190276787 with `MiscTable` we currently map levels with type `object` (which might contain strings or something else) always to our internal type `define.DataType.String`, which can lead to the following situation:

```python
index = pd.Index([[0, 0], [1, 1]], name='list')
audformat.MiscTable(index)
```
```
levels: {list: str}
```

As noted in #210 there is now a string type in Pandas that would allow us to differentiate between levels that contain string or generic object data. So maybe we should introduce `define.DataType.Object` and in the example above output:

```
levels: {list: object}
```

We would then only show `{list: str}` if the level is explicitly set to `dtype=string`.

hagenw:
Closes #210
Closes #211 

This adds the audformat dtype `audformat.define.DataType.OBJECT` and switches the use-case for `audformat.define.DataType.STRING` to objects that use the `string` dtype in `pandas`.

image

image

---

We should maybe thinking about moving `audformat.core.common.set_index_dtype()` to the public API as there is no easy way for a user to set the dtype when creating a `MultiIndex`, at least none that I'm aware of.

This means currently we use `'object'` dtype in our example for the `MiscTable` definition:

image

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -530,6 +530,20 @@ class Base(HeaderBase):
         except pickle.UnpicklingError:
             df = pd.read_pickle(path, compression='xz')
 
+        # Older versions of audformat stored columns
+        # assigned to a string scheme as 'object',
+        # so we need to convert those to 'string'
+        for column_id, column in self.columns.items():
+            if (
+                column.scheme_id is not None
+                and (
+                    self.db.schemes[column.scheme_id].dtype
+                    == define.DataType.STRING
+                )
+                and df[column_id].dtype == 'object'
+            ):
+                df[column_id] = df[column_id].astype('string', copy=False)
+
         self._df = df
 
     def _save_csv(self, path: str):
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -624,7 +624,7 @@ class MiscTable(Base):
         ... )
         >>> table['match'] = Column()
         >>> table
-        levels: {file: str, other: str}
+        levels: {file: object, other: object}
         split_id: test
         columns:
           match: {}
--- a/tests/test_misc_table.py
+++ b/tests/test_misc_table.py
@@ -685,3 +685,31 @@ def test_level_and_column_names(index, columns):
     misc = audformat.MiscTable(index)
     for column in columns:
         misc[column] = audformat.Column()
+
+
+def test_load_old_pickle(tmpdir):
+    # We have stored string dtype as object dtype before
+    # and have to fix this when loading old PKL files from cache.
+    # This does only affect columns
+    # as there was no MiscTable available.
+
+    # Create PKL file containing strings as object
+    y = pd.Series(['c'], dtype='object', name='column')
+    index = pd.Index(['i'], dtype='object', name='idx')
+
+    db = audformat.testing.create_db(minimal=True)
+    db['misc'] = audformat.MiscTable(index)
+    db.schemes['column'] = audformat.Scheme(audformat.define.DataType.OBJECT)
+    db['misc']['column'] = audformat.Column(scheme_id='column')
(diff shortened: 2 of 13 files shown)


tests/test_misc_table.py
```diff
@@ -219,6 +225,13 @@ def test_dtype_column(
             [],
             str,
             'object',
+            audformat.define.DataType.OBJECT,
```
hagenw, line 228: This is a little bit unfortunate, but as long as `pandas` decide that the following is correct we cannot change it:

```python
>>> pd.Series([], dtype=str)
Series([], dtype: object)
```

and need to force the users to use

```python
>>> pd.Series([], dtype='string')
Series([], dtype: string)
```

instead.

frankenjoe:
Cool, at a first glance it does not involve too many changes, just like I was hoping for! And we finally can distinguish strings from other objects, which makes a lot of sense I think. Will take a deeper look now :)

tests/test_misc_table.py
frankenjoe: Yes, I also don't see another solution for it. We should probably add to `audformat.MiscTable.__init__()` that a user should explicitly use `dtype='string'` if a level contains string values as otherwise the type will be set to `object`.

frankenjoe:
Looks very good, I think there is only one special case we might want to handle. And that is loading a table from a pickle file that has a column assigned to a string scheme. With the new version it would be stored with `dtype=string`, but in older version it was still stored with `dtype=object`. I think we should convert it to `dtype=string` in that case, too, cause this is also the type we will see when loading it from CSV. I did a quick check at the following was working for me:

```python
    def _load_pickled(self, path: str):

        # Older versions of audformat used xz compression
        # which produced smaller files,
        # but was slower.
        # The try-except statement allows backward compatibility
        try:
            df = pd.read_pickle(path)
        except pickle.UnpicklingError:
            df = pd.read_pickle(path, compression='xz')

        # older versions of audformat stored columns
        # assigned to a string scheme as 'object',
        # so we need to convert those to 'string'
        for column_id, column in self.columns.items():
            if (
                column.scheme_id is not None
                and (
                    self.db.schemes[column.scheme_id].dtype ==
                    define.DataType.STRING
                )
                and df[column_id].dtype == 'object'
            ):
                df[column_id] = df[column_id].astype('string')

        self._df = df
```

frankenjoe:
We should probably change the name of this PR to `Add audformat.define.DataType.OBJECT and use pd.StringDType with audformat.define.DataType.STRING'

hagenw:
Good point with the old PKL files from cache.

I added a test for them, which failed until I implemented the solution you proposed here.

audformat/core/table.py
```diff
@@ -624,7 +638,7 @@ class MiscTable(Base):
         ... )
         >>> table['match'] = Column()
         >>> table
-        levels: {file: str, other: str}
+        levels: {file: object, other: object}
```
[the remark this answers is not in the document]
frankenjoe: Ok, I created https://github.com/audeering/audformat/issues/217

PR #213, opened 2022-07-21
Ensure that column ID is differnt from index level names

### LINKED ISSUE #209: Make sure index levels and table column names to not overlap for tables
As discussed at https://github.com/audeering/audformat/pull/204#discussion_r925575049 we have to ensure that the names for the index levels and dataframe columns to not overlap for tables.

frankenjoe:
Closes #209 

image

image

image

### Example

```python
import audformat

index = pd.Index([1, 2, 3], name='name')
misc = audformat.MiscTable(index)
misc[index.name] = audformat.Column()
```
```
ValueError: Cannot add column with ID 'name' when there is already an index level with same name. Level names are: ['name'].
```

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -557,7 +557,7 @@ class Base(HeaderBase):
             raise ValueError(
                 f"Cannot add column with ID "
                 f"'{column_id}' "
-                f"when there is already an "
+                f"when there is an "
                 f"index level with same name. "
                 f"Level names are: "
                 f"{levels}."
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -90,6 +90,7 @@ class Base(HeaderBase):
         Raises:
             BadIdError: if a column with a ``scheme_id`` or ``rater_id`` is
                 added that does not exist
+            ValueError: if column ID is not different from level names
 
         """
         self.columns[column_id] = column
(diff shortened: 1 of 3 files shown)


audformat/core/table.py
```diff
@@ -547,6 +548,21 @@ def _save_pickled(self, path: str):
...
+        levels = (
+            self.index.names
+            if isinstance(self.index, pd.MultiIndex)
+            else [self.index.name]
+        )
+        if column_id in levels:
+            raise ValueError(
+                f"Cannot add column with ID "
+                f"'{column_id}' "
+                f"when there is already an "
```
hagenw, line 560:
```suggestion
                f"when there is an "
```

tests/test_table.py
```diff
@@ -648,6 +648,32 @@ def test_exceptions():
         )
         db['table']['column'] = audformat.Column(rater_id='invalid')
 
+    # level and column names must be unique
+
```
hagenw, line 652: Could you add matches for the error message here, so we make sure not a different `ValueError` is raised

hagenw (commented):
Great.

Interesting that we never added the test before, but I guess we did nor really try to use `file`, `start`, `end` as column names.

tests/test_table.py
frankenjoe: I tried with:

```
            marks=pytest.mark.xfail(
                raises=ValueError,
                match='index level with same name',
            ),
```

but it does seem to have an effect as I can put anything with `match=...` and it still works.

hagenw: No, not with `marks`, I'm talking about the tests after this line that all start with `with pytest.raises(ValueError)`, there you can add `with pytest.raises(ValueError, match=error_msg)`

PR #214, opened 2022-07-21
WIP: audformat.utils.union() supports all indices

hagenw:
Expands `audformat.utils.union()` to handle not only filewise or segmented indices, but all others as well.
This will be helpful to concat two misc tables.

This means `audformat.utils.union()` no longer raises a `ValueError` if an index is provided that is not conform to `audformat`.
If any segmented index is present, all filewise indices are still converted to segmented indices before combining them with all other indices.

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1206,10 +1206,11 @@ def union(
 ) -> pd.Index:
     r"""Create union of index objects.
 
-    Index objects must be conform to
-    :ref:`table specifications <data-tables:Tables>`.
+    Filewise indices will be first converted to segmented indeces
+    if a segmented index is given as well.
 
-    If at least one object is segmented, the output is a segmented index.
+    If ``objs`` is an empty list,
+    an empty filewise index is returned.
 
     Args:
         objs: index objects conform to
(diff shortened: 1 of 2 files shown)


hagenw:
I think it make sense to try to expand our existing `utils` functions that we need for misc tables to support not only indices conform to audformat.

If you agree, I will add tests to this pull request.
We have already full code coverage due to the single example I added to the docstring.

audformat/core/utils.py
```diff
@@ -1255,21 +1250,38 @@ def union(
...
+            if is_filewise_index(obj):
+                objs[n] = to_segmented_index(obj)
 
     # Combine all MultiIndex entries and drop duplicates afterwards,
     # faster than using index.union(),
     # compare https://github.com/audeering/audformat/pull/98
     df = pd.concat([o.to_frame() for o in objs])
-    index = df.index
+    if len(df.columns) > 1:
+        index = pd.MultiIndex.from_frame(df)
```
hagenw, line 1282: It might be that this will slow down combining only segmented indices, as before we could simply use the index returned by `pd.concat()`. So maybe we are faster if we add another check if all indices are segmented and use `index = df.index` in that case.

hagenw: I guess, I should also do some execution time testing for this.

frankenjoe:
> I think it make sense to try to expand our existing utils functions that we need for misc tables to support not only indices conform to audformat.

I am not sure if this is really a good idea:

1. it will no longer raise an error if the input is not audformat
2. you cannot count that the result is audformat
3. it's strange that we return an empty filewise index if no index is provided
4. usually we would expect that indices have to be alike, but with audformat we have the special rule that filewise and segmented can be combined

My idea was therefore to add a new function e.g. `union_misc()`. Here we first check that all input indices are alike (we have already added a utility function for it), then we can move most of the code of `union()` to it and in `union()` we only keep the code where we possibly convert `filewise` to `segmented` index and then call `union_misc()`.

frankenjoe:
To demonstrate the idea I created https://github.com/audeering/audformat/pull/215

frankenjoe:
I still think it's too dangerous that we change the behavior of this and other utility function. It introducing at least two breaking changes - not raising an error it the input is not audformat and not returning an index that is audformat, this might have unwanted effects in other libraries like `audinterface` were we rely on error messages raised if the input is not audformat or combining the results to something that is again audformat.

hagenw:
OK, I see the danger.

Then let's continue with https://github.com/audeering/audformat/pull/215.

Maybe you can take over the examples from here

frankenjoe:
> Maybe you can take over the examples from here

I already added some docstring examples there, please check if you see the need to add more from here.

hagenw:
Closing in favor of #215

PR #215, opened 2022-07-21
Add utils.union_misc()

frankenjoe:
Alternative implementation to #214 

image

image

For consistency the following docstring examples are also changed:

image

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -6,6 +6,9 @@ import typing
 
 import audiofile
 import oyaml as yaml
+
+import audformat.utils
+
 try:
     from yaml import CLoader as Loader
 except ImportError:  # pragma: nocover
@@ -463,9 +466,14 @@ class Database(HeaderBase):
         """
         def job(table):
             if table.is_segmented:
-                table.df.index = table.df.index.map(
+                index = table.df.index.map(
                     lambda x: (func(x[0]), x[1], x[2])
                 )
+                index = utils.set_index_dtypes(
+                    index,
+                    {define.IndexField.FILE: 'string'},
+                )
+                table.df.index = index
             else:
                 table.df.index = table.df.index.map(lambda x: func(x))
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1438,6 +1438,9 @@ def union_misc(
     if not objs:
         return pd.Index([])
 
+    if len(objs) == 1:
+        return objs[0]
+
     if not is_index_alike(objs):
         raise ValueError(
             'Levels and dtypes of all objects must match, '
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1339,37 +1339,33 @@ def union(
             :ref:`table specifications <data-tables:Tables>`
 
     Example:
-        >>> index1 = filewise_index(['f1', 'f2', 'f3'])
-        >>> index2 = filewise_index(['f2', 'f3', 'f4'])
-        >>> union([index1, index2])
-        Index(['f1', 'f2', 'f3', 'f4'], dtype='string', name='file')
-        >>> index3 = segmented_index(
-        ...     ['f1', 'f2', 'f3', 'f4'],
-        ...     [0, 0, 0, 0],
-        ...     [1, 1, 1, 1],
+        >>> union(
+        ...     [
+        ...         filewise_index(['f1', 'f2', 'f3']),
+        ...         filewise_index(['f2', 'f3', 'f4']),
+        ...     ]
         ... )
-        >>> index4 = segmented_index(
-        ...     ['f1', 'f2', 'f3'],
-        ...     [0, 0, 1],
-        ...     [1, 1, 2],
+        Index(['f1', 'f2', 'f3', 'f4'], dtype='object', name='file')
+        >>> union(
+        ...     [
+        ...         segmented_index(['f2'], [0], [1]),
+        ...         segmented_index(['f1', 'f2'], [0, 1], [1, 2]),
+        ...     ]
         ... )
-        >>> union([index3, index4])
-        MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:01'),
-                    ('f2', '0 days 00:00:00', '0 days 00:00:01'),
-                    ('f3', '0 days 00:00:00', '0 days 00:00:01'),
-                    ('f4', '0 days 00:00:00', '0 days 00:00:01'),
-                    ('f3', '0 days 00:00:01', '0 days 00:00:02')],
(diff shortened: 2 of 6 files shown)


hagenw:
You need to convert MultiIndex with single levels to Index before combining them.

At the moment

```python
audformat.utils.union_misc(
    [
        pd.Index(['spk1'], name='speaker'),
        pd.MultiIndex.from_arrays([['spk2']], names=['speaker']),
    ]
)
```
returns
```
Index(['spk1', ('spk2',)], dtype='object')
```
but it should return
```
Index(['spk1', 'spk2'], dtype='object', name='speaker')
```

frankenjoe:
> You need to convert MultiIndex with single levels to Index before combining them.

Very good observation. Before I fix this and add tests for it, let's wait which path we will take, https://github.com/audeering/audformat/pull/214 or this one.

audformat/core/utils.py
```diff
@@ -1265,6 +1267,60 @@ def union(
...
+        >>> index4 = pd.MultiIndex.from_arrays(
+        ...     [['b', 'c'], [1, 3]],
+        ...     names=['idx1', 'idx2'],
+        ... )
+        >>> union_misc([index3, index4])
+        MultiIndex([('a', 0),
+                    ('b', 1),
+                    ('c', 2),
+                    ('c', 3)],
+                   names=['idx1', 'idx2'])
```
hagenw, line 1312: I would skip the definition of the index entries as it makes it easier to compare them.
Hence, I would update the examples to:

```python
        >>> union(
        ...     [
        ...         pd.Index([0, 1], name='idx'),
        ...         pd.Index([1, 2], dtype='Int64', name='idx'),
        ...     ]
        ... )
        Index([0, 1, 2], dtype='Int64', name='idx')
        >>> union(
        ...     [
        ...         pd.MultiIndex.from_arrays(
        ...             [['a', 'b', 'c'], [0, 1, 2]],
        ...             names=['idx1', 'idx2'],
        ...         ),
        ...         pd.MultiIndex.from_arrays(
        ...             [['b', 'c'], [1, 3]],
        ...             names=['idx1', 'idx2'],
        ...         ),
        ...    ]
        ... )
        MultiIndex([('a', 0),
                    ('b', 1),
                    ('c', 2),
                    ('c', 3)],
                   names=['idx1', 'idx2'])
        >>> union(
        ...     [
        ...         pd.Index(['spk1'], name='speaker'),
        ...         pd.MultiIndex.from_arrays([['spk2']], names=['speaker']),
        ...     ]
        ... )
        Index(['spk1', 'spk2'], dtype='object', name='speaker')
```

To be in line I would further simplify the docstring examples of `audformat.utils.union()` to:

```python
        >>> union(
        ...     [
        ...         filewise_index(['f1', 'f2', 'f3']),
        ...         filewise_index(['f2', 'f3', 'f4']),
        ...     ]
        ... )
        Index(['f1', 'f2', 'f3', 'f4'], dtype='object', name='file')
        >>> union(
        ...     [
        ...         segmented_index(['f2'], [0], [1]),
        ...         segmented_index(['f1', 'f2'], [0, 1], [1, 2]),
        ...     ]
        ... )
        MultiIndex([('f2', '0 days 00:00:00', '0 days 00:00:01'),
                    ('f1', '0 days 00:00:00', '0 days 00:00:01'),
                    ('f2', '0 days 00:00:01', '0 days 00:00:02')],
        >>> union(
        ...     [
        ...         filewise_index(['f1', 'f2']),
        ...         segmented_index(['f1', 'f2'], [0, 0], [1, 1]),
        ...     ]
        ... )
        MultiIndex([('f1', '0 days',               NaT),
                    ('f2', '0 days',               NaT),
                    ('f1', '0 days', '0 days 00:00:01'),
                    ('f2', '0 days', '0 days 00:00:01')],
```

audformat/core/utils.py
```diff
@@ -1265,6 +1267,60 @@ def union(
...
+
+    """
+    if not objs:
+        return pd.Index([])
+
+    if not is_index_alike(objs):
```
hagenw, line 1318: I would add a short cut before this line:

```suggestion
    if len(objs) == 1:
        return objs[0]
        
    if not is_index_alike(objs):
```

This requires that `objs` are only index objects, but you designed `audformat.utils.union_misc()` anyway to only support those. Was this on purpose? `audformat.utils.union()` also supports dataframes and series.

audformat/core/utils.py
frankenjoe: done

audformat/core/utils.py
frankenjoe: done

frankenjoe:
> You need to convert MultiIndex with single levels to Index before combining them.

If we do that the result we get for your example is:

```
MultiIndex([('spk1',),
            ('spk2',)],
           names=['speaker'])
```

To make your example work I had to convert `MultiIndex` to `Index` instead if we have a single level. Which means that also for two `MultiIndex` objects we get a `Index`:

```python
audformat.utils.union_misc(
    [
        # pd.Index(['spk1'], name='speaker'),
        pd.MultiIndex.from_arrays([['spk1']], names=['speaker']),
        pd.MultiIndex.from_arrays([['spk2']], names=['speaker']),
    ]
)
```
```
Index(['spk1', 'spk2'], dtype='object', name='speaker')
```

Not sure if this is what a user would expect. The other option we have is to return `pd.Index` only if all objects are `pd.Index` or otherwise convert `pd.Index` to `pd.MultiIndex` as you proposed so that the result is a `pd.MultIndex`.

What do you think?

hagenw:
Good point.

I would not convert `pd.Index`, but add a check if all `objs` are a MultiIndex and then do not the conversion.

frankenjoe:
> I would not convert `pd.Index`, but add a check if all `objs` are a MultiIndex and then do not the conversion.

Done, I also left a note in the docstring:

image

hagenw:
Great, we are just missing a rebase here, then we are ready to merge.

audformat/core/database.py
```diff
@@ -463,9 +466,14 @@ def map_files(
         """
         def job(table):
             if table.is_segmented:
-                table.df.index = table.df.index.map(
+                index = table.df.index.map(
                     lambda x: (func(x[0]), x[1], x[2])
                 )
+                index = utils.set_index_dtypes(
```
frankenjoe, line 472: After rebasing I got an error because `utils.union()` is no more restrict about the dtypes and calling `map()` apparently changed the dtype of the file level from 'string' to 'object'

frankenjoe:
> Great, we are just missing a rebase here, then we are ready to merge.

Done, I applied two changes after rebasing, see comments.

PR #216, opened 2022-07-22
DOC: correct dtype handling in Table/Column.set()

### LINKED ISSUE #138: Setting values to a column that do not match the scheme type
There is currently no check if the values set to a column match the scheme type.

E.g. setting float values to a string column works (most likely because the dtype of a string column is `object`):

```python
import audformat

db = audformat.Database('test')

# setting float values to a string column
db.schemes['string'] = audformat.Scheme(audformat.define.DataType.STRING)
db['table']['string'] = audformat.Column(scheme_id='string')
db['table']['string'].set(['a', 1.0])
db['table']['string'].get()
```
```
file
f1    a
f2    1
Name: string, dtype: object
```

Setting strings to a float column, on the other hand, fails:

```python
db.schemes['float'] = audformat.Scheme(audformat.define.DataType.FLOAT)
db['table']['float'] = audformat.Column(scheme_id='float')
db['table']['float'].set(['a', 1.0])
```
```
ValueError: could not convert string to float: 'a'
```

hagenw:
Closes #138 

This updates the docstrings of `audformat.Table.set()` and `audformat.Column.set()` to mention that values, that do not match the schemes dtype are first tried to convert to the dtype and only raise an error if this is not possible.

**Table**

image

**Column**

image

### DIFF
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -270,7 +270,8 @@ class Column(HeaderBase):
         By default, all labels of the column are replaced,
         use ``index`` to set a subset.
         If columns is assigned to a :class:`Scheme`
-        values have to match its ``dtype``.
+        values will be automatically converted
+        to match its dtype.
 
         Examples are provided with the
         :ref:`table specifications <data-tables:Tables>`.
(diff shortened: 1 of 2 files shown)


hagenw:
As the commit mesasge is only about the docstring, should I move the new test and the code fix to another pull request?

frankenjoe:
> As the commit message is only about the docstring, should I move the new test and the code fix to another pull request?

Yes, otherwise we may forget to mention the fix in the changelog.

hagenw:
OK, I have removed them from here and will open another pull request after this is merged.

PR #218, opened 2022-07-22
Add utils.set_index_dtypes()

### LINKED ISSUE #217: Utility function to change dtype of MultiIndex
There is no straightforward way to change the dtypes of a `MultiIndex`, see https://github.com/audeering/audformat/pull/212#discussion_r927429341. We already have an internal function `audformat.core.common.set_index_dtype()` for that and should add it to the public API. Maybe extend it with the option to specify a single `dtype` that is then applied to all levels as currently the function expects a dictionary.

frankenjoe:
Closes #217 

Moves `common.set_index_type()` to `utils.set_index_types()` and makes it part of public API. Adds option to specify single `dtype` and raises an error if level names are not unique or level name does not exist.

image

Uses function in docstring example of `MiscTable`:

image

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -478,7 +478,7 @@ class Base(HeaderBase):
             if column.scheme_id is not None:
                 dtypes[column_id] = schemes[column.scheme_id].to_pandas_dtype()
             else:
-                dtypes[column_id] = 'str'
+                dtypes[column_id] = object
 
         # replace dtype with converter for dates or timestamps
         dtypes_wo_converters = {}
@@ -509,7 +509,7 @@ class Base(HeaderBase):
                 level: dtype for level, dtype in dtypes.items()
                 if level in converters
             }
-            df.index = utils.set_index_dtype(df.index, converter_dtypes)
+            df.index = utils.set_index_dtypes(df.index, converter_dtypes)
             # fix columns
             for column_id in columns:
                 if column_id in converters:
@@ -689,7 +689,7 @@ class MiscTable(Base):
                 level: 'Int64' for level, dtype in zip(levels, dtypes)
                 if pd.api.types.is_integer_dtype(dtype)
             }
-            index = utils.set_index_dtype(index, int_dtypes)
+            index = utils.set_index_dtypes(index, int_dtypes)
 
             if not all(levels) or len(levels) > len(set(levels)):
                 raise ValueError(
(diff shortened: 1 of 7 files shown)


audformat/core/table.py
```diff
@@ -479,7 +478,7 @@ def _load_csv(self, path: str):
             if column.scheme_id is not None:
                 dtypes[column_id] = schemes[column.scheme_id].to_pandas_dtype()
             else:
-                dtypes[column_id] = 'str'
+                dtypes[column_id] = object
```
frankenjoe, line 481: Not related to this PR, but I think it's safer to use `object` here.

audformat/core/table.py
```diff
@@ -508,9 +507,9 @@ def _load_csv(self, path: str):
             # fix index
             converter_dtypes = {
                 level: dtype for level, dtype in dtypes.items()
-                if level in converters
+                if level in converters and level in levels
```
frankenjoe, line 510: Otherwise may raise error that level does not exist.

audformat/core/utils.py
```diff
@@ -992,6 +975,126 @@ def replace_file_extension(
...
+
+
+def set_index_dtypes(
+        index: pd.Index,
+        dtypes: typing.Union[
+            str,
+            typing.Dict[str, str],
+        ],
+) -> pd.Index:
+    r"""Set the dtypes of an index for the given levels.
```
hagenw, line 1002:
```suggestion
    r"""Set the dtypes of an index for the given level names.
```

tests/test_utils.py
```diff
@@ -1548,6 +1548,124 @@ def test_replace_file_extension(index, extension, pattern, expected_index):
...
+                ],
+                names=['idx1', 'idx2'],
+            ),
+            {
+                'idx1': 'string',
+                'bad': 'string',
+            },
+            None,
+            marks=pytest.mark.xfail(raises=ValueError),
+        ),
```
hagenw, line 1661: To be on the safe side, I would add the following test as well:

```suggestion
        ),
        pytest.param(
            pd.Index([0], name='idx'),
            {'bad': 'string'},
            None,
            marks=pytest.mark.xfail(raises=ValueError),
        ),
```

hagenw (commented):
Great, this will be indeed helpful

audformat/core/table.py
hagenw: For now it doesn't matter, but if `pandas` changes the behavior for `str` at some point, yes.

PR #220, opened 2022-07-22
Add audformat.utils.intersect_misc()

hagenw:
This adds `audformat.utils.intersect_misc()` to calculate the intersection between indices that are alike.

image

image

I also updated the examples of `audformat.utils.intersect()` in the same fashion as we did for `audformat.utils.union()`:

image


frankenjoe:
Closing in favor of #223

Issue #221, opened 2022-07-25
Single concat, intersect and union function


frankenjoe:
In https://github.com/audeering/audformat/pull/214 @hagenw suggested to implement only a single `union()` function that accepts indices of all kind. The same could be applied to `intersect()` and `union()`. The argument against it were dependencies in other packages that may rely on the fact that these functions currently raise an error if the input is not `audformat`. However, I checked and  in fact we are not using those function in our packages. Therefore, I would now also support we stay with a single function. As @hagenw this will keep the API simpler. In addition it may also simplify the implementation of the missing methods of `TableMisc' like `expand_index()`, `update()`, etc. There's a good chance we can implement them in the `Base` class and it will work for both `Table` and `MiscTable` when they internally rely on the same utility function.

PR #222, opened 2022-07-25
Combine union() and union_misc() to single function

frankenjoe:
Applies #221 to implement a single `union()` function.

image

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1317,25 +1317,59 @@ def union(
 ) -> pd.Index:
     r"""Create union of index objects.
 
-    Index objects must be conform to
-    :ref:`table specifications <data-tables:Tables>`.
-    Otherwise use
-    :func:`audformat.utils.union_misc`.
-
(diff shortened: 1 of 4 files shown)


audformat/core/utils.py
```diff
@@ -1317,84 +1317,14 @@ def union(
...
-def union_misc(
-    objs: typing.Sequence[pd.Index],
-) -> pd.Index:
-    r"""Create union of index objects.
-
-    Requires that levels and dtypes
+    If all index objects are conform to
+    :ref:`table specifications <data-tables:Tables>`
+    and at least one object is segmented,
+    the output is a segmented index.
```
hagenw, line 1323: What happens if we have a mixture of filewise, segmented and misc indices?
I guess a misc index that is alike a segmented index, is also a segmented index?

frankenjoe: Exactly, and if is neither a filewise, nor a segmented index an error is raised because indices are not alike.

PR #224, opened 2022-07-25
Extend audformat.utils.concat() to support misc objects

### LINKED ISSUE #221: Single concat, intersect and union function

hagenw:
Closes #221 

This extends `audformat.utils.concat()` to support objects with an arbitrary index.

image

image

image

---

Requires #223 to be merged first.

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -76,6 +76,14 @@ def concat(
         dtype: Int64
         >>> concat(
         ...     [
+        ...         pd.Series([0], index=pd.Index([0]), name='col1'),
+        ...         pd.Series([1], index=pd.Index([0]), name='col2'),
+        ...     ]
+        ... )
+           col1  col2
+        0     0     1
+        >>> concat(
+        ...     [
         ...         pd.Series(
         ...             [0., 1.],
         ...             index=pd.Index(
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -32,12 +32,17 @@ def concat(
 ) -> typing.Union[pd.Series, pd.DataFrame]:
     r"""Concatenate objects.
 
-    Objects must be conform to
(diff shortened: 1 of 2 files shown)


frankenjoe:
Cool, I also started a PR as I wanted to test if `concat()` worked well with #220, but I didn't push yet, so please continue with your PR.

hagenw:
I also wanted to check if it works ;)

Pull request is ready for review now.

hagenw:
I added another small docstring example, but now I will really stop pushing more commits ;)

audformat/core/utils.py
```diff
@@ -32,12 +32,17 @@ def concat(
...
+    If all objects are conform to
+    :ref:`table specifications <data-tables:Tables>`
+    and at least one object is segmented,
+    the output has a segmented index.
+    Otherwise,
+    requires that levels and dtypes
+    of all objects match,
+    see :func:`audformat.utils.is_index_alike`.
 
     The new object contains index and columns of both objects.
```
frankenjoe, line 44:
```suggestion
    The new object contains index and columns of all objects.
```

frankenjoe:
@hagenw I have no more comments, so if you update the description we can merge.

hagenw:
done

PR #225, opened 2022-07-25
Implement [drop,extend,pick]_index for MiscTable

frankenjoe:
Moves `drop_index()`, `extend_index()` and `pick_index()` from `table.Table` to `table.Base` and adapts the code that it works for `Table` and `MiscTable`.

image

image

image

Here are two examples for the error message if input `index` does not match the index of the table:

```python
db['misc'].drop_index(db['segments'].index)
```
```
Cannot drop rows from table if input index and table index are not alike.
Expected index:
	file              string
	start    timedelta64[ns]
	end      timedelta64[ns]
but yours is:
	labels    string
```

```python
db['segments'].extend_index(db['files'].index)
```
```
Cannot extend a segmented table with a filewise index.
```

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -210,7 +210,7 @@ class Base(HeaderBase):
             index: pd.Index,
             *,
             inplace: bool = False,
-    ) -> 'Table':
+    ) -> 'Base':
         r"""Drop rows from index.
 
         Args:
@@ -221,29 +221,13 @@ class Base(HeaderBase):
             new object if ``inplace=False``, otherwise ``self``
 
         Raises:
-            ValueError: if level and dtype of index does not match table
+            ValueError: if index is not conform to table index
 
         """
         if not inplace:
             return self.copy().drop_index(index, inplace=True)
 
-        if hasattr(self, 'type'):
-            input_type = index_type(index)
-            if self.type != input_type:
-                raise ValueError(
-                    f'Cannot extend a '
-                    f'{self.type} '
-                    f'table with a '
-                    f'{input_type} '
-                    f'index.'
-                )
-        else:
-            if not utils.is_index_alike([self.index, index]):
-                # TODO: add utils.assert_index_alike()
-                raise ValueError(
-                    'Levels and dtypes of all objects must match, '
-                    'see audformat.utils.is_index_alike().'
-                )
+        _assert_table_index(self, index, 'drop from')
 
         new_index = self.df.index.difference(index)
         self._df = self.df.reindex(new_index)
@@ -273,7 +257,7 @@ class Base(HeaderBase):
             new object if ``inplace=False``, otherwise ``self``
 
         Raises:
-            ValueError: if level and dtype of index does not match table
+            ValueError: if index is not conform to table index
 
         """
         if not inplace:
@@ -281,23 +265,7 @@ class Base(HeaderBase):
                 index, fill_values=fill_values, inplace=True,
             )
 
-        if hasattr(self, 'type'):
-            input_type = index_type(index)
-            if self.type != input_type:
-                raise ValueError(
-                    f'Cannot extend a '
-                    f'{self.type} '
-                    f'table with a '
-                    f'{input_type} '
-                    f'index.'
-                )
-        else:
-            if not utils.is_index_alike([self.index, index]):
-                # TODO: add utils.assert_index_alike()
-                raise ValueError(
-                    'Levels and dtypes of all objects must match, '
-                    'see audformat.utils.is_index_alike().'
-                )
+        _assert_table_index(self, index, 'extend')
 
         new_index = self.df.index.union(index)
         self._df = self.df.reindex(new_index)
@@ -493,29 +461,13 @@ class Base(HeaderBase):
             new object if ``inplace=False``, otherwise ``self``
 
         Raises:
-            ValueError: if level and dtype of index does not match table
+            ValueError: if index is not conform to table index
 
         """
         if not inplace:
             return self.copy().pick_index(index, inplace=True)
 
-        if hasattr(self, 'type'):
-            input_type = index_type(index)
-            if self.type != input_type:
-                raise ValueError(
-                    'It is not possible to pick a '
-                    f'{input_type} '
-                    'index from a '
-                    f'{self.type} '
-                    f'index'
-                )
-        else:
-            if not utils.is_index_alike([self.index, index]):
-                # TODO: add utils.assert_index_alike()
-                raise ValueError(
-                    'Levels and dtypes of all objects must match, '
-                    'see audformat.utils.is_index_alike().'
-                )
+        _assert_table_index(self, index, 'pick from')
 
         new_index = self.df.index.intersection(index)
         self._df = self.df.reindex(new_index)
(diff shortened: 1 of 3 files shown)


frankenjoe:
@hagenw with slight modification the methods indeed work for both table types now. I put WIP as I still want to work on the error messages.

hagenw:
Cool, maybe we can indeed get most or all of the stuff working after updating `union()`, `intersect()` and `concat()`

frankenjoe:
> Cool, maybe we can indeed get most or all of the stuff working after updating `union()`, `intersect()` and `concat()`

Yes, after this PR there is only `update()` missing and then we are done!

frankenjoe:
@hagenw ready for review

audformat/core/table.py
```diff
@@ -205,6 +205,79 @@ def drop_columns(
 
         return self
 
+    def drop_index(
+            self,
+            index: pd.Index,
+            *,
+            inplace: bool = False,
+    ) -> 'Base':
```
hagenw, line 213: I guess we need something like

```suggestion
    ) -> typing.Union['Base', 'Table', 'MiscTable']:
```

or maybe

```suggestion
    ) -> typing.Union['Table', 'MiscTable']:
```

as `'Base'` is not part of the official API.
I also don't know, if you can combine strings like `'Table'` with `typing.Union`

frankenjoe: Ah good point. I will also change it with the other functions.

hagenw: Or maybe there is even a way to tell it that it returns as type the one of `self`?

hagenw: For `copy()` we have solved in a not so nice why by adding the method also to `Table` and `MiscTable` with the correct typing and docstring.

frankenjoe: They will officially `typing.Self` add it in `3.11`, see https://peps.python.org/pep-0673/. 

With `from __future__ import annotations` I can use it without error, but unfortunately it's then no longer displaying the type in Sphinx, e.g.:

image

is rendered as:

image

So we either ignore that and hope it will show up once it's supported by Sphinx or we use `typing.Union[Table, MiscTable]` as you suggested earlier.

hagenw: We are using the [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) plugin at the moment, but it turns out that newer versions of Sphinx do support the same as well.

With this we could get something like

image

I will do a few more investigations if there are any unwanted consequences, but otherwise we might think about switching to no longer use the extension. But this I would do in another pull request.

hagenw: I also still want to figure out, why for `audformat` the links to the `pandas` type hints (like `pd.Index`) do not work, but for `audinterface` they do.

frankenjoe: Ok, then I would propose we stay with `typing.Self` here.

frankenjoe:
> I also still want to figure out, why for audformat the links to the pandas type hints (like pd.Index) do not work, but for audinterface they do.

Maybe we have to change to `pandas.Index`?

hagenw:
> > I also still want to figure out, why for audformat the links to the pandas type hints (like pd.Index) do not work, but for audinterface they do.
> 
> Maybe we have to change to `pandas.Index`?

We don't do in `audinterface`, e.g.

image

leads to the following result which has clickable links

image

frankenjoe:
I thin the problem is that we define the `intersphinx_mapping` twice and overwrite the one that include `pandas`:

image

audformat/core/table.py
hagenw: I tested further using the build-in typing support of Sphinx. The big problem is that it is not able to add any custom HTML span to the text, but just marks it with `em`, which means we have no way to style it with CSS and cannot adjust it to look like we want it to look. Which means I would not recommend to use it for now.

I would suggest that we stay with `typing.Self` here and not having the return type documented in the docs. Maybe, we can open a pull request in `sphinx-autodoc-typehints` to support it, or open an issue that it would be nice if `sphinx` could add span to the type hints when not used in the function signature with `autodoc_typehints = 'description'` (when used in the function signature it adds already spans to it).

frankenjoe: Ok

tests/test_misc_table.py
```diff
@@ -606,6 +606,73 @@ def test_dtype_multiindex_single_level(
...
 
 
+def test_drop_and_pick_index():
+
+    table_id = 'misc'
+    index = pytest.DB[table_id].index[:5]
+    df_pick = pytest.DB[table_id].pick_index(index).get()
+    index = pytest.DB[table_id].index[5:]
+    df_drop = pytest.DB[table_id].drop_index(index).get()
+    pd.testing.assert_frame_equal(df_pick, df_drop)
```
hagenw, line 616: This would also be true, if both methods would return always an empty array.
Maybe add another check for the length of the dataframe?

tests/test_misc_table.py
```diff
@@ -606,6 +606,73 @@ def test_dtype_multiindex_single_level(
...
+    db['misc'].extend_index(
+        pd.Index(['1', '2'], name='idx'),
+        fill_values='a',
+        inplace=True,
+    )
+    np.testing.assert_equal(
+        db['misc']['columns'].get().values,
+        np.array(['a', 'a']),
+    )
+    index = pd.Index(['1', '3'], name='idx')
```
hagenw, line 662: What happens when we exten by `pd.MultiIndex.from_arrays([['1', '3']], anmes=['idx'])` instead?
It might result to an error as we convert a mixture of `Index` and `MultiIndex` always to `MultiIndex` in our helper functions.

tests/test_misc_table.py
frankenjoe: Good one. I had copied the test from `test_table.py`, but in fact the misc table has only 3 items so in both cases it returned all items. I adjusted the test and also the one in `test_table.py`.

tests/test_misc_table.py
frankenjoe: And another good one. Turns out that it indeed messes up the index:

```python
index = pd.MultiIndex.from_arrays([['1', '3']], names=['idx'])
db['misc'].extend_index(
    index,
    fill_values='b',
    inplace=True,
)
db['misc'].get()
```
```
     columns
1          a
2          a
(1,)       b
(3,)       b
```

To avoid such side-effect I suggest we do not allow to combine `Index` with a single-level `MultIndex` in `union()`

frankenjoe: I created https://github.com/audeering/audformat/issues/227

frankenjoe: I now switched to `utils.union()` inside `extend_index()` and added a test where we extend with a `pd.MultiIndex()`, but we need to solve #227 first.

audformat/core/table.py
hagenw: I created https://github.com/sphinx-doc/sphinx/issues/10711

tests/test_misc_table.py
frankenjoe: Ok, we are almost there. I added not tests where I drop, extend and pick also with a `pd.MultiIndex`. Two tests are failing because in `drop_index()` we are calling `self.df.index.difference(index)`, which will return a `pd.MultiIndex` instead of `pd.Index`. To solve this issue we have to add another utility function `utils.difference()` first. Will do this in a separate PR.

frankenjoe: Ok, I now switched to `utils.symmetric_difference() in `drop_index()` and tests are passing again.

hagenw:
At some point you said you expect that you can influence the order of the returned index by `pick_index()`.
Do you still want to have that feature?
Because at the moment it is not implemented:
```python
import audformat

index = audformat.filewise_index(["f1", "f2"])
table = audformat.Table(index)
new_table = table.pick_index(audformat.filewise_index(["f2", "f1"]))
print(new_table.index)
```
returns
```
```

frankenjoe:
> Because at the moment it is not implemented:

Yes, the order depends on the shortest index that we pass to `intersect()`, so we cannot rely on it.

> At some point you said you expect that you can influence the order of the returned index by pick_index().

I just wanted to put it up for discussion. But I guess if we want to optimize for speed we should not care about order.

hagenw:
> > At some point you said you expect that you can influence the order of the returned index by pick_index().
> 
> I just wanted to put it up for discussion. But I guess if we want to optimize for speed we should not care about order.

I'm fine with `pick_index()` having not a influence on the order, but I would also find it strange if the order would be different in the new index on what we had before.

For example:
```python
table = audformat.Table(index)
new_table = table.drop_index(audformat.filewise_index(['f2']))
```
This should return
```
Index(['f1', 'f3'], dtype='string', name='file')
```
and not
```
Index(['f3', 'f1'], dtype='string', name='file')
```

hagenw:
I created https://github.com/audeering/audformat/issues/248 to track the order issue

PR #229, opened 2022-07-26
WIP: Add utils.symmetric_difference()

frankenjoe:
image

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1,4 +1,5 @@
 import errno
+import itertools
 import os
 import re
 import sys
@@ -174,7 +175,7 @@ def concat(
         if not filewise.all():
             objs = [to_segmented_index(obj) for obj in objs]
     else:
-        _convert_single_level_multi_index(objs)
+        objs = _convert_single_level_multi_index(objs)
 
     # the new index is a union of the individual objects
     index = union([obj.index for obj in objs])
@@ -533,12 +534,12 @@ def intersect(
                    names=['file', 'start', 'end'])
         >>> intersect(
         ...     [
+        ...         filewise_index(['f1']),
         ...         filewise_index(['f1', 'f2']),
-        ...         segmented_index(['f1', 'f2'], [0, 0], [1, 1]),
+        ...         segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1]),
         ...     ]
         ... )
-        MultiIndex([('f1', '0 days', '0 days 00:00:01'),
-                    ('f2', '0 days', '0 days 00:00:01')],
+        MultiIndex([('f1', '0 days', '0 days 00:00:01')],
                    names=['file', 'start', 'end'])
 
     """
@@ -566,7 +567,7 @@ def intersect(
 
     else:
 
-        _convert_single_level_multi_index(objs)
+        objs = _convert_single_level_multi_index(objs)
 
         index = objs[0]
         for obj in objs[1:]:
@@ -1175,6 +1176,130 @@ def set_index_dtypes(
     return index
 
 
+def symmetric_difference(
+        objs: typing.Sequence[typing.Union[pd.Index]],
+) -> pd.Index:
+    r"""Symmetric difference of index objects.
+
+    If all index objects are conform to
+    :ref:`table specifications <data-tables:Tables>`
+    and at least one object is segmented,
+    the output is a segmented index.
+    Otherwise,
+    requires that levels and dtypes
+    of all objects match,
+    see :func:`audformat.utils.is_index_alike`.
+    When the symmetric difference of a
+    :class:`pandas.Index`
+    with a single-level
+    :class:`pandas.MultiIndex`,
+    is calculated,
+    the result is a
+    :class:`pandas.Index`.
+
+    Args:
+        objs: index objects
+
+    Returns:
+        symmetric difference of index objects
+
+    Raises:
+        ValueError: if level and dtypes of objects do not match
+
+    Example:
+        >>> symmetric_difference(
+        ...     [
+        ...         pd.Index([0, 1], name='idx'),
+        ...         pd.Index([1, 2], dtype='Int64', name='idx'),
+        ...     ]
+        ... )
+        Index([0, 2], dtype='Int64', name='idx')
+        >>> symmetric_difference(
+        ...     [
+        ...         pd.Index([0, 1], name='idx'),
+        ...         pd.MultiIndex.from_arrays([[1, 2]], names=['idx']),
+        ...     ]
+        ... )
+        Int64Index([0, 2], dtype='int64', name='idx')
+        >>> symmetric_difference(
+        ...     [
+        ...         pd.MultiIndex.from_arrays(
+        ...             [['a', 'b', 'c'], [0, 1, 2]],
+        ...             names=['idx1', 'idx2'],
+        ...         ),
+        ...         pd.MultiIndex.from_arrays(
+        ...             [['b', 'c'], [1, 3]],
+        ...             names=['idx1', 'idx2'],
+        ...         ),
+        ...     ]
+        ... )
+        MultiIndex([('a', 0),
+                    ('c', 2),
+                    ('c', 3)],
+                   names=['idx1', 'idx2'])
+        >>> symmetric_difference(
+        ...     [
+        ...         filewise_index(['f1', 'f2', 'f3']),
+        ...         filewise_index(['f2', 'f3', 'f4']),
+        ...     ]
+        ... )
+        Index(['f1', 'f4'], dtype='string', name='file')
+        >>> symmetric_difference(
+        ...     [
+        ...         segmented_index(['f1'], [0], [1]),
+        ...         segmented_index(['f1', 'f2'], [0, 1], [1, 2]),
+        ...     ]
+        ... )
+        MultiIndex([('f2', '0 days 00:00:01', '0 days 00:00:02')],
+                   names=['file', 'start', 'end'])
+        >>> symmetric_difference(
+        ...     [
+        ...         filewise_index(['f1']),
+        ...         filewise_index(['f1', 'f2']),
+        ...         segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1]),
+        ...     ]
+        ... )
+        MultiIndex([('f1', '0 days', '0 days 00:00:01'),
+                    ('f2', '0 days',               NaT),
+                    ('f2', '0 days', '0 days 00:00:01'),
+                    ('f3', '0 days', '0 days 00:00:01')],
+                   names=['file', 'start', 'end'])
+
+    """
+    if not objs:
+        return pd.Index([])
+
+    if len(objs) == 1:
+        return objs[0]
+
+    objs_filewise = [obj for obj in objs if is_filewise_index(obj)]
+    objs_segmented = [obj for obj in objs if is_segmented_index(obj)]
+
+    if (
+            len(objs_filewise) > 0
+            and len(objs_segmented) > 0
+            and len(objs_filewise + objs_segmented) == len(objs)
+    ):
+
+        # if we have a mixture of segmented and filewise
+        # intersect separately and combine afterwards
+        index_filewise = symmetric_difference(objs_filewise)
+        index_segmented = symmetric_difference(objs_segmented)
+        index = union([index_filewise, index_segmented])
+        index = index.sortlevel()[0]
+
+    else:
+
+        index_union = union(objs)
+        pairwise_intersect = []
+        for pair in itertools.combinations(objs, 2):
+            pairwise_intersect.append(intersect(pair))
+        index_intersect = union(pairwise_intersect)
+        index = index_union.difference(index_intersect)
+
+    return index
+
+
 def to_filewise_index(
         obj: typing.Union[pd.Index, pd.Series, pd.DataFrame],
         root: str,
@@ -1474,14 +1599,16 @@ def union(
                    names=['file', 'start', 'end'])
         >>> union(
         ...     [
+        ...         filewise_index(['f1']),
         ...         filewise_index(['f1', 'f2']),
-        ...         segmented_index(['f1', 'f2'], [0, 0], [1, 1]),
+        ...         segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1]),
         ...     ]
         ... )
         MultiIndex([('f1', '0 days',               NaT),
                     ('f2', '0 days',               NaT),
                     ('f1', '0 days', '0 days 00:00:01'),
-                    ('f2', '0 days', '0 days 00:00:01')],
+                    ('f2', '0 days', '0 days 00:00:01'),
+                    ('f3', '0 days', '0 days 00:00:01')],
                    names=['file', 'start', 'end'])
 
     """
@@ -1499,7 +1626,7 @@ def union(
         if not filewise.all():
             objs = [to_segmented_index(obj) for obj in objs]
     else:
-        _convert_single_level_multi_index(objs)
+        objs = _convert_single_level_multi_index(objs)
 
     # Combine all MultiIndex entries and drop duplicates afterwards,
     # faster than using index.union(),
@@ -1512,8 +1639,8 @@ def union(
 
 
 def _convert_single_level_multi_index(
-        objs: typing.List[typing.Union[pd.Index, pd.Series, pd.DataFrame]],
-):
+        objs: typing.Sequence[typing.Union[pd.Index, pd.Series, pd.DataFrame]],
+) -> typing.Sequence[typing.Union[pd.Index, pd.Series, pd.DataFrame]]:
     r"""Convert single-level pd.MultiIndex to pd.Index.
 
     If input is a mixture of single-level
@@ -1540,6 +1667,7 @@ def _convert_single_level_multi_index(
                      for index in indices)) == 2
 
     if is_single_level and is_mix:
+        objs = list(objs)
         for idx, obj in enumerate(objs):
             if isinstance(obj, pd.MultiIndex):
                 objs[idx] = obj.get_level_values(0)
@@ -1548,6 +1676,8 @@ def _convert_single_level_multi_index(
                     and isinstance(obj.index, pd.MultiIndex)
             ):
                 objs[idx].index = obj.index.get_level_values(0)
+
+    return objs
 
 
 def _is_same_dtype(d1, d2) -> bool:
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1281,20 +1281,14 @@ def symmetric_difference(
     # index.difference() does not preserve string dtype
     # for MultiIndex
     if is_segmented_index(index):
-        index = set_index_dtypes(index, {define.IndexField.FILE: 'string'})
-        # TODO: when https://github.com/audeering/audformat/issues/230
-        #  is fixed replace use:
-        #         index = set_index_dtypes(index, {
-        #                 'file': 'string',
-        #                 'start': 'timedelta64[ns]',
-        #                 'end': 'timedelta64[ns]',
-        #             },
-        #         )
-        if len(index.levels[2]) == 0:
-            index = audformat.segmented_index(
-                index.get_level_values(define.IndexField.FILE),
-                index.get_level_values(define.IndexField.START),
-            )
+        index = set_index_dtypes(
+            index,
+            {
+                define.IndexField.FILE: 'string',
+                define.IndexField.START: 'timedelta64[ns]',
+                define.IndexField.END: 'timedelta64[ns]',
+            },
+        )
     elif isinstance(index, pd.MultiIndex):
         obj = objs[0]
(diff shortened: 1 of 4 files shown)


hagenw:
Symmetric difference is union without intersection:

image
([Wikipedia](https://en.wikipedia.org/wiki/Symmetric_difference))

I created code to test for this:

```python
def expected_difference(objs):
    union = audformat.utils.union(objs)
    intersection = audformat.utils.intersect(objs)
    difference = []
    for idx1 in range(len(union)):
        for idx2 in range(len(intersection)):
            if union[idx1] != intersection[idx2]:
                difference.append(union[idx1])
    return pd.MultiIndex.from_tuples(difference, names=union.names)
```

With this your last docstring example
```python
expected_difference(
    [
        audformat.filewise_index(['f1']),
        audformat.segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1]),
    ]
)
```
should return
```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01')],
```

audformat/core/utils.py
```diff
@@ -174,7 +175,7 @@ def concat(
         if not filewise.all():
             objs = [to_segmented_index(obj) for obj in objs]
     else:
-        _convert_single_level_multi_index(objs)
+        objs = _convert_single_level_multi_index(objs)
```
hagenw, line 178: Interesting, why did the code worked before without any errors?
Because we modified the objects in place?

frankenjoe: The type hint says we have a `Sequence`, which does not support item assignment, i.e. we have to convert the sequence to a list inside the function. And since we now have a new object we must return it.

frankenjoe:
> should return
> 
> ```
> MultiIndex([('f1', '0 days',               NaT),
>             ('f2', '0 days',               NaT),
>             ('f2', '0 days', '0 days 00:00:01'),
>             ('f3', '0 days', '0 days 00:00:01')],
>            names=['file', 'start', 'end'])
> ```

We have file 'f1' in the first and second index, so I don't think it should be returned. And the segment `(f1, 0, 1)` we have only in the third index, so I think it should be part of the result.

frankenjoe:
Mhh, but when I look at the following results your suggestion indeed makes more sense:

```python
objs = [
        audformat.filewise_index(['f1']),
        audformat.segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1]),
]

index1 = audformat.utils.symmetric_difference(objs)
index2 = audformat.utils.intersect(objs)
index3 = audformat.utils.union([index1, index2])
index4 = audformat.utils.union(objs)

print(index1)
print(index2)
print(index3)
print(index4)
```
```
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01'),
            ('f2', '0 days',               NaT)],
MultiIndex([('f1', '0 days', '0 days 00:00:01')],
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01'),
            ('f2', '0 days',               NaT)],
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01')],
```

frankenjoe:
So maybe our `intersect()` function is not working is buggy?

hagenw:
I think you are right:

```
MultiIndex([('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01'),
            ('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01')],
```

should be the result.

For that also (A Δ B) Δ C = A Δ (B Δ C) holds.

My function was wrong and needs to be modified to:

```python
def expected_difference(objs):
    objs = [audformat.utils.to_segmented_index(obj) for obj in objs]
    union = audformat.utils.union(objs)
    intersection = audformat.utils.intersect(objs)
    if len(intersection) == 0:
        return union
    difference = []
    for idx1 in range(len(union)):
        for idx2 in range(len(intersection)):
            if union[idx1] != intersection[idx2]:
                difference.append(union[idx1])
    difference = list(set(difference))
    return pd.MultiIndex.from_tuples(difference, names=union.names)
```

frankenjoe:
It's still a bit unfortunate that as soon as we mix filewise and segmented index objects `union(objs)` is not necessarily `union([intersect(objs), symmetric_difference(objs)])`.

frankenjoe:
I added tests for the new function. Ready for review now.

hagenw:
> It's still a bit unfortunate that as soon as we mix filewise and segmented index objects `union(objs)` is not necessarily the same as `union([intersect(objs), symmetric_difference(objs)])`.

Yes, but I'm not sure if we want to change this.

I think a proper union would be not

```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01')],
```

but

```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```

and the corresponding symmetric difference

```
MultiIndex([('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01'),
            ('f1', '0 days 00:00:01', NaT),
```

frankenjoe:
> I think a proper union would be not

Oh yes, that explains it. So the problem is actually `union()`. We could add an optional flag that remove segments if the union also includes the whole file. Then `union([intersect(objs), symmetric_difference(objs)])` should hold.

hagenw:
> > I think a proper union would be not
> 
> Oh yes, that explains it. So the problem is actually `union()`. We could add an optional flag to remove segments if the union also includes the whole file. Then `union(objs) == union([intersect(objs), symmetric_difference(objs)])` should hold.

It's not only union, we would in general need to mask filewise times by segments, as I did in my example above.
Then we would have:

**Union**

```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```

**Intersection**

```
MultiIndex([('f1', '0 days', '0 days 00:00:01')],
```

**Symmetric Difference**

```
MultiIndex([('f1', '0 days 00:00:01',      NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```

But as I said, I don't think we want to have this.

frankenjoe:
Right, at the moment we only do this for `intersection()`.

hagenw:
There is still a difference between `audformat.utils.symmetric_difference()` and `expected_difference()`.

```python
index1 = audformat.filewise_index(['f1'])
index2 = audformat.filewise_index(['f1', 'f2'])
index3 = audformat.segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1])
```

We then get
```python
audformat.utils.symmetric_difference([index1, index2, index3])
```
```
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01'),
            ('f2', '0 days',               NaT)],
```python
expected_difference([index1, index2, index3])
```
```
MultiIndex([('f2', '0 days',               NaT),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f1', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01')],
```
which is all fine, but if we do
```python
index23 = audformat.utils.symmetric_difference([index2, index3])
audformat.utils.symmetric_difference([index1, index23])
```
we get
```
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01'),
            ('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT)],
```
whereas the other approach still leads to the desired result
```python
index23 = expected_difference([index2, index3])
index23 = audformat.utils.set_index_dtypes(index23, {'file': 'string'})
expected_difference([index1, index23])
```
```
MultiIndex([('f2', '0 days',               NaT),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f1', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01')],
```

frankenjoe:
Ok, but I wonder if anway makes more sense if we implement `symmetric_difference()` in the same way as `intersect()`, i.e. we remove segments if the whole file is still in the result or if the whole file is removed:

```python
index1 = audformat.filewise_index(['f1'])
index2 = audformat.filewise_index(['f1', 'f2'])
index3 = audformat.segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1])
audformat.symmetric_difference([index1, index2, index3])
```
```
MultiIndex([('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```

tests/test_utils.py
```diff
@@ -1819,6 +1819,242 @@ def test_set_index_dtypes(index, dtypes, expected):
...
+            marks=pytest.mark.xfail(raises=ValueError),
+        ),
+    ]
+)
+def test_symmetric_difference(objs, expected):
+    pd.testing.assert_index_equal(
+        audformat.utils.symmetric_difference(objs),
+        expected,
+    )
+
```
hagenw, line 2056: I'm not completely sure yet if (A Δ B) Δ C = A Δ (B Δ C) should also hold with our special filewise/segmented index handling.
But if yes, you could add a test for it here with:
```suggestion
    if len(objs) > 2: 
        pd.testing.assert_index_equal(
            audformat.utils.symmetric_difference(
                [
                    objs[0],
                    audformat.utils.symmetric_difference(objs[1:]),
                ]
            ),
            audformat.utils.symmetric_difference(
                [
                    audformat.utils.symmetric_difference(objs[:-1]),
                    objs[-1],
                ]
            ),
        )

```

hagenw:
> Ok, but I wonder if anway makes more sense if we implement `symmetric_difference()` in the same way as `intersect()`, i.e. we remove segments if the whole file is in the result or if the whole file is removed:
> 
> ```python
> index1 = audformat.filewise_index(['f1'])
> index2 = audformat.filewise_index(['f1', 'f2'])
> index3 = audformat.segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1])
> audformat.symmetric_difference([index1, index2, index3])
> ```
> 
> ```
> MultiIndex([('f2', '0 days',               NaT),
>             ('f3', '0 days', '0 days 00:00:01')],
>            names=['file', 'start', 'end'])
> ```

What is the desired result we need inside `audformat.MiscTable.update()`?

hagenw:
We use it at the moment in `drop_index()` with `new_index = self.df.index.difference(index)`.
So, there the desired result if you start with
```python
index = audformat.segmented_index(['f1', 'f2', 'f3'], [0, 0, 0], [1, 1, 1])
```
and want to drop
```python
drop_index = audformat.filewise_index(['f1'])
```
would be indeed
```python
audformat.utils.symmetric_difference([index, drop_index])
```

```
MultiIndex([('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01')],
```
and not the result we get at the moment, which would be
```
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f3', '0 days', '0 days 00:00:01'),
            ('f1', '0 days',               NaT)],
```

So, I think your proposed changes sounds meaningful to me.

frankenjoe:
> So, I think your proposed changes sounds meaningful to me.

I changed it accordingly and updated the tests. I added some special case to avoid #230. So we either fix #230 before we merge or we move on and do that later.

tests/test_utils.py
frankenjoe: As you already assumed it fails when a combination of filewise and segmented index objects is passed.

hagenw: Yes, but it still looks like a bug in the implementation.

The first `objs` it fails for is:
```python
index1 = audformat.segmented_index(['f1', 'f2'], [0, 0], [1, 1])
index2 = audformat.segmented_index(['f2', 'f3'], [0, 0], [1, 1])
index3 = audformat.filewise_index(['f1', 'f2'])
objs = [
    index1,
    index2,
    index3,
]
```
The result is:
```python
audformat.utils.symmetric_difference([index1, index2, index3])
```
```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```
You get the same when doing:
```python
audformat.utils.symmetric_difference(
    [
        audformat.utils.symmetric_difference([index1, index2]),
        index3,
    ]
)
```
```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```
but a different result when doing:
```python
audformat.utils.symmetric_difference(
    [
        index1,
        audformat.utils.symmetric_difference([index2, index3]),
    ]
)
```
```
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01'),
            ('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```

---

But the last result is wrong in my opinion.
You have
```python
index1
```
```
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f2', '0 days', '0 days 00:00:01')],
```
and
```python
audformat.utils.symmetric_difference([index2, index3])
```
```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```
if you take the symmetric difference between them it should be
```
MultiIndex([('f1', '0 days',               NaT),
            ('f2', '0 days',               NaT),
            ('f3', '0 days', '0 days 00:00:01')],
```
as we said we remove segments if the whole file is present.

frankenjoe: It's because when you do:

```python
audformat.utils.symmetric_difference(
    [
        index1,
        audformat.utils.symmetric_difference([index2, index3]),
    ]
)
```

you are passing only segmented index objects.

frankenjoe:
> So we either fix #230 before we merge or we move on and do that later.

I applied the fixed version of `set_index_dtypes()`.

tests/test_utils.py
frankenjoe: I agree it's not very nice that we have a different behavior now whether you pass a mixture of filewise and segmented vs. only segmented index object. But then we would need to implement a function that removes all segments that lie within another segment first.

frankenjoe: The same issue we have with `intersect)` btw:

```python
objs = [
    audformat.segmented_index('f1', 0, 1),
]
```
```
MultiIndex([('f1', '0 days', '0 days 00:00:01')],
```

vs.

```python
objs = [
    audformat.segmented_index(['f1', 'f2']),
    audformat.segmented_index('f1', 0, 1),
]
```
```
MultiIndex([('f1', '0 days', 'NaT')],
```

Here we would require a utility function that removes segments that include other segments.

hagenw: I think it would be worse the effort as from a user perspective there is not really a difference between
```python
audformat.filewise_index([f1'])
```
and
```python
audformat.segmented_index(['f1'])
```

frankenjoe: Ok, should we then stop here and try to finish #225 without this PR?

hagenw: Basically we need to convert back to filewise if we have a `NaT`. We have `audformat.utils.to_filewise_index()`, but this is doing something completely different ;)

frankenjoe: On the other hand - this might get quite complicated, e.g.:

```python
objs = [
    audformat.segmented_index('f1', 0, 2),
    audformat.segmented_index('f1', 1, 3),
]
```

This currently returns:

```
MultiIndex([], names=['file', 'start', 'end'])
```

but correct would be:

```
MultiIndex([('f1', '0 days 00:00:01', '0 days 00:00:02')],
```

frankenjoe: But there might be so many special cases and stuff, I wonder if it is really worth the effort...

frankenjoe: My opinion: continue as as is and support only a mixture of filewise + segmented as special case.

frankenjoe: But what indeed seems to be a bug is:

```python
objs = [
    audformat.segmented_index('f1'),
    audformat.segmented_index('f1', 0, 1),
]
```
```
MultiIndex([('f1', '0 days', NaT)],
```

This should return:

```
MultiIndex([], names=['file', 'start', 'end'])
```

frankenjoe: The other way round it works:

```python
objs = [
    audformat.segmented_index('f1', 0, 1),
    audformat.segmented_index('f1'),
]
```
```
MultiIndex([], names=['file', 'start', 'end'])
```

audformat/core/utils.py
```diff
@@ -1175,6 +1177,129 @@ def set_index_dtypes(
...
+
+    index_union = union(objs)
+    pairwise_intersect = []
+    for pair in itertools.combinations(objs, 2):
+        pairwise_intersect.append(intersect(pair))
+    index_intersect = union(pairwise_intersect)
+    index = index_union.difference(index_intersect)
+
+    # index.difference() does not preserve string dtype
+    # for MultiIndex
```
hagenw, line 1282:
```suggestion
    # index.difference() for MultiIndex
    # does not preserve string and timedelta64[ns] dtypes
```

tests/test_utils.py
hagenw: Sounds like we should add tests for switching the order of `objs` for our utility functions.

frankenjoe: I think it's a problem with `pd.NaT` and that it matches anything.

hagenw: > My opinion: continue as as is and support only a mixture of filewise + segmented as special case.

In order to get the first release out with `MiscTable` this sounds reasonable. But I would open an issue, stating that we should fix this before we make the 1.0.0 release.

frankenjoe: Because this works:

```python
objs = [
    audformat.segmented_index('f1', 0, 1),
    audformat.segmented_index('f1', 0, 2),
]
```
```
MultiIndex([], names=['file', 'start', 'end'])
```

and:


```python
objs = [
    audformat.segmented_index('f1', 0, 2),
    audformat.segmented_index('f1', 0, 1),
]
```
```
MultiIndex([], names=['file', 'start', 'end'])
```

frankenjoe: > In order to get the first release out with MiscTable this sounds reasonable. But I would open an issue, stating that we should fix this before we make the 1.0.0 release.

Yes, but I would only fix the order bug, whereas the following seems correct to me:

```python
objs = [
    audformat.segmented_index('f1', 0, 1),
    audformat.segmented_index('f1'),
]
```
```
MultiIndex([], names=['file', 'start', 'end'])
```

Cause let's say you have a segmented index and it includes `'f1', 0, 1` and `'f1', 0, pd.NaT`  and then you drop `'f1', 0, pd.NaT`, I don't think you necessarily expect that also `'f1', 0, 1` is dropped. If you a user really wants this, then she can do it by passing a filewise index to explicitly force that all segments belonging to `'f1'` should be removed.

frankenjoe: The correct example would be:

```python
index = audformat.segmented_index(['f1', 'f1'], [0, 0], [1, pd.NaT])
audformat.utils.intersect(
    [
        index,
        audformat.segmented_index('f1'),
    ]
)
```
```
MultiIndex([('f1', '0 days', 'NaT')],
```

vs.

``` python
index = audformat.segmented_index(['f1', 'f1'], [0, 0], [1, pd.NaT])
audformat.utils.intersect(
    [
        index,
        audformat.filewise_index('f1'),
    ]
)
```
```
MultiIndex([('f1', '0 days', '0 days 00:00:01'),
            ('f1', '0 days',               NaT)],
```

hagenw: Which means you would argue the behavior I described above is not a bug, but should be the expected behavior for `symmetric_difference()`, correct?

hagenw: I would also propose to stay with your current implementation here, the question is only we should open a bug and require (A Δ B) Δ C = A Δ (B Δ C). And if not, if we should mention somewhere in the docstrings that this cannot be assumed for a mixture of filewise and segmented indices.

hagenw: > image

No, we agreed already that we will not handle such overlapping cases. Only the case of `NaT` is considered.
The correct return value is indeed what is returned from the function at the moment.

PR #237, opened 2022-07-28
utils.intersect(): drop special treatment for filewise/segmented mix

### LINKED ISSUE #236: Keep special treatment of filewise and segmented index objects in uils.intersect()?
In https://github.com/audeering/audformat/pull/229 we started a discussion on the current behavior of `utils.interact()` if a mixture of filewise and segmented index objects are passed. Currently one can do the following:

```python
objs = [
    audformat.filewise_index(['f1']),
    audformat.segmented_index(['f1', 'f1'], [0, 1], [1, 2]),
]
```
```
MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:01'),
            ('f1', '0 days 00:00:01', '0 days 00:00:02')],
```

Here the filewise index works as kind of a filter that selects all segments from `'f1'` without adding the whole file.

### Shortcomings

The behavior is hard to predict e.g.:

```python
objs = [
    audformat.segmented_index(['f1', 'f2'], [0, 1], [1, 2]),
]
```
```
MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:01'),
            ('f2', '0 days 00:00:01', '0 days 00:00:02')],
```

but:

```python
objs = [
    audformat.segmented_index(['f1', 'f1'], [0, 1], [1, 2]),
    audformat.segmented_index(['f2', 'f2'], [0, 1], [1, 2]),
]
```
```
MultiIndex([], names=['file', 'start', 'end'])
```

or

```python
objs = [
    audformat.filewise_index(['f1']),
    audformat.filewise_index(['f2']),
    audformat.segmented_index(['f1', 'f2'], [0, 1], [1, 2]),
]
```
```
MultiIndex([], names=['file', 'start', 'end'])
```

When filewise and segment index objects are mixed the following rule no longer applies: `(A ∩ B) ∩ C` = `A ∩ (B ∩ C)`. This is because the intersection of a filewise and a segmented index is a segmented index. And no special rule is applied when combining two segmented index objects, e.g.:

```python
objs = [
    audformat.filewise_index(['f1']),
    audformat.segmented_index(['f1', 'f1'], [0, 1], [1, 2]),
]
```
```
MultiIndex([('f1', '0 days 00:00:00', '0 days 00:00:01'),
            ('f1', '0 days 00:00:01', '0 days 00:00:02')],
```

but:

```python
objs = [
    audformat.segmented_index(['f1']),
    audformat.segmented_index(['f1', 'f1'], [0, 1], [1, 2]),
]
```
```
MultiIndex([], names=['file', 'start', 'end'])
```

### Proposal

Instead i suggest to implement it in the same we do with `union()`, i.e. if a mixture of filewise and segmented index objects is passed, we convert the filewise objects to segmented objects. I.e. for

```python
objs = [
    audformat.filewise_index(['f1']),
    audformat.segmented_index(['f1', 'f1'], [0, 1], [1, 2]),
]
```

we get the same result as for:

```python
objs = [
    audformat.segmented_index(['f1']),
    audformat.segmented_index(['f1', 'f1'], [0, 1], [1, 2]),
]
```

### Compatibility

Of course this would be a breaking change. However, we don't rely on this behavior inside `audformat` yet. In `Table.pick_index()` we do not allow to pass a filewise index when the table is segmented or vice versa. Instead, we offer `Table.pick_files()` to pick all segments that belong specific files, but it's not implemented using `utils.intersect()`. Outside of `audformat` we don't use `utils.intersect()` in any of our packages and it's unlikely users rely on it since - except for a docstring example - there is no documentation on this special behavior.

frankenjoe:
Closes #236 

As discussed in #236 we now apply the same strategy to intersect filewise and segmented index objects as in `utils.union()`. That is we convert filewise to segmented objects with `utils.to_segmented_index()`. Therefore a new helper function `_maybe_convert_filewise_index()` is added. Tests are extended to check that the associative property holds for `utils.intersect()`. 

The following docstring example of `utils.intesrect()` is changed from

image

to

image

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1051,6 +1051,7 @@ def test_intersect(objs, expected):
             audformat.utils.intersect(permuted_objs),
             expected,
         )
+    # Ensure A ∩ (B ∩ C) == (A ∩ B) ∩ C
     if len(objs) > 2:
         pd.testing.assert_index_equal(
             audformat.utils.intersect(
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1051,6 +1051,21 @@ def test_intersect(objs, expected):
             audformat.utils.intersect(permuted_objs),
             expected,
         )
+    if len(objs) > 2:
+        pd.testing.assert_index_equal(
+            audformat.utils.intersect(
+                [
+                    objs[0],
+                    audformat.utils.intersect(objs[1:]),
+                ]
+            ),
+            audformat.utils.intersect(
+                [
+                    audformat.utils.intersect(objs[:-1]),
+                    objs[-1],
+                ]
+            ),
+        )
 
 
 @pytest.mark.parametrize(
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -952,7 +952,15 @@ def test_index_has_overlap(obj, expected):
                 audformat.segmented_index(['f2', 'f3'], [0, 0], [1, 1]),
                 audformat.filewise_index(['f1', 'f2']),
             ],
-            audformat.segmented_index('f2', 0, 1),
+            audformat.segmented_index(),
+        ),
+        (
+            [
+                audformat.segmented_index(['f1', 'f2'], [0, 0], [1, pd.NaT]),
+                audformat.segmented_index(['f2', 'f3'], [0, 0], [pd.NaT, 1]),
+                audformat.filewise_index(['f1', 'f2']),
+            ],
+            audformat.segmented_index('f2', 0, pd.NaT),
         ),
         (
             [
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -1043,6 +1051,21 @@ def test_intersect(objs, expected):
             audformat.utils.intersect(permuted_objs),
             expected,
         )
+    if len(objs) > 2:
```
hagenw, line 1054:
```suggestion
    # Ensure A ∩ (B ∩ C) == (A ∩ B) ∩ C
    if len(objs) > 2:
```

hagenw:
Cool, I agree with you, that the behavior is not now much easier to predict if you have a lot of different input objects.

PR #239, opened 2022-07-28
TST: ensure utils.union() is associative

hagenw:
As with `audformat.utils.intersect()` we should ensure that `audformat.utils.union()` is associative, namely Ensure A ∪ (B ∪ C) == (A ∪ B) ∪ C.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2436,3 +2436,19 @@ def test_union(objs, expected):
         audformat.utils.union(objs),
         expected,
     )
+    # Ensure A ∪ (B ∪ C) == (A ∪ B) ∪ C
+    if len(objs) > 2:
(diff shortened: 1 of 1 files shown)


frankenjoe:
I this ready for review? You haven't assigned yet.

frankenjoe:
In the tests for `intersect()` and `symmetric_difference()` we also test for permutations. Should we add it here, too?

hagenw:
No, because permutations are commutative (which was/is not working for `intersect()`) and here we only add a test for associative.
But if you want to have both in one pull request, we can also change the title of it.

frankenjoe:
We can also do it in a another PR if you prefer.

hagenw:
Yes, I think for now I would prefer to only merge this test.

PR #241, opened 2022-07-28
Add utils.symmetric_difference (alt)

frankenjoe:
Closes #229 

Alternative implementation of `utils.symmetric_difference()` without special treatment when filewise and segmented index objects are mixed.

image

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -483,12 +483,6 @@ def intersect(
     Example:
         >>> intersect(
         ...     [
-        ...         pd.Index([1, 2, 3], name='idx'),
-        ...     ]
-        ... )
-        Int64Index([1, 2, 3], dtype='int64', name='idx')
-        >>> intersect(
-        ...     [
         ...         pd.Index([1, np.nan], dtype='Int64', name='idx'),
         ...         pd.Index([1, 2, 3], name='idx'),
         ...     ]
@@ -1175,6 +1169,12 @@ def symmetric_difference(
     Raises:
         ValueError: if level and dtypes of objects do not match
     Example:
+        >>> symmetric_difference(
+        ...     [
+        ...         pd.Index([1, 2, 3], name='idx'),
+        ...     ]
+        ... )
+        Int64Index([1, 2, 3], dtype='int64', name='idx')
         >>> symmetric_difference(
         ...     [
         ...         pd.Index([0, 1], name='idx'),
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1147,6 +1147,7 @@ def symmetric_difference(
         objs: typing.Sequence[typing.Union[pd.Index]],
 ) -> pd.Index:
     r"""Symmetric difference of index objects.
+
     If all index objects are conform to
     :ref:`table specifications <data-tables:Tables>`
     and at least one object is segmented,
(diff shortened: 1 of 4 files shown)


audformat/core/utils.py
```diff
@@ -1143,6 +1143,106 @@ def set_index_dtypes(
...
+    of all objects match,
+    see :func:`audformat.utils.is_index_alike`.
+    When the symmetric difference of a
+    :class:`pandas.Index`
+    with a single-level
+    :class:`pandas.MultiIndex`,
+    is calculated,
+    the result is a
+    :class:`pandas.Index`.
+    Args:
```
hagenw, line 1165:
```suggestion

    Args:
```

audformat/core/utils.py
```diff
@@ -1143,6 +1143,106 @@ def set_index_dtypes(
...
+        objs: index objects
+    Returns:
```
hagenw, line 1167:
```suggestion

    Returns:
```

audformat/core/utils.py
```diff
@@ -1143,6 +1143,106 @@ def set_index_dtypes(
...
+        symmetric difference of index objects
+    Raises:
```
hagenw, line 1169:
```suggestion

    Raises:
```

audformat/core/utils.py
```diff
@@ -1143,6 +1143,106 @@ def set_index_dtypes(
...
+        ValueError: if level and dtypes of objects do not match
+    Example:
```
hagenw, line 1171:
```suggestion

    Example:
```

frankenjoe:
I added one more docstring example that shows what happens when a single object is passed.

audformat/core/utils.py
frankenjoe: Weird it removed those blanks when I copied the function from the other branch. I fixed this and the other spots in the last commit.

audformat/core/utils.py
```diff
@@ -1143,6 +1143,117 @@ def set_index_dtypes(
...
+        return objs[0]
+
+    objs = _maybe_convert_filewise_index(objs)
+    objs = _maybe_convert_single_level_multi_index(objs)
+
+    index = union(objs)
+    count = np.zeros(len(index))
+    for obj in objs:
+        count += index.isin(obj)
+    index = index[count == 1]
```
hagenw, line 1252: I was again wondering if it would be possible to implement this without relying on `union()` to safe time and memory for very large indices.

The current implementation requires 4.4 s on the same data I used in https://github.com/audeering/audformat/pull/234#discussion_r931901348.

If you change the code to relying on actual set operations
```python
    difference = set()
    for obj in objs:
        difference = difference ^ set(obj)
    difference = tuple(difference)
    if isinstance(obj, pd.MultiIndex):
            difference,
            names=obj.names,
        )
        index = set_index_dtypes(index, dict(obj.dtypes))
    else:
        index = pd.Index(difference, name=obj.name, dtype=obj.dtype)
```
it doesn't read that nice any longer, but it runs in 0.35 s.

Another side effect is that some tests are failing as you now get the results back in a different order.

hagenw: Maybe the the `if`-statement could also be moved to a helper function.

hagenw: I did another check, we can speed up `intersect()` by using `set()` as well to 0.1 s, but our implementation for `union()` is already very fast and will be slower if we switch to `set()` there as well.

frankenjoe: Ok, I suggest we stay with the current solution and merge here so that I can continue working on https://github.com/audeering/audformat/pull/225 and do the speed up using `set` here and in `intersect()` in another PR.

hagenw: Good idea, I created https://github.com/audeering/audformat/issues/245 and will take a look at it after merging this.

PR #242, opened 2022-07-28
Fix: intersect() returns empty index for single object

### LINKED ISSUE #238: What should interesect() return for single index?
If we pass a single index object to `utils.intersect()` we currently return that index:

```python
objs = [audformat.filewise_index(['f1', 'f2'])]
```
```
```

But wouldn't it make more sense to return an empty index?

frankenjoe:
Closes #238 

Adds one more docstring example:

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -12,6 +12,7 @@ import pandas as pd
 import audeer
 import audiofile
 
+import audformat.utils
 from audformat.core import define
 from audformat.core.common import to_audformat_dtype
 from audformat.core.database import Database
@@ -1596,9 +1597,12 @@ def _alike_index(index: pd.Index) -> pd.Index:
     elif is_segmented_index(index):
         return segmented_index()
     elif isinstance(index, pd.MultiIndex):
-        return pd.MultiIndex.from_arrays(
-            [[]] * index.nlevels,
-            names=index.names,
+        return audformat.utils.set_index_dtypes(
+            pd.MultiIndex.from_arrays(
+                [[]] * index.nlevels,
+                names=index.names,
+            ),
+            index.dtypes.to_dict(),
         )
     else:
         return pd.Index(
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -483,6 +483,12 @@ def intersect(
     Example:
         >>> intersect(
         ...     [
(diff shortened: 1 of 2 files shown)


audformat/core/utils.py
```diff
@@ -538,7 +544,17 @@ def intersect(
...
-        return objs[0]
+        if is_filewise_index(objs[0]):
+            return filewise_index()
+        elif is_segmented_index(objs[0]):
+            return segmented_index()
+        elif isinstance(objs[0], pd.MultiIndex):
+            return pd.MultiIndex.from_arrays(
+                [[]] * objs[0].nlevels,
+                names=objs[0].names,
+            )
```
hagenw, line 555: This is an interesting case. If you use `pd.testing.assert_index_equal()` it will not raise an error if two empty MultiIndex objects have different dtypes. As they can get here as we don't set the dtype.
A user is able to see this by looking at the `dtypes` attribute.

So the question is if we should ensure the same dtype?
You are also not doing this for the empty `pd.Index([])` case after `else`.

hagenw: Maybe we can make use of this helper function, which I also proposed in https://github.com/audeering/audformat/pull/241#discussion_r932147625:

```python
def create_alike_index(index: pd.Index, data: typing.Tuple) -> pd.Index:
    if isinstance(index, pd.MultiIndex):
        index = set_index_dtypes(
            pd.MultiIndex.from_tuples(data, names=index.names),
            dict(index.dtypes),
        )
    else:
        index = pd.Index(data, name=index.name, dtype=index.dtype)
    return index
```

hagenw: Then you could replace the whole `if`-statement by a single line:
```python
return alike_index(objs[0], ())
```

frankenjoe: > You are also not doing this for the empty pd.Index([]) case after else.

That was a mistake, will add it and also add that helper function you propose.

frankenjoe: done, I added `_alike_index()` but for the moment without a `data` argument.

hagenw: You are still not setting the dtype for MultiIndex, is this on purpose?

frankenjoe: Ups, no :) Should be fixed now.

PR #243, opened 2022-07-28
WIP: Ensures Table.files returns correct dtype

hagenw:
This fixes the problem that loading a databses might fail as we get a mixture of dtypes for `db.files`.

I don't know yet, why this was not covered by tests before, and I also don't know yet why we get the problem in the first place.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -932,6 +932,8 @@ class Table(Base):
         else:
             index = self.df.index.get_level_values(define.IndexField.FILE)
             index.name = define.IndexField.FILE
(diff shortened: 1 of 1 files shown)


frankenjoe:
> I don't know yet, why this was not covered by tests before, and I also don't know yet why we get the problem in the first place.

I guess we should try to figure out the root cause as it may otherwise happen in other places, too. Maybe we are still not correctly converting `object` to `string` in the index when we load from old pickle files?

hagenw:
Closing in favor of #247

PR #244, opened 2022-07-28
Add helper function _assert_index_alike()

frankenjoe:
Adds helper function `utils._assert_index_alike()` which raises an `ValueError` if `is_index_alike()` fails.

### Examples:

```python
objs = [
    pd.Index([]),
    pd.MultiIndex.from_arrays([[], []]),
]
audformat.utils.union(objs)
```
```
Levels and dtypes of all objects must match. Found different number of levels: [1, 2].
```

```python
objs = [
    pd.MultiIndex.from_arrays([[], []], names=['idx1', 'idx2']),
    pd.MultiIndex.from_arrays([[], []]),
]
audformat.utils.union(objs)
```
```
Levels and dtypes of all objects must match. Found different level names: [('idx1', 'idx2'), (None, None)].
```

```python
objs = [
    pd.Index([], dtype='string'),
    pd.Index([]),
]
audformat.utils.union(objs)
```
```
Levels and dtypes of all objects must match. Found different level dtypes: ['object', 'str'].
```

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1138,6 +1138,13 @@ def test_intersect(objs, expected):
         ),
         (
             [
+                pd.Index([1, 2, 3]),
+                pd.Index([10, 20], name='l'),
+            ],
+            "Found different level names: ['None', 'l']",
+        ),
+        (
+            [
                 pd.Index([1, 2, 3], name='l'),
                 pd.MultiIndex.from_arrays([[10, 20]], names=['l']),
             ],
@@ -1242,6 +1249,14 @@ def test_intersect(objs, expected):
             ],
             "Found different level dtypes: "
             "[('object', 'int'), ('object', 'object')]",
+        ),
+        (
+            [
+                pd.MultiIndex.from_arrays([[], []], names=['l1', 'l2']),
+                pd.MultiIndex.from_arrays([[], []]),
+            ],
+            "Found different level names: "
+            "[('None', 'None'), ('l1', 'l2')]",
         ),
         (
             [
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -1106,41 +1107,48 @@ def test_intersect(objs, expected):
...
             ],
-            True,
+            None,
+        ),
+        (
+            [
+                pd.Index([1, 2, 3]),
+                pd.Index([10, 20], name='l'),
+            ],
+            "Found different level names: ['None', 'l']",
```
hagenw, line 1144: Maybe better
```suggestion
            "Found different level names: [None, 'l']",
```
to avoid a name clash if one name is `'None'`.

BTW, that example does not work at the moment:

```python
index1 = pd.Index([])
index2 = pd.Index([], name="None")
audformat.core.utils._assert_index_alike([index1, index2])
```
returns
```
ValueError: Levels and dtypes of all objects must match.
```
without information on the non-matching names, I guess because you would have `('None', 'None')` which match.

frankenjoe: > to avoid a name clash if one name is `'None'`.

Mhh, the reason I convert to string here is that in order to have deterministic output I sort the result, but therefore I had to convert `None` to a string. However, I guess you a right, we should not convert the result to a string. I'll try to find a solution for it.

frankenjoe: Ok, we no longer convert to string. I extended the tests with your example.

frankenjoe: Locally tests are passing, but here I'm getting:

```
AssertionError: Regex pattern "Found\\ different\\ level\\ names:\\ \\[None,\\ 'None'\\]" does not match "Levels and dtypes of all objects must match. Found different level names: ['None', None].".
```

So it seems that after the change sorting is no longer deterministic....

frankenjoe: I now replaced `set()`with `dict.fromkeys()`.

hagenw: perfect

Issue #245, opened 2022-07-28
Speedup intersect() and symmetric_difference() by using set


hagenw:
See the discussion at https://github.com/audeering/audformat/pull/241#discussion_r932147625.

Both functions could be speed up by using `set()`, e.g.

```python
    index = set()
    for obj in objs:
        index = index.symmetric_difference(set(obj))
    index = tuple(index)

    if isinstance(obj, pd.MultiIndex):
        index = pd.MultiIndex.from_tuples(index, names=obj.names)
        index = set_index_dtypes(index, dict(obj.dtypes))
    else:
        index = pd.Index(index, name=obj.name, dtype=obj.dtype)
```

frankenjoe:
So the main concern would be that we don't have control over the order, correct?

hagenw:
I'm not yet completely sure if the order is deterministic.

But you can add `sort()` to the test cases.

frankenjoe:
I'm more concerned about the consequences this has if we rely e.g. on `intersect()` in `Table.pick_index()`. E.g if the index of the table is `f1, f2, f3`, then you pick `f1, f2` the index of the result might be `f2, f1`.

hagenw:
Yes, we should add tests to make sure that this doe snot happen.
If we would need to sort the index to avoid this, then I think the speedup we would get is not worth the effort.

PR #246, opened 2022-07-28
Use list comprehension in audformat.utils.intersect()

hagenw:
Addresses half of https://github.com/audeering/audformat/issues/245

This implements `audformat.utils.intersect()` by converting indices to lists before checking for intersections,
and creating a new index at the end of the operation.
It also suggest to not sort the objects, in order to allow the test proposed in https://github.com/audeering/audformat/pull/254 to pass and preserve original index orders (output is no longer commutative then).

I benchmarked the current code using `list()` against the implementation of the master. The code for the benchmark is given below.

**Results for last obj of same size as first object**

| Table type | Index samples | v0.14.3 | master | list+sort | list-sort | pull request |
| --- | --- | --- | --- | --- | --- | --- |
| filewise | (100000, 500000, 100000) | 0.16 s | 0.19 s  | 0.14 s | 0.16 s | 0.16 s |
| filewise | (100000, 500000, 10000) | 0.12 s | 0.12 s  | 0.07 s | 0.14 s | 0.07 s |
| segmented | (100000, 500000, 100000) | 6.04 s | 14.51 s | 2.81 s | 3.17 s | 2.71 s |
| segmented | (100000, 500000, 10000) | 5.14 s | 12.91 s | 0.24 s | 3.04 s | 0.27 s |

image

---

Code for benchmark:

```python
import audformat
import numpy as np
import pandas as pd
import random
import time


random.seed(0)
samples = 100000


def main():
    nsamples = 100000
    repetitions = 10

    for description, samples in [
            (
                'same length',
                [nsamples, 5 * nsamples, nsamples],
            ),
            (
                '1/10 length',
                [nsamples, 5 * nsamples, int(nsamples / 10)],
            ),
    ]:  
        for index in [filewise, segmented]:
            times = []
            for _ in range(repetitions):
                objs = index(samples)
                times.append(measure(objs))
            print(
                f'Execution time {index.__name__} {description}: '
                f'{np.mean(times):.2f}s'
            )


def measure(objs):
    start = time.time()
    _ = audformat.utils.intersect(objs)
    end = time.time()
    return end - start


def filewise(samples):
    index1 = audformat.filewise_index([f"f{n}" for n in range(samples[0])])
    index2 = audformat.filewise_index([f"f{n}" for n in range(samples[1])])
    idx = [n for n in range(len(index2))]
    index3 = index2[random.sample(idx, k=samples[2])]
    return [index1, index2, index3]


def segmented(samples):
    index1 = audformat.segmented_index(
        [f"f{n}" for n in range(samples[0])],
        random.choices([0, 1, 2, 3], k=samples[0]),
        random.choices([4, 5, 6, pd.NaT], k=samples[0]),
    )
    index2 = audformat.segmented_index(
        [f"f{n}" for n in range(samples[1])],
        random.choices([0, 1, 2, 3], k=samples[1]),
        random.choices([4, 5, 6, pd.NaT], k=samples[1]),
    )
    idx = [n for n in range(len(index2))]
    index3 = index2[random.sample(idx, k=samples[2])]
    return [index1, index2, index3]


if __name__ == '__main__':
    main()
```

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -470,6 +470,14 @@ def intersect(
     the result is a
     :class:`pandas.Index`.
 
+    The order of the resulting index
+    depends on the order of ``objs``.
+    If you require :func:`audformat.utils.intersect`
+    to be commutative_,
+    you have to sort its output.
+
+    .. _commutative: https://en.wikipedia.org/wiki/Commutative_property
+
     Args:
         objs: index objects
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -490,7 +490,7 @@ def intersect(
         >>> intersect(
         ...     [
         ...         pd.Index([1, np.nan], dtype='Int64', name='idx'),
-        ...         pd.Index([1, 2, 3], name='idx'),
+        ...         pd.Index([1, 2, 3], dtype='Int64', name='idx'),
         ...     ]
         ... )
         Index([1], dtype='Int64', name='idx')
@@ -553,14 +553,14 @@ def intersect(
     # sort objects by length
     objs = sorted(objs, key=lambda obj: len(obj))
     # start from shortest index
-    index = objs[0]
-    mask = np.ones(len(index), dtype=bool)
+    index = set(objs[0])
     for obj in objs[1:]:
-        mask &= index.isin(obj)
-        if not mask.any():
+        index = index.intersection(set(obj))
+        if len(index) == 0:
             # break early if no more intersection is possible
             break
-    index = index[mask]
+    index = sorted(list((index)))
+    index = _alike_index(obj, index)
 
     return index
 
@@ -1600,23 +1600,18 @@ def union(
     return index
 
 
-def _alike_index(index: pd.Index) -> pd.Index:
-    r"""Return empty index with same levels and dtypes."""
-    if is_filewise_index(index):
-        return filewise_index()
-    elif is_segmented_index(index):
-        return segmented_index()
-    elif isinstance(index, pd.MultiIndex):
-        return set_index_dtypes(
-            pd.MultiIndex.from_arrays(
-                [[]] * index.nlevels,
-                names=index.names,
-            ),
+def _alike_index(
+        index: pd.Index,
+        data: typing.Sequence = [],
+) -> pd.Index:
+    if isinstance(index, pd.MultiIndex):
+        return audformat.utils.set_index_dtypes(
+            pd.MultiIndex.from_tuples(data, names=list(index.names)),
             index.dtypes.to_dict(),
         )
     else:
         return pd.Index(
-            [],
+            data,
             dtype=index.dtype,
             name=index.name,
         )
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -572,6 +572,8 @@ def intersect(
 
     # Ensure we have order of first object
     index = objs[0].intersection(index)
+    if isinstance(index, pd.MultiIndex):
+        index = set_index_dtypes(index, objs[0].dtypes.to_dict())
 
     return index
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -553,13 +553,12 @@ def intersect(
     # sort objects by length
     objs = sorted(objs, key=lambda obj: len(obj))
     # start from shortest index
-    index = set(objs[0])
+    index = list(objs[0])
     for obj in objs[1:]:
-        index = index.intersection(set(obj))
+        index = [idx for idx in index if idx in obj]
         if len(index) == 0:
             # break early if no more intersection is possible
             break
-    index = sorted(list((index)))
     index = _alike_index(obj, index)
 
     return index
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -472,6 +472,8 @@ def intersect(
 
     The order of the resulting index
     depends on the order of ``objs``.
+    The dtype of the resulting index
+    is identical to the dtype of the first object.
     If you require :func:`audformat.utils.intersect`
     to be commutative_,
     you have to sort its output.
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1605,7 +1605,7 @@ def _alike_index(
         data: typing.Sequence = [],
 ) -> pd.Index:
     if isinstance(index, pd.MultiIndex):
-        return audformat.utils.set_index_dtypes(
+        return set_index_dtypes(
             pd.MultiIndex.from_tuples(data, names=list(index.names)),
             index.dtypes.to_dict(),
         )
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -558,14 +558,20 @@ def intersect(
     objs = _maybe_convert_single_level_multi_index(objs)
     _assert_index_alike(objs)
 
+    # sort objects by length
+    objs_sorted = sorted(objs, key=lambda obj: len(obj))
     # start from first object
-    index = list(objs[0])
-    for obj in objs[1:]:
+    index = list(objs_sorted[0])
+    for obj in objs_sorted[1:]:
         index = [idx for idx in index if idx in obj]
         if len(index) == 0:
             # break early if no more intersection is possible
             break
+
     index = _alike_index(objs[0], index)
+
+    # Ensure we have order of first object
+    index = objs[0].intersection(index)
 
     return index
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -560,11 +560,7 @@ def intersect(
 
     # start from first object
     index = list(objs[0])
-    nlevels = objs[0].nlevels
     for obj in objs[1:]:
-        if nlevels > 1:
-            # for multi-level indices we are faster using set()
-            obj = set(obj)
         index = [idx for idx in index if idx in obj]
         if len(index) == 0:
             # break early if no more intersection is possible
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -560,12 +560,16 @@ def intersect(
 
     # start from first object
     index = list(objs[0])
+    nlevels = objs[0].nlevels
     for obj in objs[1:]:
+        if nlevels > 1:
+            # for multi-level indices we are faster using set()
+            obj = set(obj)
         index = [idx for idx in index if idx in obj]
         if len(index) == 0:
             # break early if no more intersection is possible
             break
-    index = _alike_index(obj, index)
+    index = _alike_index(objs[0], index)
 
     return index
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -550,9 +550,7 @@ def intersect(
     objs = _maybe_convert_single_level_multi_index(objs)
     _assert_index_alike(objs)
 
-    # sort objects by length
-    objs = sorted(objs, key=lambda obj: len(obj))
-    # start from shortest index
+    # start from first object
     index = list(objs[0])
     for obj in objs[1:]:
         index = [idx for idx in index if idx in obj]
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -562,7 +562,12 @@ def intersect(
 
     # sort objects by length
     objs_sorted = sorted(objs, key=lambda obj: len(obj))
-    # start from first object
+
+    # return if the shortest obj has no entries
+    if len(objs_sorted[0]) == 0:
+        return _alike_index(objs[0])
+
+    # start from shortest object
     index = list(objs_sorted[0])
     for obj in objs_sorted[1:]:
         index = [idx for idx in index if idx in obj]
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -876,6 +876,13 @@ def test_index_has_overlap(obj, expected):
         ),
         (
             [
+                audformat.filewise_index(['f3', 'f2', 'f1']),
+                audformat.filewise_index(['f1', 'f2']),
+            ],
+            audformat.filewise_index(['f2', 'f1']),
+        ),
+        (
+            [
                 audformat.filewise_index(['f1', 'f2']),
                 audformat.filewise_index(['f1', 'f2']),
                 audformat.filewise_index('f3'),
@@ -1083,11 +1090,10 @@ def test_index_has_overlap(obj, expected):
     ]
 )
 def test_intersect(objs, expected):
-    for permuted_objs in itertools.permutations(objs):
-        pd.testing.assert_index_equal(
-            audformat.utils.intersect(permuted_objs),
-            expected,
-        )
+    pd.testing.assert_index_equal(
+        audformat.utils.intersect(objs),
+        expected,
+    )
     # Ensure A ∩ (B ∩ C) == (A ∩ B) ∩ C
     if len(objs) > 2:
         pd.testing.assert_index_equal(
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1044,10 +1044,17 @@ def test_index_has_overlap(obj, expected):
         ),
         (
             [
+                pd.Index([1, 2, 3], name='idx'),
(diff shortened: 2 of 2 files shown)


hagenw:
Tests are failing with the new `flake8` v5.0.0 release from https://github.com/PyCQA/flake8/releases/tag/5.0.0

hagenw:
https://github.com/audeering/audformat/pull/257 needs to be merged first to fix the test errors.

hagenw:
Tests are passing now.

tests/test_utils.py
```diff
@@ -1038,7 +1045,7 @@ def test_index_has_overlap(obj, expected):
         (
             [
                 pd.Index([1, np.nan], dtype='Int64', name='idx'),
-                pd.Index([1, 2, 3], name='idx'),
+                pd.Index([1, 2, 3], dtype='Int64', name='idx'),
```
frankenjoe, line 1048: I this needed now? Cause the test was supposed to show that you can `intersect()` a `int64` with `Int64`.

frankenjoe: So maybe change the dtype of the result to `int64`.

hagenw: You can, but it will then return the strange `Int64Index` and not `Index`. I can also change the return value or I can change the order to
```python
            [
                pd.Index([1, 2, 3], name='idx'),
                pd.Index([1, np.nan], dtype='Int64', name='idx'),
            ]
```
which would also fix the return value.

frankenjoe: Ah actually it's a bit more complicated that that:

```
objs = [
    pd.Index([1, np.nan], dtype='Int64', name='idx'),
    pd.Index([1, 2, 3], name='idx'),
]
```
```
Int64Index([1], dtype='int64', name='idx')
```

but:

```python
objs = [
    pd.Index([1, 2, 3], name='idx'),
    pd.Index([1, np.nan], dtype='Int64', name='idx'),
]
```
```
Index([1], dtype='Int64', name='idx')
```

So the `dtype` of the result depends on the order of the objects.

hagenw: Exactly

audformat/core/utils.py
```diff
@@ -550,17 +558,14 @@ def intersect(
...
-    objs = sorted(objs, key=lambda obj: len(obj))
-    # start from shortest index
-    index = objs[0]
-    mask = np.ones(len(index), dtype=bool)
+    # start from first object
+    index = list(objs[0])
     for obj in objs[1:]:
-        mask &= index.isin(obj)
-        if not mask.any():
+        index = [idx for idx in index if idx in obj]
```
frankenjoe, line 564: Really surprising this if faster than using `.isn()`.

hagenw: Yes, and a lot if you compare 7.81 s with 0.08 s

tests/test_utils.py
hagenw: I updated the test and now have each orders covered.

frankenjoe: So maybe we should test for both orders and use `pd.Index([1, 2, 3], name='idx')` instead of `pd.Index([1, 2, 3], dtype='Int64', name='idx')`.

frankenjoe:
> Table type 	Index entries 	master 	set 	list+sort 	list-sort
> filewise 	1,000,000 	0.82 s 	2.78 s 	0.11 s 	1.14 s
> segmented 	100,000 	7.81 s 	0.24 s 	0.08 s 	1.98 s

So list with sorting is faster than list with sorting? Isn't that strange?

hagenw:
> > Table type 	Index entries 	master 	set 	list+sort 	list-sort
> > filewise 	1,000,000 	0.82 s 	2.78 s 	0.11 s 	1.14 s
> > segmented 	100,000 	7.81 s 	0.24 s 	0.08 s 	1.98 s
> 
> So list with sorting is faster than list without sorting? Isn't that strange?

It's not, because with intersection you can gain speed in the for loop by picking the smallest index as the starting index to compare against. This you can no longer do without sorting (or if we still want to use it we would need a method to reverse the sorting at the end).

frankenjoe:
> It's not, because with intersection you can gain speed in the for loop by picking the smallest index as the starting index to compare against. This you can no longer do without sorting (or if we still want to use it we would need a method to reverse the sorting at the end).

Ah, now I get it. We are sorting only by length of the objects. That makes sense of course.

audformat/core/utils.py
frankenjoe: And it also seems strange to me that in some situation it is faster than using `set`. Did you also test `set` with sorting?

tests/test_utils.py
```diff
@@ -1083,11 +1097,10 @@ def test_index_has_overlap(obj, expected):
     ]
 )
 def test_intersect(objs, expected):
-    for permuted_objs in itertools.permutations(objs):
```
frankenjoe, line 1086: In `test_symmetric_difference()` we still have the permutation tests, but sort the results:

```python
    expected = expected.sortlevel()[0]
    for permuted_objs in itertools.permutations(objs):
        pd.testing.assert_index_equal(
            audformat.utils.symmetric_difference(permuted_objs).sortlevel()[0],
            expected,
        )
```

Maybe add it here as well?

hagenw: Unfortunately, sorting is not enough as we have the problem discussed above that the returned index type depends on the order as well, which means we would get errors like this here:

```
E       [left]:  Int64Index([1], dtype='int64', name='idx')
E       [right]: Index([1], dtype='Int64', name='idx')
```

hagenw: The other problem with sorting the results is that we cannot test for the desired order. At the moment I added the following test:

```python
        (
            [
                audformat.filewise_index(['f3', 'f2', 'f1']),
            ],
            audformat.filewise_index(['f2', 'f1']),
        ),
```
which would be meaningless if we use sort.

audformat/core/utils.py
```diff
@@ -550,17 +558,14 @@ def intersect(
     objs = _maybe_convert_single_level_multi_index(objs)
     _assert_index_alike(objs)
 
-    # sort objects by length
-    objs = sorted(objs, key=lambda obj: len(obj))
```
frankenjoe, line 554: So we are not sorting anymore? But wasn't it faster with sorting in your benchmark? If we sort, we should probably add another early stop criterion in case the first list is empty.

tests/test_utils.py
frankenjoe: In `test_symmetric_difference()` we first test that the result equals the expected result and after that we do the permutation tests. But I see your point with the changing `dtype`. One way to have a least deterministic `dtypes` would be to always use `Int64` and `boolean` in `set_index_dtypes()`.

audformat/core/utils.py
hagenw: Yes, it's faster with sorting, but sorting would fail the following test:

```python
        (
            [
                audformat.filewise_index(['f3', 'f2', 'f1']),
            ],
            audformat.filewise_index(['f2', 'f1']),
        ),
```
which we need to pass in order to support https://github.com/audeering/audformat/pull/254.

If we find a way to restore order at the end we can still use sort at the beginning, but so far I simply removed sorting which is still faster in most cases than the current implementation.

frankenjoe: There is `ordered-set`: https://pypi.org/project/ordered-set/, but we probably don't want to introduce a dependency because of that.

frankenjoe:
If we stick with the supposed solution of using list comprehension, it might still be faster to use `set()` within the comprehension:

```python
index = [idx for idx in index if idx in set(obj)]
```

tests/test_utils.py
hagenw: We could also update `_alike_index()` instead if we want to allow `set_index_dtypes()` to also return `int64`, which might be desirable in some cases.

frankenjoe: Sorry, I was not precise and meant not to change `set_index_dtypes()` but `dtypes` before we pass it to `set_index_dtypes()`.

hagenw:
> If we stick with the proposed solution of using list comprehension, it might still be faster to use `set()` within the comprehension:
> 
> ```python
> obj = set(obj)
> index = [idx for idx in index if idx in obj]
> ```

On my machine I was not able to finish the runtime test for your suggestion. Maybe it needs too much memory?
I will do another check if `timeit` is the problem or using `set()` is the problem.

hagenw:
> > If we stick with the proposed solution of using list comprehension, it might still be faster to use `set()` within the comprehension:
> > ```python
> > obj = set(obj)
> > index = [idx for idx in index if idx in obj]
> > ```
> 
> On my machine I was not able to finish the runtime test for your suggestion. Maybe it needs too much memory? I will do another check if `timeit` is the problem or using `set()` is the problem.

The problem was I used 
```python
index = [idx for idx in index if idx in set(obj)]
```
instead of
```python
obj = set(obj)
index = [idx for idx in index if idx in obj]
```

Will repeat the benchmark with the corrected code.

tests/test_utils.py
frankenjoe: But maybe it's also fine to simply state in the docstring that the results will have the same `dtypes` as the first object.

hagenw:
Using the code mentioned above we get:

filewise: 3.10 s
segmented: 0.27 s

For whatever reason `set()` struggles with our filewise indices. I saw the same result when I directly used `set().intersection()` for the implementation.

hagenw:
> Using the code mentioned above we get:
> 
> filewise: 3.10 s segmented: 0.27 s
> 
> For whatever reason `set()` struggles with our filewise indices. I saw the same result when I directly used `set().intersection()` for the implementation.

---

Also note that we use 10x more data for the filewise test. I guess this is the reason.

frankenjoe:
Yes, that would explain why we are so much slower for `filewise` with `set()`. Just that we now it's indeed the size of the input could you please repeat with `100.000` in both cases?

hagenw:
> Yes, that would explain why we are so much slower for `filewise` with `set()`. Just that we now it's indeed the size of the input could you please repeat with `100.000` entries in both cases?

If we use 100.000 entries in both cases we get the same result:
* filewise: 0.25 s
* segmented: 0.26 s

frankenjoe:
In fact, I wonder why `set()` can have any benefit, since we know that indices in `audformat` cannot have duplicated items by design.

hagenw:
> In fact, I wonder why `set()` can bring any benefit, since we know that indices in `audformat` cannot have duplicated items by design.

You are right, the length of the `set(obj)` is identical to the length of `obj`.
The difference is an artifact of the repeated measurement time as sets seem to be different cached in Python.
When only focusing on the segmented indices we get **with set**:
* first execution time: 5.80 s
* second execution time: 0.22 s

and **without set**:
* first execution time: 3.05 s
* second execution time: 1.99 s

As repeated execution is not really a feature we target here, I guess we are better without using set.
But maybe I should revisit the benchmarks I did and focus on single execution time?

frankenjoe:
> But maybe I should revisit the benchmarks I did and focus on single execution time?

Yes, I think that makes sense.

audformat/core/utils.py
hagenw: I don't know, maybe my implementation is wrong?
I think converting it to a set takes already time.
Here is how I would implement it in the current code:

```python
    # sort objects by length
    objs_sorted = sorted(objs, key=lambda obj: len(obj))
    # start from first object
    index = set(objs_sorted[0])
    for obj in objs_sorted[1:]:
        index = index.intersection(set(obj))
        if len(index) == 0:
            # break early if no more intersection is possible
            break

    index = _alike_index(objs[0], list(index))
```
and here the results compared to the current pull request

| Table type | Index samples | set | pull request |
| --- | --- | --- | --- |
| filewise | (100000, 500000, 100000) | 0.29 s | 0.16 s |
| filewise | (100000, 500000, 10000) | 0.23 s | 0.07 s |
| segmented | (100000, 500000, 100000) | 6.89 s | 2.71 s |
| segmented | (100000, 500000, 10000) | 6.02 s | 0.27 s |

hagenw:
> > But maybe I should revisit the benchmarks I did and focus on single execution time?
> 
> Yes, I think that makes sense.

I updated the benchmark code and benchmark results. I also included now v0.14.3 for comparison.

In addition, I found a way to still use sort, but restore the order at the end.

audformat/core/utils.py
hagenw: I have restored sorting.

At the end, I restore the original order by adding an additional 
```python
index = objs[0].intersection(index)
```
which requires that we convert first the `index` from a list to an index.
Maybe there is another way to directly apply the list to `objs[0]`?

tests/test_utils.py
hagenw: I added a statement to the docstring.

audformat/core/utils.py
hagenw: The disadvantage with using `set()` might be that we have to convert every object to a set, also the very long indices (here the one with 500,000 samples). Whereas with `list()` we just convert the shortest index to a list and then check for all entries in that list if they are part of the very long index, which might be faster.

audformat/core/utils.py
hagenw: I also tried to resort the list before creating the index with `_alike_index()` by
```python
index = [idx for idx in objs[0] if idx in index]
```
but interestingly this was up to 100x slower.

tests/test_utils.py
hagenw: We test now for both orders.

audformat/core/utils.py
frankenjoe: Yes, that sounds reasonable. Results look very good now!

audformat/core/utils.py
```diff
@@ -551,16 +561,21 @@ def intersect(
...
 
     # sort objects by length
-    objs = sorted(objs, key=lambda obj: len(obj))
-    # start from shortest index
-    index = objs[0]
-    mask = np.ones(len(index), dtype=bool)
-    for obj in objs[1:]:
-        mask &= index.isin(obj)
-        if not mask.any():
+    objs_sorted = sorted(objs, key=lambda obj: len(obj))
```
frankenjoe, line 564: We could immediately return with `return _alike_index(objs[0])` if `len(objs_sorted[0]) == 0`.

audformat/core/utils.py
frankenjoe: Great we now return in order of the first object, can't think of a better solution.

audformat/core/utils.py
hagenw: Good idea, I added a short cut.

frankenjoe:
I guess you plan to add a test that `Table.pick_index()` preserves order in https://github.com/audeering/audformat/pull/254, right?

hagenw:
> I guess you plan to add a test that `Table.pick_index()` preserves order in #254, right?

Yes, or would you prefer to include it here already?

audformat/core/utils.py
```diff
@@ -551,16 +561,26 @@ def intersect(
...
             # break early if no more intersection is possible
             break
-    index = index[mask]
+
+    index = _alike_index(objs[0], index)
+
+    # Ensure we have order of first object
+    index = objs[0].intersection(index)
+    if isinstance(index, pd.MultiIndex):
+        index = set_index_dtypes(index, objs[0].dtypes.to_dict())
```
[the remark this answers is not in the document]
frankenjoe: I think it's also needed for `pd.Timedelta`, will check.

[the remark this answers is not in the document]
frankenjoe: Nope, `pd.Timedelta` is preserved.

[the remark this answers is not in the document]
frankenjoe: On the other hand, we only convert levels where the `dtype` differs, so the overhead should be minimal and this way we are on the safe side.

frankenjoe:
> Yes, or would you prefer to include it here already?

No let's do it in #254, maybe after also optimizing `symmetric_differences()` using lists.

audformat/core/utils.py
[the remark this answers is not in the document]
hagenw: OK, then let's stay with the safe solution.

PR #247, opened 2022-07-29
Fix string dtype backward compatibility for PKL

hagenw:
This fixes dtypes of filewise and segmented index entries stored in PKL files in cache with `dtype == 'object'`.

We had indeed forgotten to convert the index columns as well.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -623,20 +623,6 @@ class Base(HeaderBase):
                     dtype = dtypes[column_id]
                     df[column_id] = df[column_id].astype(dtype)
 
-        # Older versions of audformat stored columns
-        # assigned to a string scheme as 'object',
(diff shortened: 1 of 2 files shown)


hagenw:
The tests are failing at the moment as we don't have a test example for this.
Should w create an old PKL file in cache or should we go with `nocover`?

frankenjoe:
> Should w create an old PKL file in cache or should we go with `nocover`?

Can't you write a test similar to `test_load_old_pickle()` in `test_misc_table.py`?

hagenw:
> > Should w create an old PKL file in cache or should we go with `nocover`?
> 
> Can't you write a test similar to `test_load_old_pickle()` that we have in `test_misc_table.py`?

Good point, we basically had already a test. I added a modified version of it to `test_table.py` and now everything is covered.

PR #249, opened 2022-07-29
Implement MiscTable.__add__()

frankenjoe:
Implement `__add()__` for `MiscTable` by moving to `Base`.

image

Extends the docstring example of `MiscTable` by:

image

and renames `index_ex` to `index_new` in docstring example of `Table`:

image

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -61,6 +61,44 @@ class Base(HeaderBase):
         self._db = None
         self._id = None
 
+    def __add__(self, other: typing.Self) -> typing.Self:
+        r"""Create new table by combining two tables.
+
+        The new table contains index and columns of both tables.
+        Missing values will be set to ``NaN``.
+        If at least one table is segmented, the output has a segmented index.
+
+        Columns with the same identifier are combined to a single column.
+        This requires that:
+
+        1. both columns have the same dtype
+        2. in places where the indices overlap the values of both columns
+           match or one column contains ``NaN``
+
+        Media and split information,
+        as well as,
+        references to schemes and raters are discarded.
+        If you intend to keep them,
+        use :meth:`audformat.Table.update`.
+
+        Args:
+            other: the other table
+
+        Raises:
+            ValueError: if columns with the same name have different dtypes
+            ValueError: if values in the same position do not match
+
+        """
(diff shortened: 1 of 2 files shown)


audformat/core/table.py
```diff
@@ -61,6 +61,48 @@ def __init__(
...
+        references to schemes and raters are discarded.
+        If you intend to keep them,
+        use ``update()``.
+
+        Args:
+            other: the other table
+
+        Raises:
+            ValueError: if columns with the same name have different dtypes
+            ValueError: if values in the same position do not match
```
hagenw, line 93:
```suggestion
            ValueError: if values in the same position do not match
            ValueError: if level and dtypes of indices do not match
```

frankenjoe: Yes, I just realized that we will no longer be able to combine all `MiscTable` as we can with `Table`. So I put WIP and will add some tests.

frankenjoe:
@hagenw I added tests and extended the docstring example of `MiscTable` similar to `Table`. Ready for review.

hagenw:
> extended the docstring example of `MiscTable` similar to `Table`

I cannot find this

frankenjoe:
> I cannot find this

I forgot to push, should be there now.

audformat/core/table.py
```diff
@@ -770,6 +813,51 @@ class MiscTable(Base):
...
+        >>> table_ex.get()
+                    match
+        file other
+        f1   f2      True
+             f3     False
+        f2   f3      True
+        f4   f1       NaN
+        >>> table_ex.set(
+        ...     {'match': True},
+        ...     index=index_ex,
```
hagenw, line 836: Here I was wondering first as I expected that `index_ex` would cover all entries in `table_ex` as it has the same name, but this is not the case.
But we have the same example for `Table`, so maybe we just stick with it.

frankenjoe: Yes, it's slightly confusing, maybe rename `index_ex` to `index_2`?

hagenw: Yes, please rename to `index_new` in the example for `MiscTable` and `Table`

frankenjoe: done

PR #251, opened 2022-07-29
DOC: include methods/attributes from HeaderBase

### LINKED ISSUE #186: description and meta not documented for Database, Table, Column, Scheme, Media, Rater, Split
As discussed in https://github.com/audeering/audformat/pull/185 the `description` attribute is defined in `HeaderBase`, which means it is not documented for all classes that inherit from `HeaderBase` at the moment.

hagenw:
Closes #186 

This includes documentation for all methods and attributes from `HeaderBase`, which are:
* `description`
* `meta`
* `dump()`
* `to_dict()`
* `from_dict()`

I thought `from_dict()` and `to_dict()` might also be interesting as this would be another way to create the object.
If we decide that we don't want to support this we can still hide them by one of the following methods:
* make them private, e.g. `_from_dict()`
* remove docstring in definition in `HeaderBase`

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -19,6 +19,7 @@ Column
 
 .. autoclass:: Column
     :members:
+    :inherited-members:
 
 filewise_index
 --------------
@@ -35,6 +36,7 @@ Database
 
 .. autoclass:: Database
     :members:
+    :inherited-members:
     :special-members:
 
 index_type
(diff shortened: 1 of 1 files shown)


hagenw:
@jwagner please take a look at the three different options I mention in the description and see what would be your favorite.

frankenjoe:
> @jwagner please take a look at the three different options I mention in the description and see what would be your favorite.

Yes, let's include them.

PR #252, opened 2022-07-29
Implement update() for MiscTable

frankenjoe:
Implements `update()` for `MiscTable` by moving the method to `Base`

image

In addition changes the description of errors in `drop_index()`, `pick_index()` and `extend_index()` to:

image

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -572,7 +572,7 @@ class Base(HeaderBase):
     ):
         r"""Set labels.
 
-        By default all labels of the table are replaced,
+        By default, all labels of the table are replaced,
         use ``index`` to select a subset.
         If a column is assigned to a :class:`Scheme`
         values will be automatically converted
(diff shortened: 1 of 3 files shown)


frankenjoe:
We still need to fix `Database.update()` to also update `MiscTable` objects. But I'd suggest we do it in a separate PR.

audformat/core/table.py
```diff
@@ -593,6 +593,182 @@ def set(
...
+
+        Raises:
+            RuntimeError: if table is not assign to a database
+            ValueError: if split or media does not match
+            ValueError: if overlapping columns reference different schemes
+                or raters
+            ValueError: if a missing scheme or rater cannot be copied
+                because a different object with the same ID exists
+            ValueError: if values in same position overlap
+            ValueError: if index is not conform to table index
```
hagenw, line 638: With conform you are referring to alike?

So maybe
```suggestion
            ValueError:  if level and dtypes of table indices do not match
```

tests/test_table.py
```diff
@@ -1426,8 +1426,7 @@ def test_type():
 def test_update(table, overwrite, others):
     df = table.get()
     table.update(others, overwrite=overwrite)
-    if isinstance(others, audformat.Table):
-        others = [others]
+    others = audeer.to_list(others)
```
hagenw, line 1429: What does this change do?

frankenjoe: Just a simplification to ensure `others` is a list or what do you mean?

hagenw: I mean the removal of `if isinstance(others, audformat.Table):`

hagenw: Ah, sorry, no I get it.

audformat/core/table.py
frankenjoe: We had this wording in three other places, changed it there, too.

frankenjoe: Mhh, actually in the other places we have an index as input, so I will rephrase there to: `if level and dtypes of index does not match table index`.

hagenw: I copied the wording from the utils function, like `intersect`

frankenjoe: done

Issue #248, opened 2022-07-29
Test for preserving index order in drop_index() and pick_index()


hagenw:
As discussed in https://github.com/audeering/audformat/pull/225#issuecomment-1198972974 we might want to preserve the order of the original index entries when modifying them with `audformat.Tabe.drop_index()` or `audformat.Table.pick_index()`, e.g.

```python
index = audformat.filewise_index(['f1', 'f2', 'f3'])
table = audformat.Table(index)
new_table = table.drop_index(audformat.filewise_index(['f2']))
```
This should return
```
Index(['f1', 'f3'], dtype='string', name='file')
```
and not
```
Index(['f3', 'f1'], dtype='string', name='file')
```

frankenjoe:
Currently the order is preserved, but that may change if we optimize `symmetric_difference()` using `set()`.

hagenw:
The same holds for `intersect()`, for which it is proposed to use `set()` in https://github.com/audeering/audformat/pull/246.
At the moment it returns a sorted order, which would be not order preserving.

So maybe if we don't experience any speed issue it might be the easiest solution to stay with the current implementation.

To close this issue, it would still be a good idea to add some tests that ensure preserved order (and would for example fail in https://github.com/audeering/audformat/pull/246)

PR #254, opened 2022-07-29
TST: [pick|extend|drop]_index() should preserve order

hagenw:
Closes https://github.com/audeering/audformat/issues/248

This adds tests preserving the order of the original table index by `drop_index()`, `extend_index()` and `pick_index()`
for `audformat.Table` and `audformat.MiscTable`.


hagenw:
This is now ready for review.

PR #255, opened 2022-07-29
Update MiscTable objects in Database.update()

frankenjoe:
Update `MiscTable` objects when `Database.update()` is called.

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -759,6 +759,9 @@ class Database(HeaderBase):
 
         # join tables
         for other in others:
+            # update misc tables first
+            # as they might be used in schemes
+            # linked by audformat tables
             for misc_id, misc in other.misc_tables.items():
                 if misc_id in self.misc_tables:
                     self[misc_id].update(misc, overwrite=overwrite)
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -758,6 +758,11 @@ class Database(HeaderBase):
 
         # join tables
         for other in others:
+            for misc_id, misc in other.misc_tables.items():
+                if misc_id in self.misc_tables:
+                    self[misc_id].update(misc, overwrite=overwrite)
+                else:
+                    self[misc_id] = misc.copy()
             for table_id, table in other.tables.items():
                 if table_id in self.tables:
                     self[table_id].update(table, overwrite=overwrite)
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -657,8 +657,9 @@ class Database(HeaderBase):
             ValueError: if database has different license or usage
             ValueError: if different media, rater, scheme or split with
                 same ID is found
-            ValueError: if table data cannot be combined (e.g. values in
-                same position overlap)
+            ValueError: if tables cannot be combined
+                (e.g. values in same position overlap or
+                level and dtypes of table indices do not match)
             RuntimeError: if ``copy_media=True``,
                 but one of the involved databases was not saved
                 (contains files but no root folder)
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -2,6 +2,7 @@ import datetime
 import filecmp
 import os
 import re
+import typing
 
 import audeer
 import audiofile
@@ -10,6 +11,19 @@ import pytest
 
 import audformat
 import audformat.testing
+
+
+def create_misc_table(
+        obj: typing.Union[pd.Series, pd.DataFrame],
+) -> audformat.MiscTable:
+    r"""Helper function to create Table."""
+    table = audformat.MiscTable(obj.index)
+    if isinstance(obj, pd.Series):
+        obj = obj.to_frame()
+    for name in obj:
+        table[name] = audformat.Column()
+        table[name].set(obj[name].values)
+    return table
 
 
 def full_path(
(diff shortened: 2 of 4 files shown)


audformat/core/database.py
```diff
@@ -758,6 +759,11 @@ def join_field(
 
         # join tables
         for other in others:
+            for misc_id, misc in other.misc_tables.items():
```
frankenjoe, line 762: I think it makes sense to first update `MiscTable` objects as they may be used in schemes that are referenced by `Table` objects. Within the `MiscTable` objects we probably have to update `MiscTable` objects used in schemes first, since those schemes may be used by other `MiscTable` objects.... but I'm not 100% sure :)

tests/test_database.py
```diff
@@ -657,7 +671,7 @@ def test_update(tmpdir):
     db = audformat.testing.create_db(minimal=True)
     db.author = 'author'
     db.organization = 'organization'
-    db.meta['meta'] = 'meta'
+    db.meta['key'] = 'value'
```
frankenjoe, line 674: To avoid https://github.com/audeering/audformat/issues/256

tests/test_misc_table.py
```diff
@@ -54,7 +54,6 @@ def create_misc_table(
     for name in obj:
         table[name] = audformat.Column()
         table[name].set(obj[name].values)
-    table._df = table.df.astype(obj.dtypes)
```
frankenjoe, line 57: When I copied this function include this line to `test_database.py` and used it in `test_update()` I got an error since calling `pd.concat()` inside `MiscTable.update()` was changing the `dtype` for `int64` to `Int64`. First I thought I found a bug, but then I realized that a `MiscTable` cannot have a column with `dtype` other than `object` unless the column is assigned to a scheme. So with this line we would create invalid `Table` objects. After removing the line the tests worked, so I decided to also remove this line here and in the function `create_table()` in `test_table.py`. I then changed the `dtype` of all empty `pd.Series' to `'object', it's not really necessary to do that, but I think it makes more sense since it is anyway set to `'object'` when creating the `MiscTable` or `Table`. I had to slightly adjust one test and remove one, will put a comment there.

tests/test_table.py
```diff
@@ -141,19 +140,19 @@ def test_access():
...
                     pd.Series(
                         index=audformat.filewise_index(),
-                        dtype=float,
+                        dtype='object',
                     )
                 ),
                 create_table(
                     pd.Series(
-                        [1.],
+                        ['a'],
```
frankenjoe, line 148: Since we cannot create a `Table` with `dtype='float'` without assigning it to a scheme I changed to `'object'` to make the test pass. The actual dtype is not relevant for the test since we only want to show that an empty `Table` can be combined with a non-empty `Table`. So should be ok.

tests/test_table.py
```diff
@@ -300,14 +299,7 @@ def test_access():
                 pytest.DB['files'],
                 pytest.DB['segments'],
             ],
-            create_table(
-                audformat.utils.concat(
```
frankenjoe, line 304: Since the tables `pytest.DB['files']` and `pytest.DB['segments']` have references to schemes that change the dtype of the columns, but when we use `create_table()` we cannot change to those dtypes, since we cannot use schemes. So I changed to `+` here, though I'm not sure how meaningful the test is, since `+` is exactly what we want to test for here :)

frankenjoe:
This is hopefully the last bit we need to get full support for `MiscTable`. It really payed off to change `utils.concat()` etc. to accept all indices now. What I'm not checking here is the case where a `MiscTable` is used a scheme and then updated in `Database.update()`. If we think we should have such a test I recommend we do in a separate PR.

audformat/core/database.py
hagenw: I guess, would be nice to add a comment here in code that we need to update misc tables first.

hagenw: > probably update MiscTable objects used in schemes first, since those schemes may be used by other MiscTable objects

I guess even this will not be sufficient as inside a misc table used as a scheme you can also define a scheme by a misc table. So I'm afraid we need something like a dependency resolution to figure out which misc table has to be updated first.

hagenw: I created https://github.com/audeering/audformat/issues/258 to cover the dependency stuff, which should be handled in a different pull request.

tests/test_database.py
```diff
@@ -12,6 +13,25 @@
 import audformat.testing
 
 
+def create_misc_table(
```
hagenw, line 16: Should we add `audformat.testing.add_misc_table()` or update `audformat.testing.add_table()`?

I don't know if it really is useful, but the code we use in the tests looks like this at the moment:

```python
audformat.testing.add_table(other1, 'table', ...)
other1['misc'] = create_misc_table(...)
```
which automatically raises the question.

hagenw: On the other hand we also have `create_table()` inside `tests/test_table.py`

frankenjoe: > update audformat.testing.add_table()

Yes, I also thought about updating `testing.add_table()` or adding `testing.add_misc_table()` to support `MiscTable` objects. Should we do it in this PR?

The reason we have additionally have `create_table()` and `create_misc_table()` is to create a `Table` / `MiscTable` without a database object.

hagenw: Yes, I would also propose to not include this in the next release.
I created https://github.com/audeering/audformat/issues/260

audformat/core/database.py
frankenjoe: I added a comment.

PR #259, opened 2022-08-01
Fix and speedup utils.symmetric_difference()

### LINKED ISSUE #245: Speedup intersect() and symmetric_difference() by using set

hagenw:
Closes #245

This fixes the results and tests for `audformat.utils.symmetric_difference()` for more than two objects,
and speeds up the implementation by basing it on `list` and `collections.Counter`.
It has the same speed as using `set`, but has the advantage of preserving the order.

**Benchmark results (code below)**

| Table type | Index samples | v0.14.3 | master | pull request |
| --- | --- | --- | --- | --- | 
| filewise | (100000, 500000, 100000) | - | 0.54 s  | 0.41 s |
| filewise | (100000, 500000, 10000) | - | 0.41 s  | 0.36 s |
| segmented | (100000, 500000, 100000) | - | 22.39 s | 8.55 s |
| segmented | (100000, 500000, 10000) | - | 19.74 s | 8.03 s |

image

---

Code for benchmark:

```python
import audformat
import numpy as np
import pandas as pd
import random
import time


random.seed(0)
samples = 100000


def main():
    nsamples = 100000
    repetitions = 10

    for description, samples in [
            (
                'same length',
                [nsamples, 5 * nsamples, nsamples],
            ),
            (
                '1/10 length',
                [nsamples, 5 * nsamples, int(nsamples / 10)],
            ),
    ]:  
        for index in [filewise, segmented]:
            times = []
            for _ in range(repetitions):
                objs = index(samples)
                times.append(measure(objs))
            print(
                f'Execution time {index.__name__} {description}: '
                f'{np.mean(times):.2f}s'
            )


def measure(objs):
    start = time.time()
    _ = audformat.utils.symmeric_difference(objs)
    end = time.time()
    return end - start


def filewise(samples):
    index1 = audformat.filewise_index([f"f{n}" for n in range(samples[0])])
    index2 = audformat.filewise_index([f"f{n}" for n in range(samples[1])])
    idx = [n for n in range(len(index2))]
    index3 = index2[random.sample(idx, k=samples[2])]
    return [index1, index2, index3]


def segmented(samples):
    index1 = audformat.segmented_index(
        [f"f{n}" for n in range(samples[0])],
        random.choices([0, 1, 2, 3], k=samples[0]),
        random.choices([4, 5, 6, pd.NaT], k=samples[0]),
    )
    index2 = audformat.segmented_index(
        [f"f{n}" for n in range(samples[1])],
        random.choices([0, 1, 2, 3], k=samples[1]),
        random.choices([4, 5, 6, pd.NaT], k=samples[1]),
    )
    idx = [n for n in range(len(index2))]
    index3 = index2[random.sample(idx, k=samples[2])]
    return [index1, index2, index3]


if __name__ == '__main__':
    main()
```

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2411,6 +2411,20 @@ def test_to_segmented_index(obj, allow_nat, files_duration, root, expected):
         ),
         (
             [
+                pd.Index([0, 1], dtype='int64', name='idx'),
+                pd.Index([1, 2], dtype='Int64', name='idx'),
+            ],
+            pd.Index([0, 2], dtype='int64', name='idx'),
+        ),
+        (
+            [
+                pd.Index([1, 2], dtype='Int64', name='idx'),
+                pd.Index([0, 1], dtype='int64', name='idx'),
+            ],
+            pd.Index([2, 0], dtype='Int64', name='idx'),
+        ),
+        (
+            [
                 pd.Index([0, 1], name='idx'),
                 pd.MultiIndex.from_arrays([[1, 2]], names=['idx']),
             ],
@@ -2454,12 +2468,6 @@ def test_symmetric_difference(objs, expected):
         audformat.utils.symmetric_difference(objs),
         expected,
     )
-    expected = expected.sortlevel()[0]
-    for permuted_objs in itertools.permutations(objs):
-        pd.testing.assert_index_equal(
-            audformat.utils.symmetric_difference(permuted_objs).sortlevel()[0],
-            expected,
-        )
     # Ensure (A Δ B) Δ C = A Δ (B Δ C)
     if len(objs) > 2:
         pd.testing.assert_index_equal(
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2251,7 +2251,15 @@ def test_to_segmented_index(obj, allow_nat, files_duration, root, expected):
                 audformat.filewise_index(['f1', 'f2']),
                 audformat.filewise_index(['f2', 'f3']),
             ],
-            audformat.filewise_index('f3'),
+            audformat.filewise_index(['f2', 'f3']),
+        ),
+        (
+            [
+                audformat.filewise_index(['f2', 'f3']),
+                audformat.filewise_index(['f1', 'f2']),
+                audformat.filewise_index(['f1', 'f2']),
+            ],
+            audformat.filewise_index(['f3', 'f2']),
         ),
         (
             [
@@ -2294,7 +2302,7 @@ def test_to_segmented_index(obj, allow_nat, files_duration, root, expected):
                 audformat.segmented_index(['f2', 'f1'], [0, 0], [1, 1]),
                 audformat.segmented_index(['f2', 'f3'], [0, 0], [1, 1]),
             ],
-            audformat.segmented_index('f3', 0, 1),
+            audformat.segmented_index(['f2', 'f3'], [0, 0], [1, 1]),
         ),
         (
             [
@@ -2302,11 +2310,7 @@ def test_to_segmented_index(obj, allow_nat, files_duration, root, expected):
                 audformat.segmented_index(['f2', 'f1'], [0, 0], [1, 1]),
                 audformat.segmented_index(['f2', 'f3'], [1, 1], [2, 2]),
             ],
-            audformat.segmented_index(
-                ['f2', 'f3'],
-                [1, 1],
-                [2, 2],
-            ),
+            audformat.segmented_index(['f2', 'f3'], [1, 1], [2, 2]),
         ),
         (
             [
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -2251,7 +2251,15 @@ def test_to_segmented_index(obj, allow_nat, files_duration, root, expected):
                 audformat.filewise_index(['f1', 'f2']),
                 audformat.filewise_index(['f2', 'f3']),
             ],
-            audformat.filewise_index('f3'),
+            audformat.filewise_index(['f2', 'f3']),
```
hagenw, line 2254: Turns out symmetric difference is slightly more complicated and we had a few tests wrong.

Here we have:
```
A = ['f1', 'f2']
B = ['f1', 'f2']
C = ['f2', 'f3']
Then we get:
(A Δ B) Δ C = [] Δ ['f2', 'f3'] = ['f2', 'f3'] = ['f1', 'f2'] Δ ['f1', 'f3'] = A Δ (B Δ C)

Comment (frankenjoe) [reply tests/test_utils.py:2254]: I would argument that the result is correct since `f3` is the only item that is only in one of the three index objects.

Comment (frankenjoe) [reply tests/test_utils.py:2254]: I would not apply it in sequence but look at all three objects at once.

**Comment by hagenw** [comment]
This is now ready for review

File: tests/test_utils.py (hunk 1 above)
Comment (hagenw) [reply tests/test_utils.py:2254]: OK, but this is not the definition of symmetric difference.

Comment (frankenjoe) [reply tests/test_utils.py:2254]: Maybe we should define it as `union() - interesct()`, then the ouput will be `['f1', 'f3']`.

Comment (hagenw) [reply tests/test_utils.py:2254]: You would do `union() - intersect()` also piecewise and the output would be the same. This was my first approach of implementing it.

Comment (hagenw) [reply tests/test_utils.py:2254]: See https://github.com/audeering/audformat/pull/259/commits/7e099899429d6a26ff72c5812b64e2ecf513ff49

Comment (hagenw) [reply tests/test_utils.py:2254]: If a user wants to have your behavior she can still do:
```python
symmetric_difference([index1, union([index2, index3])])
```

frankenjoe: But then you are again splitting up the objects. For me the function is seems pretty useless if we define it like this.

frankenjoe: I would only make sense if you had `obj.symmetric_difference(other_objects)`.

frankenjoe: Sorry, forget about my last two comments :)

frankenjoe: So, neither with the old nor with the new implementation `symmetric_difference(objs) == union(objs) - intersect(objs)` holds, correct?

frankenjoe: So I think we should ask ourselves what is the use-case for the function. If the got with `['f2', 'f3']` (even if this is correct according to the definition of symmetric difference), I don't see at the moment when this will ever be useful.

hagenw: Yes, symmetric difference for three objects would be

venn-symmetric-difference

What you want to have is

venn-desired

Maybe, this is indeed better when using it in our cases, but I think we need to find a different name, or make sure to describe it correctly in the docstring.

frankenjoe: Funny, there seems to be no name for it: https://math.stackexchange.com/questions/2415928/union-of-differences-between-three-sets-where-nothing-intersects-how-is-it-cal :)

An alternative would be that the function takes two objects instead of a list of objects as input. But then it would not fit so well with `union()` and `intersect()`.

hagenw: I would not change it to take only two objects.

I see the following options for switching to the behavior proposed by you:
a) stay with `symmetric_difference()` and the current implementation + add something like `difference()` that has your proposed behavior
b) stay with `symmetric_difference()` and add a comment to the docstring how it behaves for more than two objects
c) change the name to something like `difference()` and change to the proposed behavior

hagenw: I implemented c) in https://github.com/audeering/audformat/pull/264

frankenjoe:
I propose we close this PR as `difference()` with two objects covers `symmetric_difference()` and with more than two objects `symmetric_difference()` is not really useful.

hagenw:
Agree, closing in favor of https://github.com/audeering/audformat/pull/264

PR #261, opened 2022-08-01
Add testing.add_misc_table()

### LINKED ISSUE #260: Add testing.add_misc_table()
It would be a good idea to add `testing.add_misc_table()` or expand `testing.add_table()` as discussed in https://github.com/audeering/audformat/pull/255#discussion_r934238298

frankenjoe:
Closes #260 

image

In `testing.add_table()` we change to:

image

### DIFF
--- a/audformat/core/testing.py
+++ b/audformat/core/testing.py
@@ -49,13 +49,11 @@ def add_misc_table(
         p_none: float = None,
         split_id: str = None,
         media_id: str = None,
-) -> Table:
-    r"""Adds a table with random values.
+) -> MiscTable:
+    r"""Add a miscellaneous table with random values.
 
     By default, adds one column for every scheme in the database.
     To create a specific set of columns use ``columns``.
-    If a ``media_id`` is passed, the file format will be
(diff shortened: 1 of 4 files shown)


audformat/core/testing.py
```diff
@@ -34,6 +34,52 @@
...
+    To create a specific set of columns use ``columns``.
+
+    Args:
+        db: a database
+        table_id: ID of table that will be created
+        index: index object
+        columns: a list of scheme_ids or a dictionary with column names as
+            keys and tuples of ``(scheme_id, rater_id)`` as values. ``None``
+            values are allowed
+        p_none: probability to draw invalid values
```
hagenw, line 65: This I do not get completely, how can we assign values that do not match the scheme without getting an error?

frankenjoe: With `invalid` we refer to `None` values. I will rephrase.

frankenjoe: done

hagenw: Ah, this makes indeed more sense, thanks.

PR #266, opened 2022-08-02
Update scheme labels if index of MiscTable is changed

### LINKED ISSUE #265: Columns linked to MiscTable not updated when index is changed
The following does not work:

```python

db['labels'] = audformat.MiscTable(pd.Index(['a', 'b'], dtype='string', name='idx'))
db.schemes['labels'] = audformat.Scheme(dtype='str', labels='labels')

db['misc'] = audformat.MiscTable(pd.Index([1, 2], name='idx'))
db['misc']['column'] = audformat.Column(scheme_id='labels')
db['misc']['column'].set(['a', 'b'])

db['labels'].drop_index(pd.Index(['a'], dtype='string', name='idx'), inplace=True)
db['misc'].get()
```
```
    column
idx       
1        a
2        b
```

Expected output:

```
    column
idx       
1      NaN
2        b
```

frankenjoe:
Closes #265 

Affects  `MiscTable.drop_index()`,  `MiscTable.extend_index()`, and `MiscTable.pick_index()` if table is used in a scheme and `inplace=True`.

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -271,15 +271,17 @@ class Base(HeaderBase):
             ValueError: if level and dtypes of index does not match table index
 
         """
-        if not inplace:
-            return self.copy().drop_index(index, inplace=True)
+        table = self if inplace else self.copy()
 
-        _assert_table_index(self, index, 'drop rows from')
+        _assert_table_index(table, index, 'drop rows from')
 
-        new_index = utils.difference([self.index, index])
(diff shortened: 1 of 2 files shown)


hagenw:
I see, that's the downside when using a table as a scheme, we need to find tests for everything that can be affected.
At least, conceptual it is straightforward and hopefully after one or two bug fix releases we found all the missing tests ;)

PR #267, opened 2022-08-03
DOC: extend emodb example with misc table

hagenw:
This uses a misc table for the `speaker` scheme in the emodb example.
This allows us to use schemes for the column entries and it's also nice to have an example for the usage of a misc table as scheme.

image

[...]

image

[...]

image

[...]

image

### DIFF
--- a/docs/emodb-example.rst
+++ b/docs/emodb-example.rst
@@ -181,19 +181,23 @@ to the emotion table.
 
     male = audformat.define.Gender.MALE
     female = audformat.define.Gender.FEMALE
-    language = audformat.utils.map_language('de')
-    speaker_mapping = {
-        3: {'gender': male, 'age': 31, 'language': language},
-        8: {'gender': female, 'age': 34, 'language': language},
-        9: {'gender': female, 'age': 21, 'language': language},
-        10: {'gender': male, 'age': 32, 'language': language},
-        11: {'gender': male, 'age': 26, 'language': language},
-        12: {'gender': male, 'age': 30, 'language': language},
-        13: {'gender': female, 'age': 32, 'language': language},
-        14: {'gender': female, 'age': 35, 'language': language},
-        15: {'gender': male, 'age': 25, 'language': language},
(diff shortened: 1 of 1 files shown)


hagenw:
See also https://github.com/audeering/emodb/pull/5

docs/emodb-example.rst
```diff
@@ -254,24 +258,47 @@ and assign the information to it.
         description='Six basic emotions and neutral.',
     )
     db.schemes['confidence'] = audformat.Scheme(
-        audformat.define.DataType.FLOAT,
+        'float',
```
frankenjoe, line 261: According to the type hints we do not support string. I created https://github.com/audeering/audformat/issues/269 to discuss if we should include it.

frankenjoe:
The following line can be removed:

https://github.com/audeering/audformat/pull/267/files#diff-e7691a7d67d893ad4ebeaab16f491be22d0a8afef31f2e0544890962055ffad0L126

frankenjoe:
Ok, except for the one unused import this looks good. Should we merge this one now or wait until we have successfully published and tested it with `audb`?

hagenw:
> The following line can be removed:
> 
> https://github.com/audeering/audformat/pull/267/files#diff-e7691a7d67d893ad4ebeaab16f491be22d0a8afef31f2e0544890962055ffad0L126

I cannot find the line when clicking on the link

hagenw:
I would merge this now.
If we find an error or would to change something, I would do a bug fix afterwards.

hagenw:
> > The following line can be removed:
> > https://github.com/audeering/audformat/pull/267/files#diff-e7691a7d67d893ad4ebeaab16f491be22d0a8afef31f2e0544890962055ffad0L126
> 
> I cannot find the line when clicking on the link

I found an `import audiofile as af` that was not used and removed it.

frankenjoe:
> I found an `import audiofile as af` that was not used and removed it.

Yes that was the line I was referring to. I had expanded the file before I copied the link, probably that was the reason it was not working.

PR #268, opened 2022-08-03
WIP: DOC: use misc table as scheme in map schemes

hagenw:
I think this change is not needed, but I just add it here as I did it anyway to test things.


hagenw:
Closing as this is not needed.

PR #273, opened 2022-08-03
Always convert to Int64 in difference(), intersect(), union()

### LINKED ISSUE #272: difference() int dtype cannot always be given by first object
At the moment we define the int dtype by the first object given to `audformat.utils.difference()`:

```python
audformat.utils.difference(
    [
        pd.Index([0, np.NaN], dtype='Int64', name='idx'),
        pd.Index([0], dtype='int64', name='idx'),
    ]
)
```
results in
```
Index([<NA>], dtype='Int64', name='idx')
```

But if we reverse the order it is not possible to stay with `int64` as dtype due to `np.NaN`:

```python
audformat.utils.difference(
    [
        pd.Index([0], dtype='int64', name='idx'),
        pd.Index([0, np.NaN], dtype='Int64', name='idx'),
    ]
)
```
results in
```
...
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NAType'
```
as `_alike_index(objs[0], index)` is not able to create the index in this case.

hagenw:
Closes #272 

This adjust the return type for `audformat.utils.difference()`, `audformat.utils.intersect()`, `audformat.utils.union()`
to always return `Int64` for integers.

image

image

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -596,6 +596,8 @@ def intersect(
     requires that levels and dtypes
     of all objects match,
     see :func:`audformat.utils.is_index_alike`.
+    Integer dtypes don't have to match,
+    but the result will always be of dtype ``Int64``.
     When a :class:`pandas.Index`
     is intersected with a single-level
     :class:`pandas.MultiIndex`,
@@ -604,8 +606,6 @@ def intersect(
 
     The order of the resulting index
     depends on the order of ``objs``.
-    The dtype of the resulting index
-    is identical to the dtype of the first object.
     If you require :func:`audformat.utils.intersect`
     to be commutative_,
     you have to sort its output.
@@ -628,11 +628,11 @@ def intersect(
         ...         pd.Index([1, 2, 3], name='idx'),
         ...     ]
         ... )
-        Int64Index([], dtype='int64', name='idx')
+        Index([], dtype='Int64', name='idx')
         >>> intersect(
         ...     [
         ...         pd.Index([1, np.nan], dtype='Int64', name='idx'),
-        ...         pd.Index([1, 2, 3], dtype='Int64', name='idx'),
+        ...         pd.Index([1, 2, 3], name='idx'),
         ...     ]
         ... )
         Index([1], dtype='Int64', name='idx')
@@ -642,7 +642,7 @@ def intersect(
         ...         pd.MultiIndex.from_arrays([[1, 2]], names=['idx']),
         ...     ]
         ... )
-        Int64Index([1], dtype='int64', name='idx')
+        Index([1], dtype='Int64', name='idx')
         >>> intersect(
         ...     [
         ...         pd.MultiIndex.from_arrays(
@@ -686,7 +686,7 @@ def intersect(
         return pd.Index([])
 
     if len(objs) == 1:
-        return _alike_index(objs[0])
+        return _alike_index(_maybe_convert_int_dtype(objs[0]))
 
     objs = _maybe_convert_filewise_index(objs)
     objs = _maybe_convert_single_level_multi_index(objs)
@@ -707,7 +707,7 @@ def intersect(
             # break early if no more intersection is possible
             break
 
-    index = _alike_index(objs[0], index)
+    index = _alike_index(_maybe_convert_int_dtype(objs[0]), index)
 
     # Ensure we have order of first object
     index = objs[0].intersection(index)
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1740,20 +1740,6 @@ def _levels(obj):
         return [obj.name]
 
 
-def _maybe_convert_int_dtype(
-        index: pd.Index,
-) -> pd.Index:
-    r"""Convert integer dtypes to Int64."""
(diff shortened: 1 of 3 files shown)


hagenw:
> Closes #272

audformat/core/utils.py
```diff
@@ -395,7 +393,15 @@ def difference(
     counting = collections.Counter(index)
     index = [idx for idx, count in counting.items() if count == 1]
 
-    index = _alike_index(objs[0], index)
+    # select object with highest dtype for integers
+    alike_obj = objs[0]
+    if 'int64' in _dtypes(alike_obj):
```
frankenjoe, line 398: What if we have two integer levels but the object that is now selected has only `Int64` in one of them, whereas another one has it in the other place?

frankenjoe: I wonder if we should simply always convert integer levels to `Int64` in the result of `union()`, `intersect()` and `difference()`?

hagenw: > What if we have two integer levels but the object that is now selected has only Int64 in one of them, whereas another one has it in the other place?

You mean something like this:
```python
index = pd.MultiIndex.from_arrays([[0], [1]], names=['idx1', 'idx2'])
index1 = audformat.utils.set_index_dtypes(index, {'idx1': 'Int64', 'idx2': 'int64'})
index2 = audformat.utils.set_index_dtypes(index, {'idx1': 'int64', 'idx2': 'Int64'})
audformat.utils.difference([index1, index2]).dtypes
```
There we expect that it returns
```
idx1    Int64
idx2    Int64
dtype: object
```
but what we get is
```python
idx1    int64
idx2    Int64
dtype: object
```

frankenjoe: yes

frankenjoe: We can probably fix this, but code will become quite ugly. So always converting to `Int64` makes sense I think. Also because we anyway convert to `Int64` with `Table` and `MiscTable`.

hagenw: OK, I changed to always converting to `Int64`.

audformat/core/utils.py
```diff
@@ -1730,6 +1728,24 @@ def _is_same_dtype(d1, d2) -> bool:
     return d1.name == d2.name
 
 
+def _maybe_convert_int_dtype(
```
frankenjoe, line 1731: We have more or less the same code in `MiscTable.__init__()` so maybe we can call this function there, too?

hagenw: Not so easy as we need access to `levels` and `dtypes` after the code block.

frankenjoe:
I think in `intersect()` we still have one case where we do not convert:

```python
    # return if the shortest obj has no entries
    if len(objs_sorted[0]) == 0:
        return _alike_index(objs[0])
```

frankenjoe:
And then we also have this part:

```python
    # Ensure we have order of first object
    index = objs[0].intersection(index)
    if isinstance(index, pd.MultiIndex):
        index = set_index_dtypes(index, objs[0].dtypes.to_dict())
```

Couldn't this revert back to `int64`?

audformat/core/utils.py
hagenw: OK, I reuse the helper function now.

frankenjoe:
To not miss a case by accident, maybe it would be simpler to always start the functions with:

```python
objs = [_maybe_convert_int_dtype(obj) for obj in objs]
```

?

hagenw:
> To not miss a case by accident, maybe it would be simpler to always start the functions with:
> 
> ```python
> objs = [_maybe_convert_int_dtype(obj) for obj in objs]
> ```
> 
> ?

I tried this, this will lead to several other failures ;)

frankenjoe:
> > 
> 
> I tried this, this will lead to several other failures ;)

Do we know why?

hagenw:
> ```python
> objs = [_maybe_convert_int_dtype(obj) for obj in objs]
> ```

Right, I added a failing test in https://github.com/audeering/audformat/pull/273/commits/020c30fc59737a58928bd451a9e01bcc800725a4 and fixed it in https://github.com/audeering/audformat/pull/273/commits/0764c29efdeacad01f7f715e09938615859051aa

hagenw:
> And then we also have this part:
> 
> ```python
>     # Ensure we have order of first object
>     index = objs[0].intersection(index)
>     if isinstance(index, pd.MultiIndex):
>         index = set_index_dtypes(index, objs[0].dtypes.to_dict())
> ```
> 
> Couldn't this revert back to `int64`?

Yes, tests were not failing, because some expected the wrong output. I fixed the test in https://github.com/audeering/audformat/pull/273/commits/95f2ce4fde3e73081c4a8512eb4a78f54a18bedf and changed the code in https://github.com/audeering/audformat/pull/273/commits/31f1a3acfd5e3f927f8dc68515b422ff2a2229fb

hagenw:
> To not miss a case by accident, maybe it would be simpler to always start the functions with:
> 
> ```python
> objs = [_maybe_convert_int_dtype(obj) for obj in objs]
> ```
> 
> ?

You were right, this works. I changed the code accordingly now.

What was failing for me before was only changing the first object.

Issue #258, opened 2022-08-01
Make sure misc tables are updated in the correct order in Database.update()


hagenw:
As stated at https://github.com/audeering/audformat/pull/255#discussion_r934199932 and https://github.com/audeering/audformat/pull/255#issuecomment-1199435645 we need to figure out the order in which misc tables are updated as they can define schemes used in other misc tables, and those tables should be loaded first before the scheme is used.

It might even be that we have to forbid circular definitions, or at least check if it can be loaded. E.g. if table A uses a scheme from table B and vice versa.

frankenjoe:
Usually, the fact hat we iterate the `MiscTable` objects in the order they were added ensures that a `MiscTable` used as scheme in another `MiscTable` is updated first, because the scheme has to exist when it is linked by the table. However, there might be one exception and that is if the user updates a scheme from a list / dict of labels to a `MiscTable`. I'm not sure if this is currently possible or if we want to support it. But if we do, then we need some kind of dependency resolving. As far as I see this is also the only possible way to create a circular dependency.

frankenjoe:
> However, there might be one exception and that is if the user updates a scheme from a list / dict of labels to a `MiscTable`.

With the fix provided in #263 it is indeed possible to replace a list with a table in a scheme:

```python
db = audformat.testing.create_db(minimal=True)
db.schemes['labels'] = audformat.Scheme(labels=['a', 'b'])

db['misc'] = audformat.MiscTable(pd.Index([1, 2], name='idx'))
db['misc']['column'] = audformat.Column(scheme_id='labels')
db['misc']['column'].set(['a', 'b'])

db['labels'] = audformat.MiscTable(pd.Index(['b', 'c'], dtype='string', name='idx'))
db.schemes['labels'].replace_labels('labels')

db['misc'].get()
```
```
    column
idx       
1      NaN
2        b
```

If we now iterate the misc tables, the scheme table comes after the table that references the scheme:

```python
for table in db.misc_tables:
  print(table)
```
```
misc
labels
```

frankenjoe:
I'm not sure yet about the consequences, though :)

frankenjoe:
One way to avoid circular definitions would be that we forbid that a `MiscTable` that is used as scheme can have columns that are again linked to another  `MiscTable`.

hagenw:
The same we have to forbid. For the other I'm not completely sure. Will try to find an example where it might be useful.
If I cannot find one, I guess we can simply forbid it.

[...]

Issue #61, opened 2021-04-15
audformat.Table.update() too restrictive


hagenw:
At the moment we require that the scheme of a table that is use to update an existing one has to match, but there are certain scenarios where it makes totally sense that it is only included, but does not match.

E.g. consider the following scenario that matches speaker ID labels:

```python
import audformat.testing


db = audformat.testing.create_db(minimal=True)
db.schemes['s'] = audformat.Scheme(str, labels=['1', '2'])
db['t'] = audformat.Table(audformat.filewise_index(['a', 'b']))
db['t']['s'] = audformat.Column(scheme_id='s')
db['t']['s'].set(['1', '2'])

db_new = audformat.testing.create_db(minimal=True)
db_new.schemes['s'] = audformat.Scheme(str, labels=['1'])
db_new['t'] = audformat.Table(audformat.filewise_index(['c']))
db_new['t']['s'] = audformat.Column(scheme_id='s')
db_new['t']['s'].set(['1'])

db.update(db_new)
```
this fails with
```
...
ValueError: Cannot update database, found different value for 'db.schemes['s']':
dtype: str
labels: ['1', '2']
!=
dtype: str
labels: ['1']
```

hagenw:
There is also no easy workaround as we cannot simply replace the new scheme with the original one:

```python
db_new.schemes['s'] = db.schemes['s']
db.update(db_new)
```
fails with
```
...
ValueError: Found two columns with name 's' buf different dtypes:
CategoricalDtype(categories=['1', '2'], ordered=False) != CategoricalDtype(categories=['1'], ordered=False).
```

frankenjoe:
So what you suggest is that we not raise an error as long as the scheme in `db` contains all categories from scheme `db_new`?

hagenw:
Yes, I don't see a reason to raise an error as long as the new scheme is a subset of the old one.

frankenjoe:
> Yes, I don't see a reason to raise an error as long as the new scheme is a subset of the old one.

Is this the only corner case you see?

hagenw:
If you disagree it would be nice to show what we can do instead ;)

hagenw:
> > Yes, I don't see a reason to raise an error as long as the new scheme is a subset of the old one.
> 
> Is this the only corner case you see?

Don't know yet, I guess we will discover more when using it.

frankenjoe:
> There is also no easy workaround as we cannot simply replace the new scheme with the original one:
> 
> ```python
[...]

PR #275, opened 2022-08-04
Raise error for misc table in join_labels() + join_schemes()

### LINKED ISSUE #277: Raise a ValueError instead of TypeError if dtype of labels do not match in Database.update()
As discussed in https://github.com/audeering/audformat/pull/276/files#r937776206 we raise the wrong error type at the moment.

frankenjoe:
Closes #277 

Raise custom error message for misc table schemes in `utils.join_schemes()`.

image

A custom error message is raised when a user calls `utils.join_labels()` with the ID of a misc table.

Also fixes the issue described in #277


frankenjoe:
I started by adding a failing test.

frankenjoe:
@hagenw ready for review

tests/test_utils.py
```diff
@@ -1790,33 +1790,95 @@ def test_join_labels(labels, expected):
...
+    scheme_misc2 = audformat.Scheme(dtype='int', labels='misc')
+    db2.schemes['scheme_misc_id'] = scheme_misc2
+    audformat.utils.join_schemes([db1, db2], 'scheme_misc_id')
+    expected = audformat.utils.union(
+        [
+            db1['misc'].index,
+            db2['misc'].index,
+        ],
+    )
+    pd.testing.assert_index_equal(db1['misc'].index.sortlevel()[0], expected)
```
frankenjoe, line 1831: It's a bit unfortunate we end up with different index order in the misc tables. We could avoid it if in `MiscTable.extend_index()` we change:

```python
new_index = utils.union([table.index, index])
```

to:

```python
new_index = utils.union([index, table.index])
```

frankenjoe: But on the other hand it means we add new items to the front of the index...

frankenjoe: Another option would be that we reindex inside `join_schemes()`.

hagenw: I don't understand immediately why we have this problem and why changing `extend_index()` in the strange way you propose would fix the issue ;)

But I guess the desired index is what `union()` returns, so we could indeed simply reindex to the order of union.

frankenjoe: > I don't understand immediately why we have this problem and why changing extend_index() in the strange way you propose would fix the issue ;)

E.g. we have [1, 2] in the first misc table and [2, 3] in the second. The union of the two is [1, 2, 3]. Now when we extend the first table we get [1, 2, 3], but for the second we get [2, 3, 1]. This is because the shorter index comes first and then the missing items are appended.

frankenjoe: > But I guess the desired index is what union() returns, so we could indeed simply reindex to the order of union.

done

hagenw: Thanks for the explanation and fixing this.

audformat/core/utils.py
```diff
@@ -936,9 +940,26 @@ def join_schemes(
...
 
     """
-    labels = join_labels([db.schemes[scheme_id].labels for db in dbs])
-    for db in dbs:
-        db.schemes[scheme_id].replace_labels(labels)
+    is_misc_table = np.array(
+        [isinstance(db.schemes[scheme_id].labels, str) for db in dbs],
+    )
+    if is_misc_table.any():
+        if not is_misc_table.all():
```
frankenjoe, line 947: Maybe it makes more sense we check for this case in `join_labels()` and raise the error there?

hagenw: `join_labels()` doe snot support misc tables yet. Would you add support for it there as well?
Then you don't need to update `join_schemes()` at all maybe.

frankenjoe: We cannot support it there because we need access to the database to get the misc table. But we can at least check if a user tries to combine a list of labels with labels from a misc table.

hagenw: OK, but then it will not help here as `join_labels()` should not be called when joining schemes with misc tables.

hagenw: > But we can at least check if a user tries to combine a list of labels with labels from a misc table.

We also state 
```python
labels: typing.Sequence[typing.Union[typing.List, typing.Dict]],
```
in `join_labels()` so a user is not supposed to provide a string (and hence a misc table).
So far we have not raised errors for cases that do not fit the typing.
But maybe it would be helpful to remind the user in this case.

hagenw: But I think I get it now, you will use the first `if`-statement only if all tables are mics tables and otherwise handover to `join_labels()` which should then raise an error if any misc table is left, correct?

audformat/core/utils.py
```diff
@@ -865,6 +866,12 @@ def join_labels(
     if not isinstance(labels, list):
         labels = list(labels)
 
+    if any([isinstance(x, str) for x in labels]):
+        raise ValueError(
+            'Cannot join labels from a misc table '
+            'with other label types.'
```
hagenw, line 872: This might be misleading for a user who just provided a string `'xyz'` instead `['xyz']` and is not dealing with a misc table at all.

audformat/core/utils.py
frankenjoe: Yes, I updated it accordingly.

frankenjoe: > in join_labels() so a user is not supposed to provide a string (and hence a misc table).
So far we have not raised errors for cases that do not fit the typing.
But maybe it would be helpful to remind the user in this case.

Good point, in principle we do not have to consider this case. But now we raise a specific error message if a misc table ID is passed.

audformat/core/utils.py
hagenw: So maybe you should repeat the string here, and mention that we assume a misc table is meant, because a string was provided?

frankenjoe: So how would you rephrase?

hagenw: Maybe along those lines:
```python
raise ValueError(
    f"A string value was provided as scheme: '{x}', "
    "which assumes labels defined in a misc table. "
    "'join_labels()' does not support labels from misc tables."
)
```

frankenjoe: If we also rely on this error message in `join_schemes()` we need to point out that it is possible if all schemes are misc table schemes. So maybe:

```
['misc'] are IDs of misc tables and cannot be joined with 'join_labels()'. Instead use join_schemes(). However, this requires that all schemes that are joined use labels from a misc table.
```

audformat/core/utils.py
```diff
@@ -936,9 +947,17 @@ def join_schemes(
           labels: [a, b]
 
     """
-    labels = join_labels([db.schemes[scheme_id].labels for db in dbs])
-    for db in dbs:
-        db.schemes[scheme_id].replace_labels(labels)
+    if all([isinstance(db.schemes[scheme_id].labels, str) for db in dbs]):
+        tables = [db[db.schemes[scheme_id].labels] for db in dbs]
+        index = union([table.index for table in tables])
```
frankenjoe, line 952: Here we only the index of the tables, but maybe we also want to concatenate the columns?

frankenjoe: On the other hand, this is something that `Database.update()` will take care of. So I guess it's enough to combine the index here.

audformat/core/utils.py
frankenjoe: The other option we have is raise different errors in `join_schemes()` and `join_labels()`.

hagenw: > ```
> ['misc'] are IDs of misc tables
> ```

This is not true in all cases, as the user can provide a random string.

Maybe it will be easier to use two different error messages in `join_labels()` and `join_schemes()`.

If you still prefer a single one, I would change to:

```
String values were provided for labels(s): ['misc'],
which assumes those labels are IDs of misc tables.
Labels defined in misc tables cannot be joined with 'join_labels()'.
Instead use 'join_schemes()'.
However, this requires that all schemes that are joined use labels from a misc table.
```

audformat/core/utils.py
hagenw: Good point, so in the tests we should not only create misc tables with an index, but with columns as well (and most likely columns that use a scheme).

frankenjoe: I added those tests in #276 with `Database.update()`. If we agree that `join_schemes()` will only update the index I don' t think we need to extend the tests here.

hagenw: But it will be strange if we don't join the columns. This means we can end up with two databases that have the same scheme (as stated in the docstring of `join_schemes()`), but when using `.get(map=)` it might fail for one of the databases or return different results.

frankenjoe: But it also means that we have to repeat a lot of checks from `Database.update()` like if all columns have the same dtype etc. I would rather add a note in the docstring that it will not updated the columns and the user should call `Database.update()`.

frankenjoe: Also check for rater stuff, etc.

audformat/core/utils.py
frankenjoe: I decided to raise two different error messages.

`join_labels()`:

```
The following string values were provided: '['misc_id']'. 
This assumes that labels are defined in misc tables
 with according IDs, which is not supported by 'join_labels()'.
```

`join_schemes()`:

```
Cannot join schemes with labels from a misc table 
with schemes that do not receive their labels from a misc table.
```

audformat/core/utils.py
hagenw: The main reason we have `join_schemes()` is that `Database.update()` does not support updating schemes, see https://github.com/audeering/audformat/issues/61.
So, I guess we would then need to add the missing functionality to `Table.update()`.

The other question is why would I ever call `join_schemes()` if it does not correctly update the scheme of the databases?

audformat/core/utils.py
hagenw: Great, looks good now.

audformat/core/utils.py
hagenw: > I would rather add a note in the docstring that it will not updated the columns

Before doing that I would then not support misc tables in `join_schemes()`, extend `Database.update()` and `Table.update()` to support non matching schemes, and state in the error message of `join_schemes()` that the user should use `Database.update()` for misc tables.

My original idea of updating `join_schemes()` was, that we can then reuse it inside `Table.update()` to add support for non matching schemes and updating misc tables, but if you think this can be better or only implemented directly inside `audformat.core.table.Base.update()` then I would vote for not supporting misc tables in `join_schemes()`.

frankenjoe: Currently `Database.update()` fails when the index of the misc tables that are used in a scheme do not match. The can be fixed by calling `join_schemes()`. Then `Database.update()` will take care of also updating the columns. So I think the combination of `join_schemes()` + `Database.update()` might work and #276 should now pass.

I mean, I see your point that it would be nice if `join_schemes()` also directly updates the columns. But it's not easy to achieve, e.g. if the misc table of one database has an additional column that links to a scheme that is not in the other database we have to copy that scheme. In `Database.update()` we do such things, but I'm not too motivated to implement the same kind of logic also in `join_schemes()`. It's also more complicated since in `Database.update()` we only have to update the database we call from, but in `join_schemes()` we have to apply it to all databases that are passed.

If you want open an issue, but I would prefer not to work in this PR and first see if #276 will also work with the current solution. Then we at least offer a way to update misc tables schemes. If we later change `join_schemes()` it should still work.

frankenjoe: > Before doing that I would then not support misc tables in `join_schemes()`, extend `Database.update()` and `Table.update()` to support non matching schemes, and state in the error message of `join_schemes()` that the user should use `Database.update()` for misc tables.

Ok that is of course also a solution. I will try to implement it directly in #276

hagenw: OK, thanks for the detailed explanation.

One alternative would be to implement updating the index inside `Database.update()` and not adding support for misc tables to `join_schemes()`. Then we could also later add this functionality.

But I guess we need `join_schemes()` already in `Database.update()` for the other schemes (non misc tables schemes) and it would make it easier if it supports misc tables as well, correct?

hagenw: I wasn't fast enough, you will try already ;)

frankenjoe:
@hagenw I removed support for misc tables from `join_schemes()` and fixed #275.

audformat/core/utils.py
```diff
@@ -851,47 +851,65 @@ def join_labels(
...
+            or not ``list`` or ``dict``
 
     Example:
         >>> join_labels([{'a': 0, 'b': 1}, {'b': 2, 'c': 2}])
         {'a': 0, 'b': 2, 'c': 2}
 
     """
     if len(labels) == 0:
-        return labels
+        return []
```
hagenw, line 863: This changes the type for an empty dict compared to the behavior we had before. Is this intended?

audformat/core/utils.py
```diff
@@ -919,11 +937,21 @@ def join_schemes(
...
+    instead.
+
     Args:
         dbs: sequence of databases
         scheme_id: scheme ID of a scheme with labels
             that should be joined
 
+    Raises:
+        ValueError: if joining schemes with different dtype
+            or only a subset of the schemes use labels from a misc table
```
hagenw, line 953: This is no longer true as we don't support misc tables, so maybe:

```suggestion
        ValueError: if scheme labels are of different dtype
            or not ``list`` or ``dict``
```

tests/test_utils.py
```diff
@@ -1748,6 +1748,10 @@ def test_iter_by_file(obj, expected):
             [],
             [],
         ),
+        (
+            [[], []],
+            [],
+        ),
```
hagenw, line 1754: Could you add tests for an empty dict and a dict containing empty dicts as well.

tests/test_utils.py
```diff
@@ -1809,40 +1823,72 @@ def test_iter_by_file(obj, expected):
...
+        audformat.utils.join_schemes([db4], 'scheme_id')
+
+    # Fail for schemes with different label type
+    db5 = audformat.Database('db')
+    db5.schemes['scheme_id'] = audformat.Scheme('int', labels={0: 'a'})
+    error_msg = "Elements or keys must have the same dtype"
+    with pytest.raises(ValueError, match=error_msg):
+        audformat.utils.join_schemes([db1, db5], 'scheme_id')
+
+    # Fail if only a subset of schemes use labels from misc tables
```
hagenw, line 1885:
```suggestion
    # Fail if some schemes use labels from misc tables
```

hagenw:
Could you update the description, e.g.

> As shown in https://github.com/audeering/audformat/issues/258#issuecomment-1204850958 we need to add support for misc tables schemes in utils.join_schemes().

seems to be wrong to me.

audformat/core/utils.py
frankenjoe: You cannot pass an empty dict, only a list containing an empty dict, but in that case `labels` is not empty. I thought it's more obvious what happens when we change to `[]` here.

tests/test_utils.py
frankenjoe: According to the type string we do not support passing an (empty) dict, but I extended the tests with the case of a list that contains empty dict(s).

frankenjoe:
> seems to be wrong to me.

You are right, I updated the description accordingly.

audformat/core/utils.py
hagenw: Ah, yes, sorry forgot that you provide always a sequence of lists or dicts.

hagenw:
could you also update the screenshot of the docs in the description as it still shows an old text for the `ValueError`. After that we should be fine to merge here.

hagenw:
I updated the screenshot and will now merge.

PR #276, opened 2022-08-04
Join schemes in Database.update()

### LINKED ISSUE #61: audformat.Table.update() too restrictive

frankenjoe:
Closes #61

`Database.update()` now automatically updates schemes if labels do not match. This also works for misc table schemes and in that case missing columns / values will be added. 

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -755,6 +755,9 @@ class Database(HeaderBase):
                     self_labels = self.schemes[scheme_id].labels
 
                     # join labels from misc tables
+                    # by combining the index of all tables,
+                    # columns values will be updated
+                    # later when the tables are joined
                     if (
                         isinstance(other_labels, str)
                         or isinstance(self_labels, str)
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -756,7 +756,7 @@ class Database(HeaderBase):
 
                     # join labels from misc tables
                     # by combining the index of all tables,
-                    # columns values will be updated
+                    # column values will be updated
                     # later when the tables are joined
                     if (
                         isinstance(other_labels, str)
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -626,8 +626,11 @@ class Database(HeaderBase):
 
         In order to :ref:`update a database <update-a-database>`,
         *license* and *usage* have to match.
+        Labels and values of *schemes*
+        with the same ID are combined.
         *Media*, *raters*, *schemes* and *splits* that are not part of
-        the database yet are added. Other fields will be updated by
+        the database yet are added.
+        Other fields will be updated by
         applying the following rules:
 
         ============= =====================================
@@ -657,6 +660,8 @@ class Database(HeaderBase):
             ValueError: if database has different license or usage
             ValueError: if different media, rater, scheme or split with
                 same ID is found
+            ValueError: if schemes cannot be combined,
+                e.g. labels have different dtype
             ValueError: if tables cannot be combined
                 (e.g. values in same position overlap or
                 level and dtypes of table indices do not match)
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -734,13 +734,69 @@ class Database(HeaderBase):
             assert_equal(other, 'license')
             assert_equal(other, 'usage')
 
-        # can only join databases with relatvie paths
+        # can only join databases with relative paths
         for database in [self] + others:
             if not database.is_portable:
                 raise RuntimeError(
                     f"You can only update with databases that are portable. "
                     f"The database '{database.name}' is not portable."
                 )
+
+        # join schemes with labels
+        for other in others:
+            for scheme_id in other.schemes:
+                if scheme_id in self.schemes:
+                    other_labels = other.schemes[scheme_id].labels
+                    self_labels = self.schemes[scheme_id].labels
+
+                    # join labels from misc tables
+                    if (
+                        isinstance(other_labels, str)
+                        or isinstance(self_labels, str)
+                    ):
+
+                        if (
+                            isinstance(other_labels, str)
+                            != isinstance(self_labels, str)
+                        ):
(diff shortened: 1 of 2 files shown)


frankenjoe:
Requires https://github.com/audeering/audformat/pull/275 to be merge first.

frankenjoe:
As discussed in #275 we decided against updating `utils.join_schemes()`, instead schemes are directly combined by `Database.update()`. Either by calling `utils.join_schemes()` or - in case of misc table schemes - the index of all misc tables is combined, which later allows it to combine the tables.

@hagenw this is ready for review.

tests/test_database.py
```diff
@@ -687,6 +688,38 @@ def test_update(tmpdir):
 
     assert db.update(db) == db
 
+    # databases with schemes that are not compatible
+
+    other_bad = audformat.testing.create_db(minimal=True)
+    other_bad.schemes['labels'] = audformat.Scheme(labels=[1, 2, 3])
+    with pytest.raises(TypeError):
```
frankenjoe, line 695: Currently a `TypeError` is raised if the dtype of the labels do not match. I suggest we change it to a `ValueError` in a separate PR.

tests/test_database.py
```diff
@@ -740,44 +775,29 @@ def test_update(tmpdir):
...
-    with pytest.raises(ValueError):
-        audformat.utils.concat(
-            [db['table'].df, other1['table'].df],
-            overwrite=True,
-        )
-
-    # replace labels to avoid error
+    # update db with [other1, other2]
 
-    db.schemes['labels'].replace_labels(other1.schemes['labels'].labels)
```
frankenjoe, line 753: It's no longer required that we call `replace_labels()` here, since this is now handled by `Database.update()`.

audformat/core/database.py
```diff
@@ -734,14 +739,70 @@ def join_field(
...
+                            [
+                                other_table.index,
+                                self_table.index,
+                            ],
+                        )
+                        other_table.extend_index(index, inplace=True)
+                        self_table.extend_index(index, inplace=True)
+                        # ensure same index order in both tables
+                        other_table._df = other_table.df.reindex(index)
+                        self_table._df = self_table.df.reindex(index)
```
hagenw, line 797: You said it cannot happen when updating a database that the columns of the tables are different, so we don't have to care here.
I guess the reason is that it will raise an error when updating the misc tables.
If this is the case, could you maybe add a comment to the code here that we only need to update the index of the table as the columns are handled somewhere else.

audformat/core/database.py
```diff
@@ -759,15 +820,8 @@ def join_field(
...
-            # as they might be used in schemes
-            # linked by audformat tables
-            for misc_id, misc in other.misc_tables.items():
-                if misc_id in self.misc_tables:
-                    self[misc_id].update(misc, overwrite=overwrite)
-                else:
-                    self[misc_id] = misc.copy()
-            for table_id, table in other.tables.items():
+            for table_id in list(other.misc_tables) + list(other.tables):
+                table = other[table_id]
```
hagenw, line 824: To stay with `.items()` I did the following in `audb`, but I guess it is harder to understand and I should also change to your solution:
```python
for table_id, table in {**db.tables, **db.misc_tables}.items():
```

tests/test_database.py
hagenw: I created https://github.com/audeering/audformat/issues/277

hagenw:
This pull request also closes #61:

```python
import audformat.testing


db.schemes['s'] = audformat.Scheme(str, labels=['1', '2'])
db['t'] = audformat.Table(audformat.filewise_index(['a', 'b']))
db['t']['s'] = audformat.Column(scheme_id='s')
db['t']['s'].set(['1', '2'])

db_new = audformat.testing.create_db(minimal=True)
db_new.schemes['s'] = audformat.Scheme(str, labels=['1'])
db_new['t'] = audformat.Table(audformat.filewise_index(['c']))
db_new['t']['s'] = audformat.Column(scheme_id='s')
db_new['t']['s'].set(['1'])

db.update(db_new)
```
now results in
```
name: unittest
source: internal
usage: unrestricted
languages: [deu, eng]
schemes:
  s:
    dtype: str
    labels: ['1', '2']
tables:
  t:
    type: filewise
    columns:
      s: {scheme_id: s}
```

frankenjoe:
> This pull request also closes #61:

Yes, I already noted this in the description.

audformat/core/database.py
frankenjoe: Correct, at this stage we only ensure that the indices are combined. This is required later on when we update the tables as they assume that the dtypes of the indices are alike. I added the following comment:

```
                    # join labels from misc tables
                    # by combining the index of all tables,
                    # columns values will be updated
                    # later when the tables are joined
```

audformat/core/database.py
hagenw: I changed to 

```python
for table_id in list(db):
    table = db[table_id]
```
in `audb`.

PR #278, opened 2022-08-05
Release 0.15.0

hagenw:
My try on summarizing our big chunk of work:

image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ Version 0.15.0 (2022-08-05)
   which can store labels
   not associated with media files
 * Added: usage of a misc table as labels in a scheme
-* Added: dictionary ``audformat.Database.misc_table``
+* Added: dictionary ``audformat.Database.misc_tables``
   holding misc tables of a database
 * Added: ``audformat.utils.difference()``
   for finding index entries
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,12 +32,10 @@ Version 0.15.0 (2022-08-05)
 * Changed: ``audformat.Database.update()``
   can now join schemes
   with different labels
-* Changed: ``audformat.utils.union()``
-  and ``audformat.utils.intersect()``
+* Changed: ``audformat.utils.union()``,
+  ``audformat.utils.intersect()``,
+  and ``audformat.utils.concat()``
   now support any kind of index
-* Changed: ``audformat.utils.concat()``
-  can now concatenate objects
-  that contain any kind of index
 * Changed: ``audformat.Database.__iter__``
   iterates through all (misc) tables,
   e.g. a user can do ``list(db)``
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -71,7 +71,7 @@ Version 0.15.0 (2022-08-05)
   ``audformat.Split``,
   ``audformat.Table``
 * Fixed: type hint of argument ``dtype``
-  in ``audformat.Scheme``
+  in the documentation of ``audformat.Scheme``
 * Removed: support for Python 3.7
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,7 +53,7 @@ Version 0.15.0 (2022-08-05)
   is converted to matching dtype of scheme
   before checking if the value lies in the max-min-range
   of the scheme
-* Fixed: links to ``pandas`` objects in the documentation
+* Fixed: links to ``pandas`` API in the documentation
 * Fixed: include methods
   ``to_dict()``,
   ``from_dict()``,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,10 +49,11 @@ Version 0.15.0 (2022-08-05)
   in the emodb example
   in the documentation
 * Fixed: ensure column IDs are different from index level names
-* Fixed: make sure incoming data
-  is converted to matching dtype of scheme
-  before checking if the value lies in the max-min-range
-  of the scheme
+* Fixed: make sure
+  ``audformat.Column.set()``
+  converts data to dtype of scheme
+  before checking if values are in min-max-range
+  of scheme
 * Fixed: links to ``pandas`` API in the documentation
 * Fixed: include methods
   ``to_dict()``,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -62,7 +62,7 @@ Version 0.15.0 (2022-08-05)
   and attributes
   ``description``,
   ``meta``
-  for the classes
+  in the documentation for the classes
   ``audformat.Column``,
   ``audformat.Database``,
   ``audformat.Media``,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,9 +11,9 @@ Version 0.15.0 (2022-08-05)
 ---------------------------
 
 * Added: ``audformat.MiscTable``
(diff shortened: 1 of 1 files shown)


hagenw:
It's still marked as WIP as we have other open pull requests, but feel free to review already the text of the changelog.

hagenw:
All needed pull requests are now merged and we are ready for the release.

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 0.15.0 (2022-08-05)
+---------------------------
+
+* Added: ``audformat.MiscTable``
+  which can store labels
```
frankenjoe, line 14: I would even use a more generic here:

```suggestion
  which can store data
```

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
...
+  not associated with media files
+* Added: usage of a misc table as labels in a scheme
```
frankenjoe, line 16:
```suggestion
* Added: store scheme labels in a misc table
```

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
...
+* Added: ``audformat.utils.set_index_dtypes()``
+  to change dtypes of an index
+* Added: ``audformat.testing.add_misc_table()``
+* Changed: ``audformat.Database.update()``
+  can now join schemes
+  with different labels
+* Changed: ``audformat.utils.union()``
+  and ``audformat.utils.intersect()``
+  now support any kind of index
+* Changed: ``audformat.utils.concat()``
```
frankenjoe, line 38: I would also add this to the previous bullet

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
...
+  can now concatenate objects
+  that contain any kind of index
+* Changed: ``audformat.Database.__iter__``
+  iterates through all (misc) tables,
+  e.g. a user can do ``list(db)``
```
frankenjoe, line 43:
```suggestion
  e.g. a user can do ``list(db)``
  to get a list of all (misc) tables
```

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
...
+* Changed: require ``pandas>=1.4.1``
```
frankenjoe, line 44: Should we mention that this also requires `Python >= 3.7`?

hagenw: It requires `Python >= 3.8`.
So far I decided to just mention at the end, that we removed support for Python 3.7 without any reason

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
...
+* Changed: use ``pandas`` dtype ``'string'``
+  instead of ``'object'``
+  for storing ``audformat`` dtype ``'str'`` entries
+* Changed: use a misc table
+  to store the ``'speaker'`` scheme labels
+  in the emodb example
+  in the documentation
+* Fixed: ensure column IDs are different from index level names
+* Fixed: make sure incoming data
```
frankenjoe, line 52: I think we should mention that this applies to `Column.set()` so maybe:

```
Column.set()
converts data to dtype of scheme
before checking if values 
are in range
```

CHANGELOG.rst
hagenw: I changed it accordingly.

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
...
+  ``audformat.Column``,
+  ``audformat.Database``,
+  ``audformat.Media``,
+  ``audformat.Rater``,
+  ``audformat.Scheme``,
+  ``audformat.Split``,
+  ``audformat.Table``
+* Fixed: type hint of argument ``dtype``
+  in the documentation of ``audformat.Scheme``
+* Removed: support for Python 3.7
```
frankenjoe, line 74: Ah ok, you mention it down here :)

CHANGELOG.rst
```diff
@@ -7,6 +7,74 @@ The format is based on `Keep a Changelog`_,
...
+* Changed: use a misc table
+  to store the ``'speaker'`` scheme labels
+  in the emodb example
+  in the documentation
+* Fixed: ensure column IDs are different from index level names
+* Fixed: make sure incoming data
+  is converted to matching dtype of scheme
+  before checking if the value lies in the max-min-range
+  of the scheme
+* Fixed: links to ``pandas`` objects in the documentation
```
frankenjoe, line 56:
```suggestion
* Fixed: links to ``pandas`` API in the documentation
```

frankenjoe:
Maybe also add:

```
* Changed: audformat.utils.join_labels()`
  raises ``ValueError`` if labels are of
  different dtype 
```

CHANGELOG.rst
hagenw: I mention `Column.set()` now here

hagenw:
> different dtype

Good idea, I added it as well

CHANGELOG.rst
```diff
@@ -7,6 +7,77 @@ The format is based on `Keep a Changelog`_,
...
+  to change dtypes of an index
+* Added: ``audformat.testing.add_misc_table()``
+* Changed: ``audformat.Database.update()``
+  can now join schemes
+  with different labels
+* Changed: ``audformat.utils.union()``,
+  ``audformat.utils.intersect()``,
+  and ``audformat.utils.concat()``
+  now support any kind of index
+* Changed: ``audformat.Database.__iter__``
```
frankenjoe, line 39: I think we added this feature, so it should be:

```suggestion
* Added: ``audformat.Database.__iter__``
```

CHANGELOG.rst
```diff
@@ -7,6 +7,77 @@ The format is based on `Keep a Changelog`_,
...
+* Added: ``audformat.utils.set_index_dtypes()``
+  to change dtypes of an index
+* Added: ``audformat.testing.add_misc_table()``
+* Changed: ``audformat.Database.update()``
+  can now join schemes
+  with different labels
+* Changed: ``audformat.utils.union()``,
+  ``audformat.utils.intersect()``,
+  and ``audformat.utils.concat()``
+  now support any kind of index
```
frankenjoe, line 42: I think we also need to mention that we convert any integer type to `Int64`

CHANGELOG.rst
hagenw: Yes, that's correct.
I added it in a commit to also move it upwards.

CHANGELOG.rst
hagenw: This is not really a change.
Before integer dtype was not supported.

frankenjoe: Ah yes, that's true. But what we dropped is the special treatment when providing a mix of `filewise` and `segmented` index objects.

frankenjoe:
Cool, except for that last conversation that is still open we are good to go I think!

CHANGELOG.rst
hagenw: Good point, I tried to summarize this in an additional point now with:

```rst
* Changed: ``audformat.utils.intersect()``
  no longer removes segments
  from a segmented index
  that are contained
  in a filewise index
```

PR #279, opened 2022-08-05
WIP: Make Scheme.[dtype|labels] a property

### LINKED ISSUE #63: Make schemes safe against user changes
At the moment a user can change the dtype or labels of a scheme, but this will not automatically update the related tables.
So we should make those attributes properties that return a copy and provide setter functions. BTW, you can add a setter function to a Python property, so a user should be able to do something like `schemes['my-scheme'].labels[0] = 'a'` if we want to allow for it.

hagenw:
Closes #63.

But I'm not sure if this really is a good idea.

### DIFF
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -155,14 +155,29 @@ class Scheme(common.HeaderBase):
                     )
                 dtype = dtype_labels
 
-        self.dtype = dtype
-        r"""Data type"""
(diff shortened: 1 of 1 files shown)


frankenjoe:
The problem is that with the current logic `dtype` and `lables` are no longer stored as YAML when you make them private attributes and introduce a property instead.

PR #282, opened 2022-08-09
Add Scheme.uses_table

### LINKED ISSUE #280: Property to check if scheme uses a misc table
Currently, we use `isinstance(scheme.labels, str)` to check if a scheme is linked to a misc table. To make the code easier to read we should think about a property, maybe something like  `Scheme.is_table`?

frankenjoe:
Closes #280 

image

### DIFF
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -166,7 +166,7 @@ class Scheme(common.HeaderBase):
 
     @property
     def uses_table(self) -> bool:
-        r"""Scheme has labels stored in a table.
+        r"""Scheme has labels stored in a misc table.
 
         If property is ``True``
         the attribute ``labels``
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -176,7 +176,7 @@ class Scheme(common.HeaderBase):
         values are stored.
 
         Returns:
-            ``True`` if scheme has labels stored in a table
+            ``True`` if scheme has labels stored in a misc table
 
         """
         return isinstance(self.labels, str)
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -166,10 +166,17 @@ class Scheme(common.HeaderBase):
 
     @property
     def has_table(self) -> bool:
-        r"""Scheme has labels stored in a misc table.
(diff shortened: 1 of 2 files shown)


frankenjoe:
Turned out there are not so many places within `audformat` where we can actually use the property, but I think it will also make code outside of the package easier to understand, e.g. in https://github.com/audeering/audb/pull/222

hagenw:
Yes, I also think it will be helpful outside of `audformat` even if not used to much inside `audformat` itself.

What I'm not completely convinced yet is the name. `has_table` somehow implies that the scheme has a table, which is not completely correct as the labels of the scheme are stored in a misc table that exists outside of the scheme.

One alternative would be to start the name also with `labels` to make it appear near the `labels` attribute in the documentation, e.g. `labels_stored_in_table`.

Otherwise, maybe `uses_table` might be better than `has_table`?

frankenjoe:
Ok, I would prefer `uses_table`  over `labels_stored_in_table`.

hagenw:
> Ok, I would prefer `uses_table` over `labels_stored_in_table`.

Good, I'm fine with `uses_table`

frankenjoe:
I renamed the property to `uses_table`

audformat/core/scheme.py
```diff
@@ -164,12 +164,29 @@ def __init__(
         self.maximum = maximum if self.is_numeric else None
         r"""Maximum value"""
 
+    @property
+    def uses_table(self) -> bool:
+        r"""Scheme has labels stored in a table.
```
hagenw, line 169:
```suggestion
        r"""Scheme has labels stored in a misc table.
```

audformat/core/scheme.py
```diff
@@ -164,12 +164,29 @@ def __init__(
...
+
+        If property is ``True``
+        the attribute ``labels``
+        is set to an ID of a
+        :class:`audformat.MiscTable`
+        where the actual label
+        values are stored.
+
+        Returns:
+            ``True`` if scheme has labels stored in a table
```
hagenw, line 179:
```suggestion
            ``True`` if scheme has labels stored in a misc table
```

PR #284, opened 2022-08-10
DOC: fix description of Schemes.labels

### LINKED ISSUE #283: Documentation for Scheme.labels indicates wrongly that it returns a list
See

image

but it might also return a dictionary or a string.

hagenw:
Closes #283 

image


frankenjoe:
Do we need to mention that `lables` can be empty?

hagenw:
I don't think so.

PR #291, opened 2022-08-11
Fix: [Table,MiscTable].drop_index() by consider only intersection of table and index

### LINKED ISSUE #290: Table.drop_index() may extend the index
Segments or files that are not in the table will be added.

```python
db['table'].drop_index(audformat.filewise_index(['f2', 'f3']), inplace=True)
db['table'].index
```
```
Index(['f1', 'f3'], dtype='string', name='file')
```

expected:

```
Index(['f1'], dtype='string', name='file')
```

frankenjoe:
Closes #290

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -316,6 +316,7 @@ class Base(HeaderBase):
 
         _assert_table_index(table, index, 'drop rows from')
 
+        index = utils.intersect([table.index, index])
         new_index = utils.difference([table.index, index])
         table._df = table.df.reindex(new_index)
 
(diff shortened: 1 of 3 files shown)


audformat/core/table.py
```diff
@@ -316,6 +316,7 @@ def drop_index(
 
         _assert_table_index(table, index, 'drop rows from')
 
+        index = utils.intersect([table.index, index])
```
hagenw, line 319: It's a little bit unfortunate that we need to add another step.

I checked if there is a faster implementation by testing:

```python
        new_index = [idx for idx in list(table.index) if idx not in index]
        new_index = utils._alike_index(table.index, new_index)
        table._df = table.df.reindex(new_index)
```
but it took more or less the same time, so I would propose to stay with your implementation.

hagenw:
Thanks for adding all those new tests.

I will copy and paste them for `pick_index()` after this is merged.

Issue #287, opened 2022-08-10
Add option inplace to expand_file_path() and map_file_path()?


hagenw:
If you want to change an index of a table with one of those utility functions, the only way to achieve this is by:

```python
index = audformat.utils.map_file_path(table.index, mapping)
table._df.index = index
```

which is not how a user is supposed to interact with the tables.
So, it would made sense to add an option `inplace` to those functions to allow changing the index directly in the table:

```python
audformat.utils.map_file_path(table.index, mapping, inplace=True)
```

frankenjoe:
But that means we have to pass the `Table` object to the function.

frankenjoe:
The method we currently have for this use case is `Database.map_files()`, but it applies it to all tables of course.

hagenw:
An alternative to `inplace` would be to add a method to `audformat.Table` that allows to replace the index (instead of only extending it).

frankenjoe:
Or we add `Table.map_files()` that works like `Database.map_files()` on one table?

hagenw:
If this is able to replace the usage of the two utility functions (`replace_file_extension()` is another one) at

https://github.com/audeering/audb/blob/v1.3.0/audb/core/load.py#L631-L647

then it might be a solution.

frankenjoe:
`map_files()` takes a function as input, so I guess you can do whatever you like with the file names.

PR #293, opened 2022-08-12
Add Table.map_files()

frankenjoe:
Closes https://github.com/audeering/audformat/issues/287

* Database:

image

* Table:

image

* utils:

image

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1485,40 +1485,15 @@ class Table(Base):
     ):
         r"""Apply function to file names in table.
 
-        Relies on :meth:`pandas.Index.map`,
-        which can be slow.
         If speed is crucial,
-        consider to change the index directly.
-        In the following example we prefix every file with a folder:
-
-        .. code-block:: python
-
-            root = '/root/'
-            if table.is_filewise:
-                table.df.index = root + table.df.index
-                table.df.index.name = audformat.define.IndexField.FILE
-            elif len(table.df.index) > 0:
-                table.df.index.set_levels(
-                    root + table.df.index.levels[0],
-                    audformat.define.IndexField.FILE,
-                    inplace=True,
-                )
+        see :func:`audformat.utils.map_file_path`
+        for further hints how to optimize your code.
 
         Args:
             func: map function
 
         """
-        if self.is_segmented:
-            index = self.df.index.map(
-                lambda x: (func(x[0]), x[1], x[2])
-            )
-            index = utils.set_index_dtypes(
-                index,
-                {define.IndexField.FILE: 'string'},
-            )
-            self.df.index = index
-        else:
-            self.df.index = self.df.index.map(lambda x: func(x))
+        self.df.index = utils.map_file_path(self.df.index, func)
 
     def pick_files(
             self,
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1479,6 +1479,47 @@ class Table(Base):
 
         return result
 
+    def map_files(
+            self,
+            func: typing.Callable[[str], str],
+    ):
+        r"""Apply function to file names in table.
+
+        Relies on :meth:`pandas.Index.map`,
+        which can be slow.
+        If speed is crucial,
+        consider to change the index directly.
+        In the following example we prefix every file with a folder:
+
+        .. code-block:: python
+
+            root = '/root/'
+            if table.is_filewise:
+                table.df.index = root + table.df.index
+                table.df.index.name = audformat.define.IndexField.FILE
+            elif len(table.df.index) > 0:
+                table.df.index.set_levels(
+                    root + table.df.index.levels[0],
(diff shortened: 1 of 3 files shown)


audformat/core/table.py
```diff
@@ -1479,6 +1479,47 @@ def get(
...
+            index = self.df.index.map(
+                lambda x: (func(x[0]), x[1], x[2])
+            )
+            index = utils.set_index_dtypes(
+                index,
+                {define.IndexField.FILE: 'string'},
+            )
+            self.df.index = index
+        else:
+            self.df.index = self.df.index.map(lambda x: func(x))
```
hagenw, line 1521: I think we should use
```python
self.df.index = audformat.utils.map_file_path(table.df.index, func)
```
here and make sure that the `string` dtype is preserved inside that function.

audformat/core/table.py
```diff
@@ -1479,6 +1479,47 @@ def get(
...
+            root = '/root/'
+            if table.is_filewise:
+                table.df.index = root + table.df.index
+                table.df.index.name = audformat.define.IndexField.FILE
+            elif len(table.df.index) > 0:
+                table.df.index.set_levels(
+                    root + table.df.index.levels[0],
+                    audformat.define.IndexField.FILE,
+                    inplace=True,
+                )
```
hagenw, line 1505: This example could be replaced by

```python
table.df.index = audformat.utils.expand_file_path(table.df.index, root)
```

hagenw (commented):
I wasn't aware that you can do `table.df.index` and thought a user would need `table._df.index` which would be a no go.
But with `table.df.index` maybe we do not need `map_files()` at all?

audformat/core/table.py
frankenjoe: I would not change to a utility function here since the purpose is to give the user some hint how to optimize independent of the actual task.

hagenw: OK, then maybe change to an example that cannot be done by a utility function?

audformat/core/table.py
frankenjoe: Changed it accordingly. It's a bit unfortunate we have chosen a different name for the utility function.

audformat/core/table.py
frankenjoe: I changed the example and moved it to `utils.map_file_path()`.

audformat/core/utils.py
```diff
@@ -1034,6 +1055,11 @@ def map_file_path(
     else:
         index = index.map(func)
 
+    index = set_index_dtypes(
+        index,
+        {define.IndexField.FILE: 'string'},
+    )
```
hagenw, line 1061: I think in the corresponding test `pd.testing.assert_index_equal` doe snot check for the correct dtype.
We need to use `pd.testing.assert_index_equal(..., exact=True)` or manualy test for the right dtype as the test should have failed before the addition of this code.

audformat/core/utils.py
```diff
@@ -1003,6 +1003,27 @@ def map_file_path(
 ) -> pd.Index:
     r"""Apply callable to file path in index.
 
+    Relies on :meth:`pandas.Index.map`,
+    which can be slow.
+    If speed is crucial,
+    consider to change the index directly.
+    In the following example we prefix every file with a folder
+    and add a new extension:
```
hagenw, line 1011: I'm now fine with staying with the example, but maybe we should add a link to the corresponding utility function that can achieve the same?

```suggestion
    and add a new extension,
    compare also :func:`audformat.utils.expand_file_path`
    and :func:`audformat.utils.replace_file_extension`:
```

audformat/core/utils.py
frankenjoe: I think it would it would fail if the `dtypes` do not match, e.g.:

```python
pd.testing.assert_index_equal(pd.Index(['a']), pd.Index(['a'], dtype='string'))
```
```
Attribute "dtype" are different
[left]:  object
[right]: string[python]
```

But I also tried to create an example where `string` is converted to `object` and couldn't find one. Unless you do e.g. `lambda x: 0`, but that does not match the type hint. So I guess we could also skip `set_index_dtypes()`. On the other hand, it has almost no overhead since it converts if the dtype is not `'string'`.

hagenw: I think it is indeed not possible to find a conversion function that converts to `object` from `string`.

When we leave the current code, we implicitly support mapping functions like `lambda x: 0` as everything that can be converted to a string will be converted to a string by it. So the question is if we want to have this or not?

frankenjoe: Ok, if you agree I will remove it.

hagenw: Yes, fine with me.

frankenjoe: done

hagenw: Looks like you forgot to push.

PR #308, opened 2022-09-15
Fix minimum/maximum scheme checking for None

### LINKED ISSUE #307: Checking of minimum/maximum in scheme fails for array containing None
If you try to set a `np.array` to a column containing a scheme with minimum/maximum the checking fails if the array is of dtype object and contains `None` instead of `np.NaN`:

```python
import audformat.testing
import numpy as np


audformat.testing.add_table(db, 'table', 'filewise', num_files=2)
db.schemes['scheme'] = audformat.Scheme('float', minimum=1.0)
db['table']['column'].set(np.array([1.0, None]))
```
which returns
```
TypeError: '<=' not supported between instances of 'float' and 'NoneType'
```

Thanks to @hesamsagha for reporting.

hagenw:
Closes #307 

When a user tries to assign `np.array([1.0, None])` to a float scheme it will be automatically converted to float:

```python
import audformat.testing
import numpy as np


audformat.testing.add_table(db, 'table', 'filewise', num_files=2)
db.schemes['scheme'] = audformat.Scheme('float')
db['table']['column'].set(np.array([1.0, None]))
print(db['table']['column'].get())
```
returns
```
file
audio/001.wav    1.0
audio/002.wav    NaN
Name: column, dtype: float64
```

But if the scheme contains a minimum or maximum it fails as it needs to do the conversion first (see also https://github.com/audeering/audformat/pull/219) otherwise it fails as shown in #307.

This is fixed here by converting to float before looking for min and max in the array.

### DIFF
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -42,6 +42,10 @@ def assert_values(
                 values = values.values
             if isinstance(values, np.ndarray):
                 if scheme.is_numeric:
+                    # Support type object with None entries,
+                    # which need to be converted to NaN
+                    # to support min/max
+                    values = values.astype(float)
                     values = [
                         np.nanmin(values),
                         np.nanmax(values),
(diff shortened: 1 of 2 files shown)


hagenw:
@hesamsagha as @frankenjoe is not available at the moment, I assigned you as a reviewer.

PR #310, opened 2022-09-16
Show values not matching scheme in Column.set()

### LINKED ISSUE #303: Display which values are not consistent with the scheme
Description:

When a value is not consistent with the given scheme, it would be hard to debug to find which values cause it. 
The audformat just displays `ValueError: Some value(s) do not match scheme`.

Feature request:
Add a report showing which values are not in the scheme. Something like this code-snippet in the column.py

``` 
  except ValueError:
      illegal = [value not in scheme for value in values]
      raise ValueError(values[illegal])
```

hagenw:
Closes #303.

When setting values to a column that do not match a scheme we did not list the wrong values before, this is now changed:

```python
import audformat

db = audformat.Database('db')
db.schemes['scheme'] = audformat.Scheme(labels=['a'])
db['table']['column'].set(['b', 'c'])
```
which returns
```
ValueError: Some value(s) do not match scheme
dtype: str
labels: [a]
with scheme ID 'scheme':
'b', 'c'
```

Only unique values are listed:

```python
db['table']['column'].set(['b', 'b'])
```
returns
```
ValueError: Some value(s) do not match scheme
dtype: str
labels: [a]
with scheme ID 'scheme':
'b'
```

If more than 50 values are affected, they are cut to 50 and `, ...` is added at the end of the error message.

### DIFF
--- a/audformat/core/column.py
+++ b/audformat/core/column.py
@@ -27,7 +27,7 @@ def assert_values(
         scheme: Scheme,
 ):
     r"""Raise error if values do not match scheme."""
-    ok = True
+    bad_values = []
 
     if (
             scheme.labels is not None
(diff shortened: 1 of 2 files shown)


hagenw:
@hesamsagha is the error message understandable?

audformat/core/column.py
```diff
@@ -36,26 +36,31 @@ def assert_values(
...
+                values = [
+                    np.nanmin(values),
+                    np.nanmax(values),
+                ]
+        # Get unique values and preserve order
+        values = list(dict.fromkeys(values))
+        bad_values = [value for value in values if value not in scheme]
+
+    if len(bad_values) > 0:
+        values = str(bad_values[:50])[1:-1]
```
frankenjoe, line 57: I suggest we restrict to 10 items as we also do in https://github.com/audeering/audformat/blob/7c22a3e026e11e58466be6f530ebec4dc790ef34/audformat/core/utils.py#L237 and https://github.com/audeering/audformat/blob/7c22a3e026e11e58466be6f530ebec4dc790ef34/audformat/core/index.py#L139

hagenw: I introduced a `max_display` and set it to 10, which I also think will be easier to read.

PR #318, opened 2022-10-18
Add audformat.Attachment

### LINKED ISSUE #313: Add possibility to add attachments to a database
See also https://github.com/audeering/audb/issues/136

For some databases it might be of great help to also include additional files that are neither tables nor media to the dabase, like JPG, PDF, or JSON files. Those could be needed for a better explanation of the setup when recording the data.

One idea might be to add a `audformat.Attachment` object to the database, that stores a description of the attachment and a relative path to the attachment files. For example:

```python
db.attachments['measurement'] = audformat.Attachment(
    'pics/measurement.jpg',
    description='Photo of the impulse response measurement setup.',
)
```
and then:
```python
>>> db.attachments['measurement']
pics/measurement.jpg
>>> db.attachments['measurement'].description
'Photo of the impulse response measurement setup.'
```

The absolute path could easily be generated by the user as we have `db.root`.

`audb` could then use the `db.attachments` entries to know that those files need also be added to an archive and uploaded.

/cc @hesamsagha

hagenw:
Closes #313 

This allows to store additional files/folders as attachments to a database.

It is a little bit tricky as we not only need a header entry, but also the actual file/folder.
I designed it that you can assign a file/folder path when the file/folder does not exist as long as the database has `db.root = None`. As soon as this changes, e.g. by saving or loading a database or assigning the attachment to an existing database with a given `root` it will then raise an error.

This adds:
* Attachment to the specifications of the database
* `audformat.Attachment`
* `audformat.testing.create_attachemnt_files()`

### Specifications

image

[...]

image

### audformat.Attachment

image

[...]

### audformat.testing.create_attachment_files()

image

### DIFF
--- a/audformat/core/attachment.py
+++ b/audformat/core/attachment.py
@@ -76,7 +76,7 @@ class Attachment(HeaderBase):
 
     def _check_path(
             self,
-            root: str,
+            root: typing.Optional[str],
     ):
         r"""Check if path exists and is not a symlink."""
         if root is None:
--- a/audformat/core/attachment.py
+++ b/audformat/core/attachment.py
@@ -11,10 +11,14 @@ from audformat.core.common import (
 class Attachment(HeaderBase):
     r"""Database attachment.
 
-    Adds a file as attachment to a database.
+    Adds a file or folder
+    as attachment to a database.
+    If a folder is provided,
+    all of its sub-folders
+    and files are included.
 
     Args:
-        path: relative path to file
+        path: relative path to file or folder
         description: attachment description
         meta: additional meta fields
 
@@ -60,9 +64,9 @@ class Attachment(HeaderBase):
                 f"of attachment '{self._id}' "
                 "does not exist."
             )
-        if not os.path.isfile(audeer.path(root, self.path)):
-            raise FileNotFoundError(
+        if os.path.islink(os.path.join(root, self.path)):
+            raise RuntimeError(
                 f"The provided path '{self.path}' "
                 f"of attachment '{self._id}' "
-                "is not a file."
+                "is not allowed to be a symlink."
             )
--- a/audformat/core/attachment.py
+++ b/audformat/core/attachment.py
@@ -19,8 +19,6 @@ class Attachment(HeaderBase):
         ValueError: if ``path`` is absolute
 
     Example:
-        >>> import audeer
-        >>> _ = audeer.touch('file.txt')
         >>> Attachment('file.txt', description='Attached file')
         {description: Attached file, path: file.txt}
 
--- a/audformat/core/attachment.py
+++ b/audformat/core/attachment.py
@@ -2,7 +2,10 @@ import os
 
 import audeer
 
-from audformat.core.common import HeaderBase
+from audformat.core.common import (
+    HeaderBase,
+    is_relative_path,
+)
 
 
 class Attachment(HeaderBase):
@@ -33,14 +36,7 @@ class Attachment(HeaderBase):
     ):
         super().__init__(description=description, meta=meta)
 
-        if (
-                os.path.isabs(path)
-                or '\\' in path
-                or path.startswith('./')
-                or '/./' in path
-                or path.startswith('../')
-                or '/../' in path
-        ):
+        if not is_relative_path(path):
             raise ValueError(
                 f"The provided path '{path}' needs to be relative "
                 "and not contain '\\', '.', or '..'."
--- a/audformat/core/attachment.py
+++ b/audformat/core/attachment.py
@@ -25,6 +25,13 @@ class Attachment(HeaderBase):
     Raises:
         ValueError: if ``path`` is absolute
             or contains ``\``, ``..`` or ``.``
+        RuntimeError: when assigning an attachment
+            to a database,
+            but the database contains another attachment
+            with an path
+            that is identical
+            or nested
+            compared to the current attachment path
 
     Example:
         >>> Attachment('file.txt', description='Attached file')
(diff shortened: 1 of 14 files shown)


hagenw:
After adding this feature here, we will also need to extend `audb` as the attached files needs to be archived and uploaded just as the media files.

audformat/core/attachment.py
```diff
@@ -0,0 +1,72 @@
...
+    ):
+        super().__init__(description=description, meta=meta)
+
+        if (
+                os.path.isabs(path)
+                or '\\' in path
+                or path.startswith('./')
+                or '/./' in path
+                or path.startswith('../')
+                or '/../' in path
```
hagenw, line 42: I will move this to a utility function as we also use it inside `Database.is_portable()`

hagenw: done.

hagenw:
I found folders a little bit dangerous as we don't know what's supposed inside.
One solution to this problem could be to calculate MD5 sums for each attachment, maybe it is possible to also get a value representing a bunch of files in a given structure.

Would this make sense?

frankenjoe:
> maybe it is possible to also get a value representing a bunch of files in a given structure.

We could probably use `audeer.uid()`.

Or we simply list all files under `Attachment` in that case?

hagenw:
Sounds like it should be possible. I will have a look into this and update the code, will mark it as `WIP` in the mean time.

hagenw:
I added now the possibility to also store a folder as an attachment.
As the file/folder needs only to exist when assigning the attachment to a database I decided against storing the structure represented inside an attached folder. I think this is only needed inside `audb` where we have to version the attachment files anyway.

What I do here instead is check that a user does not add nested or identical attachments to a database.

audformat/core/attachment.py
```diff
@@ -0,0 +1,95 @@
...
+                f"Attachments '{self.path}' and '{other.path}' "
+                "are nested."
+            )
+
+    def _check_path(
+            self,
+            root: str,
+    ):
+        r"""Check if path exists and is not a symlink."""
+        if root is None:
```
frankenjoe, line 82: Why do we need this case? According to the type hint `None` is not supported.

audformat/core/database.py
```diff
@@ -583,6 +589,10 @@ def job(obj_id, obj):
                 task_description='Save tables',
             )
 
+            # Check attachments exist
+            for attachment_id in list(self.attachments):
+                self.attachments[attachment_id]._check_path(root)
```
frankenjoe, line 594: Is this is enough for folders or do we also need to check there are no symlinks within a folder? But probably it's sufficient to check this in `audb.publish()` as a user could anyway add files to a folder afterwards.

audformat/core/database.py
```diff
@@ -935,6 +945,11 @@ def load(
         Returns:
             database object
 
+        Raises:
+            FileNotFoundError: if a file associated with an attachment
+                cannot be found
+                or is not a file
```
frankenjoe, line 951:
```suggestion
                or is not a file or folder
```

frankenjoe: Or actually I think we can remove "or is not a file" now that we support folders?

frankenjoe:
What might be handy is a property `Attachment.files` that returns a list of files belonging to the `Attachment`. Or maybe a method `Attachment.list()` that supports arguments like `full_path`, `filetype`, etc.

audformat/core/attachment.py
hagenw: Before I also executed this check when an attachment was added to a database that was not yet stored to disk.
I changed this and no execute this test only when a database has `self.root`.

audformat/core/database.py
hagenw: Yes, I would indeed check only really necessary cases inside `audformat` and would do the rest in `audb`.

hagenw:
> What might be handy is a property `Attachment.files` that returns a list of files belonging to the `Attachment`. Or maybe a method `Attachment.list()` that supports arguments like `full_path`, `filetype`, etc.

Yes, but I would add this in another pull request.

audformat/core/database.py
hagenw: Which I also wrote for `save()`.

frankenjoe:
> Yes, but I would add this in another pull request.

I created https://github.com/audeering/audformat/issues/326

audformat/core/testing.py
```diff
@@ -186,6 +188,31 @@ def add_table(
     return db[table_id]
 
 
+def create_attachment_files(
```
frankenjoe, line 191: I guess this introduces a breaking change, since we now have to call this method before we can save a database created with `audformat.testing.create_db()`. But I guess there is no other solution unless we choose a table as attachment, which does not make much sense.

hagenw: One solution would be that we do not create an attachment with `audformat.testing.create_db()`, but that the user has to add one herself. This might be a cleaner solution?

frankenjoe: Yes, it might be safer, but it means we will not have an `Attachment` entry in the default database, which is also not nice. I'm undecided what's the better way to go :)

frankenjoe: Maybe it's not so bad to introduce a breaking change here as it will probably only affect `audb`, where we anyway need to add support for `Attachment` now.

hagenw: It should also not affect the user part of any package, but only the tests. At least I hope that we use `audformat.testing.create_db()` only for testing inside `audb`.

frankenjoe: Ok, let's stick with the implementation as is.

PR #322, opened 2022-10-31
DOC: improve sketches of database structure

hagenw:
Two changes to improve definition of an `audformat` database

## Use lower case for column names

Changes

image

to

image

## Mention misc table in header

We show how header entries and tables are connected, but where listing misc tables as table before.
To improve this I changed

image

to

image

and

image

to

image

### DIFF
--- a/docs/pics/audformat-misc-table.dot
+++ b/docs/pics/audformat-misc-table.dot
@@ -3,12 +3,12 @@ digraph structs {
     rankdir=LR
     node[shape=record]
 
-    header [label="Header | ... | <table> table | <split> split | media | <rater> rater | <scheme> scheme | ... "]
+    header [label="Header | ... | <misctable> misc table | <table> table | <split> split | media | <rater> rater | <scheme> scheme | ... "]
     table [label="MiscTable | { idx | <column> x | y }"]
 
     header:split -> table [style="dotted"]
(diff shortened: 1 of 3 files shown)


docs/pics/audformat-misc-table.dot
```diff
@@ -3,12 +3,12 @@ digraph structs {
     rankdir=LR
     node[shape=record]
 
-    header [label="Header | ... | <table> table | <split> split | media | <rater> rater | <scheme> scheme | ... "]
+    header [label="Header | ... | <misctable> misc table | <table> table | <split> split | media | <rater> rater | <scheme> scheme | ... "]
```
frankenjoe, line 6: There is no arrow from `media` to `misc table` but actually we support setting a `media ID`: https://audeering.github.io/audformat/api.html#audformat.MiscTable

hagenw: I thought it makes only sense to have media ID for recordings, but for a misc table we don't have recordings.

frankenjoe: Yes, but currently we can set a media ID, so I think this should be reflected also by the diagram.

hagenw: I added the media link for misc tables and updated the screenshot in the description of this pull request.

PR #327, opened 2022-12-06
WIP: DOC: additional information on audformat.define module

hagenw:


hagenw:
I spotted a few minor errors and will set to WIP until I have fixed them.

hagenw:
> I spotted a few minor errors and will set to WIP until I have fixed them.

Fixed typos and removed WIP, it's ready for review now.

audformat/core/define.py
```diff
@@ -26,39 +40,79 @@ class DataType(DefineBase):
...
-        :ref:`table specifications <data-tables:Tables>`.
+    r"""Index types of a table.
+
+    Specifies the string values
+    representing a filewise or segmented index.
+    Those string values are returned by
+    :attr:`audformat.Table.type`
+    and :func:`audformat.index_type`.
+    The exact string values are not part
+    of the :ref:`table specifications <data-tables:Tables>`,
```
hagenw, line 88: I just realized that this is not correct.

The `IndexType` is part of the format specifications, see

image

I will update accordingly.

hagenw: I fixed this.

docs/data-header.rst
```diff
@@ -165,8 +169,8 @@ This part of the header is represented by :class:`audformat.Split`.
 Field           Mandatory  Description
 ==============  =========  ====================================================
 id              yes        Unique identifier of split
-type            yes        Split type,
-                           typically one of :class:`audformat.define.SplitType`
```
frankenjoe, line 169: Doesn't make sense to refer to the class here? In case we ever extend it we have to also change it here.

hagenw: As we have the format specification here my goal was to formulate this independent of any implementation.
One consequence is of cause that we have to change it at two different places if we ever want to change it.
But I would argue those changes should happen extremely rarely ;)

frankenjoe: Ok :)

Issue #317, opened 2022-10-17
Always return categorical dtype when get() is called with a mapping


frankenjoe:
Example:

```python
db = audformat.Database('db')
db.schemes['text'] = audformat.Scheme(dtype='str')
db['misc'] = audformat.MiscTable(
    pd.Index(
      ['a', 'b', 'c'],
      name='speaker',
      dtype='string',
    )
)
db['misc']['text'] = audformat.Column(scheme_id='text')
db['misc']['text'].set(['A', 'B', 'C'])

db.schemes['scheme'] = audformat.Scheme('str', labels='misc')
db['table'] = audformat.MiscTable(audformat.filewise_index(['f1', 'f2', 'f3']))
db['table']['column'] = audformat.Column(scheme_id='scheme')
db['table']['column'].set(['a', 'b', 'c'])

db['table']['column'].get(map='text').dtype
```
```
category
```

and if `'text'` contains `None`, we get:

```python
db['misc']['text'].set(['A', 'B', None])
db['table']['column'].get(map='text').dtype
```
```
object
```

But in both cases we would expect ~~`string`~~ `category`.

hagenw:
It's not straightforward what the expected outcome should be.
`string` makes sense as this is the dtype of the column in the misc table.

On the other hand labels where handled quite differently before:

```python
db = audformat.Database('db')
db.schemes['scheme'] = audformat.Scheme('str', labels={'a': 'A', 'b': 'B', 'c': 'C'})
db['table'] = audformat.MiscTable(audformat.filewise_index(['f1', 'f2', 'f3']))
db['table']['column'] = audformat.Column(scheme_id='scheme')
db['table']['column'].set(['a', 'b', 'c'])
db['table']['column'].get(map='text').dtype
```
which returns
```
CategoricalDtype(categories=['A', 'B', 'C'], ordered=False)
```

So I guess it some point we need the `CategoricalDtype` when dealing with labels?

frankenjoe:
@hagenw is right, the expected behavior would be to have a categorical dtype in both cases. I will rename the issue accordingly.

hagenw:
A short note, I don't get `category` as response but the following:

```python
>>> db['table']['column'].get(map='text').dtype
CategoricalDtype(categories=['A', 'B', 'C'], ordered=False)
```

and the second call I expect to look like this:

```python
>>> db['misc']['text'].set(['A', 'B', None])
>>> db['table']['column'].get(map='text').dtype
[...]

Issue #328, opened 2022-12-22
Building documentation raises warning


hagenw:
The warning looks like this:

```
executing update-database update-database
                                                                   
WARNING: Cell printed to stderr:
/home/audeering.local/hwierstorf/.envs/audformat/lib/python3.8/site-packages/pandas/core/internals/blocks.py:2323: RuntimeWarning: invalid value encountered in cast
  values = values.astype(str)                                
                                                                   
WARNING: Cell printed to stderr:
/home/audeering.local/hwierstorf/.envs/audformat/lib/python3.8/site-packages/pandas/core/internals/blocks.py:2323: RuntimeWarning: invalid value encountered in cast
  values = values.astype(str)   
                                                              
looking for now-outdated files... none found
```

and I can recreate it after installing in a fresh virtual environment.

hagenw:
This is still mysterious to me.

I'm able to reproduce this error locally when building the docs.
On the other hand, when executing all the code in the usage section manually, there are now warnings popping up.

Issue #324, opened 2022-11-09
CategoricalDtype not preserved when using index in Table.get()


hagenw:
When indexing a table that contains a scheme with labels with an index when getting it with `get(index)` we are loosing the information about allowed categories as it returns no longer a `CategoricalDtype` object. This can also break your code as the column can no longer be accessed via `.cat` which is possible without using an index.

This is related to https://github.com/audeering/audformat/issues/317 as we also loose this information when using `map`.

```python
import audformat


db = audformat.Database('db')

db.schemes['spk'] = audformat.Scheme('str', labels=['a', 'b'])
index = audformat.filewise_index(['f1'])
db['files'] = audformat.Table(index)
db['files']['spk'] = audformat.Column(scheme_id='spk')
db['files']['spk'].set(['a'])

db.schemes['label'] = audformat.Scheme('int')
index = audformat.segmented_index(['f1', 'f1'], [0, 1], [1, 2])
db['segments'] = audformat.Table(index)
db['segments']['label'] = audformat.Column(scheme_id='label')
db['segments']['label'].set([0, 1])
```

Now we can do the following:

```python
>>> df = db['files'].get()
>>> df.spk.cat.categories
Index(['a', 'b'], dtype='object')
>>> df.loc['f1', 'spk'] = 'c'
...
TypeError: Cannot setitem on a Categorical with a new category (c), set the categories first
```

But, when indexing by the other table we loose all this information:

```python
>>> df = db['files'].get(db['segments'].index)
>>> df.spk.cat.categories
...
AttributeError: Can only use .cat accessor with a 'category' dtype
>>> df.loc['f1', 'spk'] = 'c'
>>> df
                                     spk
file start           end                
f1   0 days 00:00:00 0 days 00:00:01   c
     0 days 00:00:01 0 days 00:00:02   c
```

hagenw:
The question is also what should be the desired behavior?

Returning a `CategoricalDtype` column has the disadvantage that you cannot easily modify the values in that column, but need first to change the assigned categories. As `get()` returns a copy of the data anyway, which you would like to be able to modify, this is maybe not a good behavior in the first place.

frankenjoe:
To me the cleanest solution would be that the mapped column has type `CategoricalDtype`.

Issue #256, opened 2022-07-29
Cannot load database if meta contains key 'meta'


frankenjoe:
Minimal example that fails:

```python
db = audformat.testing.create_db(minimal=True)
db.meta['meta'] = 'value'
db.save('db')
audformat.Database.load('db')
```
```
Traceback (most recent call last):
  File "/media/jwagner/Data/Git/pyaudformat/my/my.py", line 24, in <module>
    audformat.Database.load('db')
  File "/media/jwagner/Data/Git/pyaudformat/audformat/core/database.py", line 917, in load
    db = Database.load_header_from_yaml(header)
  File "/media/jwagner/Data/Git/pyaudformat/audformat/core/database.py", line 977, in load_header_from_yaml
    db.from_dict(header, ignore_keys=['media', 'misc_tables', 'raters',
  File "/media/jwagner/Data/Git/pyaudformat/audformat/core/common.py", line 191, in from_dict
    assert isinstance(value, type(self.__dict__[key]))
AssertionError
```

PR #333, opened 2022-12-22
Release 0.16.0

hagenw:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
-Version 1.0.0 (2023-01-11)
+Version 1.0.0 (2023-01-12)
 ---------------------------
 
 * Added: ``audformat.Attachment`` to store
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,17 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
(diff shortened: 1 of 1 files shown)


hagenw:
I thought it's time for the big 1.0.0 release, do you agree?

frankenjoe:
> I thought it's time for the big 1.0.0 release, do you agree?

Yes, good idea. We have a couple of issues marked as bug. If we go for 1.0.0 we should probably check whether those bugs still exist and if yes try to solve them first.

hagenw:
We have the following open bugs at the moment:

* https://github.com/audeering/audformat/issues/256: this shouldn't be urgent
* https://github.com/audeering/audformat/issues/317: this will be not that easy to solve
* https://github.com/audeering/audformat/issues/324: this is related to the previous issue. In summary we need to make sure that `CategoricalDtype` is preserved by methods like `get()`
* https://github.com/audeering/audformat/issues/328: this should not affect the user, only developers

It might be that we want to fix #317 and #324 first before making the release. The downside is that we need a release to continue working on the attachment feature in `audb`. We could change the version of this release to `0.16.0`, but I'm also fine with doing `1.0.0` and release the somehow breaking change we are doing when solving #317 and #324 as a bug fix.

frankenjoe:
I created a fix for #256 

As you already said, #317 and #324 are not that straightforward to solve. Yet, I think we should solve them before we release 1.0.0. So I would go for 0.16.0 for the moment.

hagenw:
I changed the version to `0.16.0` and mention your fix from #256

hagenw:
#256 is merged and this should now be ready for review.

PR #339, opened 2023-01-10
Always set dtype of file level to 'string'

### LINKED ISSUE #338: Mismatch between string and object type in Database.files
Currently, we are not forcing the `dtype` of the `file` level of an `audformat` index to be `string`. This can lead to the following problem:

```python
import pandas as pd

import audformat

index = audformat.filewise_index(['f1', 'f2'])  # dtype is string
db['t1'] = audformat.Table(index)
db['t1'].files
```
```
```

```python
files = ['f3', 'f4']
index = pd.Index(['f3', 'f4'], name='file')  # dtype is object
db['t2'] = audformat.Table(index)
db['t2'].files
```
```
Index(['f3', 'f4'], dtype='object', name='file')
```

```python
db.files
```
```
ValueError: Levels and dtypes of all objects must match. Found different level dtypes: ['str', 'object'].
```

I guess we have several options:

1. require that the `dtype` is `string`, then the second part would simply fail
2. implicitly convert from `object` to `string` when the table is created (and probably also in other functions like `extend_index()`
3. at least convert to `string` in `Table.index` 

Since 1. is not backward compatible I guess 2. is the safest option.

frankenjoe:
Closes #338 

Ensures file level of an audformat index is set to dtype 'string' in the following methods:

* `Table.__init__()`
* `Table.drop_index()`
* `Table.extend_index()`
* `Table.pick_index()`

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1560,13 +1560,13 @@ def _assert_file_level_dtype_is_string(
 ) -> pd.Index:
     r"""Possibly set dtype of file level to 'string'."""
     if (
-        (
-            is_filewise_index(index)
-            and index.dtype == 'object'
-        ) or (
-            is_segmented_index(index)
-            and index.dtypes[define.IndexField.FILE] == 'object'
-        )
+            (
+                is_filewise_index(index)
+                and index.dtype == 'object'
+            ) or (
+                is_segmented_index(index)
+                and index.dtypes[define.IndexField.FILE] == 'object'
+            )
     ):
         index = utils.set_index_dtypes(
             index,
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -314,7 +314,7 @@ class Base(HeaderBase):
         """
         table = self if inplace else self.copy()
 
-        index = _assert_file_level_dtype_is_string(index)
+        index = _maybe_convert_dtype_to_string(index)
         _assert_table_index(table, index, 'drop rows from')
 
         index = utils.intersect([table.index, index])
@@ -354,7 +354,7 @@ class Base(HeaderBase):
         """
         table = self if inplace else self.copy()
 
-        index = _assert_file_level_dtype_is_string(index)
+        index = _maybe_convert_dtype_to_string(index)
         _assert_table_index(table, index, 'extend')
 
         new_index = utils.union([table.index, index])
@@ -560,7 +560,7 @@ class Base(HeaderBase):
         """
         table = self if inplace else self.copy()
 
-        index = _assert_file_level_dtype_is_string(index)
+        index = _maybe_convert_dtype_to_string(index)
         _assert_table_index(table, index, 'pick rows from')
 
         new_index = utils.intersect([table.index, index])
@@ -920,7 +920,7 @@ class Base(HeaderBase):
             ):
                 df[column_id] = df[column_id].astype('string', copy=False)
         # Fix index entries as well
-        df.index = _assert_file_level_dtype_is_string(df.index)
+        df.index = _maybe_convert_dtype_to_string(df.index)
 
         self._df = df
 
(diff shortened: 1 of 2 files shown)


frankenjoe:
`Table.get()` and `Column.set()` I did not change as they are not affected by the bug described in #338 or should we convert the index dtype for consistency there, too?

hagenw:
> `Table.get()` and `Column.set()` I did not change as they are not affected by the bug described in #338 or should we convert the index dtype for consistency there, too?

How can `Table.get()` be affected? For this we have to load the data first from CSV (or from PKL, which was also created from CSV) and there we always set the index to `string` dtype. It could only be affected by some old cache files. So, I think we don't need to change it.

Can `Column.set()` change the dtype of the underlying index? Or is the idea to fix the case when you provide a non `string` type index to `set()`?

frankenjoe:
>  How can Table.get() be affected? For this we have to load the data first from CSV (or from PKL, which was also created from CSV) and there we always set the index to string dtype.

Mh, I don't think you have to load a database from disk before you can use `get()` and `set()`, e.g.:

```python
import audformat

index = audformat.filewise_index(['f1', 'f2'])  # dtype is string
db['t'] = audformat.Table(index)
db['t']['c'] = audformat.Column()

index = pd.Index(['f1'], name='file')
db['t']['c'].set(['a'], index=index)
db['t'].get(index)
```
```
      c
file   
f1    a
```

hagenw:
To me it feels like

```python
index = pd.Index(['f1'], name='file')  # dtype is object
db['t']['c'].set(['a'], index=index)
```
should raise an error as it tries to add a non-supported index.

Your solution would be to silently convert to `string` dtype, which would also be fine, but then we need to do it inside `set()` as well. `get()` could be affected if a user manipulates the dataframe manually, but I'm not sure if we really should add a conversion here as well.

frankenjoe:
> should raise an error as it tries to add a non-supported index.

Yes that was the other solution I suggested in #338, but would be a breaking change.

frankenjoe:
> but then we need to do it inside set() as well

We only use it to select the rows for which we want to set new values and currently, `df.loc[index]` does not care about the `dtype`, so there is not really a need to change it. But maybe it's  safer as they may change the behavior in pandas. But then of course we should also apply the conversion in `get()`, too.

audformat/core/table.py
```diff
@@ -1562,6 +1555,26 @@ def _get_by_index(
         return result, result_is_copy
 
 
+def _assert_file_level_dtype_is_string(
```
hagenw, line 1558: `assert` sounds to me as if it would raise an error if dtype is not `'string'`.

Maybe rename to `_convert_file_level_dtype_to_string()` or `_ensure_file_level_dtype_is_string()`?

audformat/core/table.py
```diff
@@ -1562,6 +1555,26 @@ def _get_by_index(
...
+) -> pd.Index:
+    r"""Possibly set dtype of file level to 'string'."""
+    if (
+        (
+            is_filewise_index(index)
+            and index.dtype == 'object'
+        ) or (
+            is_segmented_index(index)
+            and index.dtypes[define.IndexField.FILE] == 'object'
+        )
```
hagenw, line 1569:
```suggestion
            (
                is_filewise_index(index)
                and index.dtype == 'object'
            ) or (
                is_segmented_index(index)
                and index.dtypes[define.IndexField.FILE] == 'object'
            )
```

hagenw:
> > but then we need to do it inside set() as well
> 
> We only use it to select the rows for which we want to set new values and currently, `df.loc[index]` does not care about the `dtype`, so there is not really a need to change it. But maybe it's safer as they may change the behavior in pandas. But then of course we should also apply the conversion in `get()`, too.

OK, I don't have a strong opinion on this. Maybe we don't add the conversion until we really need it, but we add a test to make sure it works?

frankenjoe:
> OK, I don't have a strong opinion on this. Maybe we don't add the conversion until we really need it, but we add a test to make sure it works?

Ok, I added additional tests to ensure that `get()` and `set()` work when the dtype is `object`.

PR #347, opened 2023-01-17
Return categorical dtype when get() is called with a mapping

### LINKED ISSUE #317: Always return categorical dtype when get() is called with a mapping

frankenjoe:
Closes #317 

Solves the issue described in #317 and adds a test for it.

### DIFF
--- a/tests/test_column.py
+++ b/tests/test_column.py
@@ -198,6 +198,8 @@ def test_map(column, map):
 
 def test_map_returns_categorical_dtype():
 
+    # See https://github.com/audeering/audformat/issues/317
+    
     # use scheme with labels
 
     db = audformat.testing.create_db(minimal=True)
--- a/tests/test_column.py
+++ b/tests/test_column.py
@@ -196,6 +196,71 @@ def test_map(column, map):
     pd.testing.assert_series_equal(result, expected)
 
 
+def test_map_returns_categorical_dtype():
+
+    # use scheme with labels
+
+    db = audformat.testing.create_db(minimal=True)
+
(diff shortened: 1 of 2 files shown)


tests/test_column.py
```diff
@@ -196,6 +196,71 @@ def test_map(column, map):
     pd.testing.assert_series_equal(result, expected)
 
 
+def test_map_returns_categorical_dtype():
+
+    # use scheme with labels
```
hagenw, line 201: Maybe it makes sense to track to which issue this belongs?

```suggestion
    # See https://github.com/audeering/audformat/issues/317
    
    # use scheme with labels
```

PR #348, opened 2023-01-17
DOC: enhance utils.map_file_path() example

hagenw:
To avoid the following warnings:

```
FutureWarning: In a future version of pandas all arguments of MultiIndex.set_levels except for the argument 'levels' will be keyword-only.
FutureWarning: inplace is deprecated and will be removed in a future version.
```

when executing the alternative example code I updated it by adding a keyword and removed `inplace`.

image


hagenw:
`inplace` will also be removed in a future version, I will update the code accordingly.

hagenw:
Done, is ready for review now.

PR #352, opened 2023-02-13
WIP: Use ruff for PEP8 checks

hagenw:
This uses [ruff](https://github.com/charliermarsh/ruff) for PEP8 checking, which is much faster than `flake8`.

It requires `pyproject.toml` for it's configuration, which is also added here.
But I suggest to first make an extra pull request in which we switch to `pyproject.toml` as it now seems mayor enough and we can replace `setup.cfg` and `setup.py` by using it.

TODO:
* check if build Python package still looks ok
* editable builds as done with `pip -e .` [still](https://twitter.com/pganssle/status/1241161328137515008) do [require](https://github.com/pypa/setuptools/issues/1698) a `setup.py`, so we should check if we really need this


frankenjoe:
> This uses ruff for PEP8 checking, which is much faster than flake8.

How much faster is it? `flake8` never seemed like a bottleneck to me.

frankenjoe:
> But I suggest to first make an extra pull request in which we switch to pyproject.toml as it now seems mayor enough and we can replace setup.cfg and setup.py by using it.

Yes, makes sense.

hagenw:
> > This uses ruff for PEP8 checking, which is much faster than flake8.
> 
> How much faster is it? `flake8` never seemed like a bottleneck to me.

image

Now that we use `pre-commit` to check for `flake8` you have a slight delay every time you do a commit. For most of our projects you do not feel this, but for larger ones like `audformat` there is a small delay.

PR #355, opened 2023-03-22
DOC: add ipykernel to requirements

hagenw:
As discussed in https://github.com/audeering/audformat/pull/354#issuecomment-1479352467,
building the documentation fails as no `python3` kernel can be found.

It seems that the kernel was before installed by some of the package `jupyter-sphinx` depend on, but I was not able to track this down.

I would suggest, we simply add it to the dependencies.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,7 @@ jobs:
 
     - name: Test building documentation
       run: |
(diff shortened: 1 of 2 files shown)


hagenw:
Adding `ipykernel` as a requirement does indeed help:

image

PR #356, opened 2023-03-27
Allow to save database with missing attachments

hagenw:
This removes checks introduced in https://github.com/audeering/audformat/pull/318 that ensure the path of an attachment exists and is no symlink when saving a database to disk or assigning an attachment to a database.

The idea is that we support saving a database with empty attachments and treat them like media files.
The actual checking if the files exist are then done by `audb` when publishing the database, see https://github.com/audeering/audb/pull/262

I did not delete `audformat.Attachment._check_path()` as the method is still used to raise the error for non-existing files or symlinks when calling `audformat.Attachment.files`.

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -970,10 +970,8 @@ class Database(HeaderBase):
             database object
 
         Raises:
-            FileNotFoundError: if a file or folder
-                associated with an attachment
+            FileNotFoundError: if the header file of the database
                 cannot be found
-                when using ``load_data=True``
 
         """
         ext = '.yaml'
@@ -990,11 +988,6 @@ class Database(HeaderBase):
 
             params = []
             table_ids = []
-
-            if 'attachments' in header and header['attachments']:
-                if load_data:
-                    for attachment_id in header['attachments']:
-                        db.attachments[attachment_id]._check_path(root)
 
             if 'tables' in header and header['tables']:
                 for table_id in header['tables']:
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -1169,8 +1169,6 @@ class Database(HeaderBase):
     ) -> Attachment:
         attachment._db = self
         attachment._id = attachment_id
-        if self.root is not None:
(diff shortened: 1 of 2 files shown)


audformat/core/database.py
```diff
@@ -1181,8 +1169,6 @@ def _set_attachment(
     ) -> Attachment:
         attachment._db = self
         attachment._id = attachment_id
-        if self.root is not None:
-            attachment._check_path(self.root)
         for other_id in list(self.attachments):
             attachment._check_overlap(self.attachments[other_id])
```
frankenjoe, line 1173: This can raise a RuntimeError, which we do not mention in the docstring yet.

hagenw: I created https://github.com/audeering/audformat/issues/357 and will handle it in another pull request.

hagenw:
Yes, you are right, `load_data` should be about tables and not attachments.

I updated the code and we now allow `audformat.Database.load(..., load_data=True)` to succeed when attachment files are missing.

frankenjoe:
Maybe we should change the title of the PR to "Allow to save database with missing attachments"?

tests/test_attachment.py
```diff
@@ -98,18 +88,11 @@ def test_attachment(tmpdir):
 
     # Load database
     #
-    # path needs to exist when requesting data
+    # path needs not to exist when loading the database
```
frankenjoe, line 91:
```suggestion
    # path must not exist when loading the database
```

frankenjoe:
Instead of removing the code here and then re-introduce it in `audb.publish()` we could also add an argument like `check_attachments=False`.

hagenw:
> Instead of removing the code here and then re-introduce it in `audb.publish()` we could also add an argument like `check_attachments=False`.

We don't need to do this. https://github.com/audeering/audb/pull/262 passes already all tests when using this branch of `audformat`. The reason is that the error is still raised by `audformat.Attachment.files` which is sufficient.

PR #360, opened 2023-03-29
CI: make doc test an extra Action

hagenw:
Separates the doc test into an extra Action.

It also re-enables failing the doc test when a warning is raised since https://github.com/audeering/audformat/issues/328 is closed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
     - name: Prepare Ubuntu
       run: |
         sudo apt-get update
-        sudo apt-get install --no-install-recommends --yes graphviz libsndfile1
+        sudo apt-get install --no-install-recommends --yes libsndfile1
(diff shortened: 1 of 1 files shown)


.github/workflows/doc.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Ubuntu - install libsndfile
+      run: |
+        sudo apt-get update
+        sudo apt-get install --no-install-recommends --yes libsndfile1
```
frankenjoe, line 35: do we need libsndfile for the documentation?

frankenjoe: and if so, can't we install it together with `graphviz`?

hagenw: I have removed `libsndfile1` as we do indeed not need it and combined the two sections.

PR #361, opened 2023-04-04
Avoid pandas DeprecationWarning in Column

hagenw:
In `pandas` 1.5.1, 1.5.2, or 1.5.3 they changed the emitted `UserWarning` we were already suppressing to a `DeprecationWarning` which I propose to also catch.


audformat/core/column.py
```diff
@@ -343,6 +343,12 @@ def set(
                 # as introduced at
                 # https://pandas.pydata.org/docs/dev/whatsnew/v1.5.0.html#inplace-operation-when-setting-values-with-loc-and-iloc
                 warnings.simplefilter(action='ignore', category=FutureWarning)
+                # Avoid DeprecationWarning
+                # for newer versions of pandas (<2.0.0)
```
frankenjoe, line 347: I don't really understand this comment :) What about newer versions >= 2.0.0.

hagenw: From version 2.0.0 of `pandas` it will always try to set values in place (what we are doing here anyway) and no longer raise a warning. For older version it might do something else if setting in place does not work.

I reformulated now the comment.

PR #367, opened 2023-04-12
DOC: add raises section to Database.attachments

### LINKED ISSUE #357: When assigning an attachment object to a database a RuntimeError can be raised
As mentioned in https://github.com/audeering/audformat/pull/356#discussion_r1150185783 we need to add to the docstring of `audformat.Database.__setitem__()` that a `RuntimeError` is raised if the path of the new attachment overlaps with the path of already assigned attachments.

hagenw:
Closes #357

This adds a raises section to the `audformat.Database.attachments` property. It feels a little bit strange to add it to a property, but I couldn't find another matching place to add this error that might happen when a new attachment is assigned to a database object, e.g. `db.attachments['file'] = audformat.Attachment('file.txt')`.

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -188,7 +188,13 @@ class Database(HeaderBase):
             value_type=Attachment,
             set_callback=self._set_attachment,
         )
-        r"""Dictionary of attachments"""
+        r"""Dictionary of attachments.
+
+        Raises:
+            RuntimeError: if the path of a newly assigned attachment
+                overlaps with the path of already assigned attachments
+
+        """
         self.media = HeaderDict(value_type=Media)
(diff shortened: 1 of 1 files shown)


audformat/core/database.py
```diff
@@ -188,7 +188,13 @@ def __init__(
             value_type=Attachment,
             set_callback=self._set_attachment,
         )
-        r"""Dictionary of attachments"""
+        r"""Dictionary of attachments.
+
+        Raises:
+            RuntimeError: if the path of a newly assigned attachment
+                overlaps with the path of already assigned attachments
```
frankenjoe, line 195:
```suggestion
                overlaps with the path of an existing attachment
```

PR #369, opened 2023-04-13
Add audformat.Scheme.labels_as_list

### LINKED ISSUE #301: Add helper function to access labels from Scheme
At the moment we have `Scheme.labels`, which is an attribute and has to contain a string value when a misc table is used as scheme.
This means there is no direct way to access the labels of this scheme.
Internally we have `Scheme._labels_to_list()` and `Scheme._labels_to_dict()`. Maybe we could create a single public method to access the labels in a desired fromat.

hagenw:
Closes #301 

This adds the `labels_as_list` property to `audformat.Scheme` to provide a way to the user to always get a list of all labels independent how they are stored in the scheme.

image

### DIFF
--- a/audformat/core/scheme.py
+++ b/audformat/core/scheme.py
@@ -184,10 +184,14 @@ class Scheme(common.HeaderBase):
 
     @property
     def labels_as_list(self) -> typing.List:
-        r"""Scheme labels as list
+        r"""Scheme labels as list.
+
+
+        If scheme does not define labels
+        an empty list is returned.
 
         Returns:
-            list of scheme labels
+            list of labels
 
         """
         if self.labels is None:


audformat/core/scheme.py
```diff
@@ -172,6 +172,21 @@ def __init__(
...
+
+        Returns:
+            ``True`` if data type is numeric
+
+        """
+        return self.dtype in (define.DataType.INTEGER, define.DataType.FLOAT)
+
+    @property
+    def labels_as_list(self) -> typing.List:
+        r"""Scheme labels as list"""
```
frankenjoe, line 187: Maybe extend the docstring with  `Returns: ...` as we do with the other properties?

audformat/core/scheme.py
```diff
@@ -172,6 +172,21 @@ def __init__(
...
+        return self._labels_to_list(self.labels)
```
frankenjoe, line 188: I think we need to handle the case where the scheme has no labels, e.g.:

```python
>>> audformat.Scheme().labels_as_list
TypeError: 'NoneType' object is not iterable
```

Maybe return an empty list in that case?

hagenw: Yes, good point.

I now return an empty list and also added a test for that case.

audformat/core/scheme.py
hagenw: I added the "Returns" section to the docstring.

audformat/core/scheme.py
```diff
@@ -172,6 +172,29 @@ def __init__(
...
+
+        Returns:
+            ``True`` if data type is numeric
+
+        """
+        return self.dtype in (define.DataType.INTEGER, define.DataType.FLOAT)
+
+    @property
+    def labels_as_list(self) -> typing.List:
+        r"""Scheme labels as list
```
frankenjoe, line 187:
```suggestion
        r"""Scheme labels as list.
```

PR #370, opened 2023-04-14
DOC: add example to utils.to_filewise_index()

### LINKED ISSUE #288: Add example to audformat.utils.to_filewise_index()
Currently, we don't have any example for this function as it requires existing files.

One possible usage for this function is getting VAD segments and splitting up existing files accordingly. Maybe we could create an example along those lines.

hagenw:
Closes #288 

This adds a docstring example to `audformat.utils.to_filewise_index()`.

image

### DIFF
--- a/audformat/core/conftest.py
+++ b/audformat/core/conftest.py
@@ -17,8 +17,6 @@ def prepare_docstring_tests(doctest_namespace):
 
         # audformat.utils.to_filewise_index()
         audiofile.write(audeer.path(tmp, 'f.wav'), np.ones((1, 8000)), 8000)
-        assert os.path.exists(audeer.path(tmp, 'f.wav'))
-        assert audiofile.duration(audeer.path(tmp, 'f.wav')) == 1.0
 
         current_dir = os.getcwd()
         os.chdir(tmp)
--- a/audformat/core/conftest.py
+++ b/audformat/core/conftest.py
@@ -15,11 +15,11 @@ def prepare_docstring_tests(doctest_namespace):
     with tempfile.TemporaryDirectory() as tmp:
         doctest_namespace['tmp'] = tmp
 
-        # audformat.utils.to_filewise_index()
-        audiofile.write(audeer.path(tmp, 'f.wav'), np.ones((1, 8000)), 8000)
-
         current_dir = os.getcwd()
         os.chdir(tmp)
+
+        # audformat.utils.to_filewise_index()
+        audiofile.write('f.wav', np.ones((1, 8000)), 8000)
 
         yield
 
--- a/audformat/core/conftest.py
+++ b/audformat/core/conftest.py
@@ -7,7 +7,7 @@ import pytest
 import audiofile
 
 
-@pytest.fixture(scope='session', autouse=True)
+@pytest.fixture(scope='package', autouse=True)
 def prepare_docstring_tests(doctest_namespace):
 
     # Prepare files and tmp folder needed in doctests
--- a/audformat/core/conftest.py
+++ b/audformat/core/conftest.py
@@ -4,7 +4,6 @@ import tempfile
 import numpy as np
 import pytest
 
-import audeer
 import audiofile
 
 
--- a/audformat/core/conftest.py
+++ b/audformat/core/conftest.py
@@ -1,3 +1,4 @@
+import os
 import tempfile
 
 import numpy as np
@@ -16,5 +17,12 @@ def prepare_docstring_tests(doctest_namespace):
 
         # audformat.utils.to_filewise_index()
         audiofile.write(audeer.path(tmp, 'f.wav'), np.ones((1, 8000)), 8000)
+        assert os.path.exists(audeer.path(tmp, 'f.wav'))
+        assert audiofile.duration(audeer.path(tmp, 'f.wav')) == 1.0
 
-    yield
+        current_dir = os.getcwd()
+        os.chdir(tmp)
+
+        yield
+
+        os.chdir(current_dir)
--- a/audformat/core/conftest.py
+++ b/audformat/core/conftest.py
@@ -1,4 +1,4 @@
-import os
+import tempfile
 
 import numpy as np
 import pytest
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -24,4 +24,9 @@ def create_audio_files():
         file_duration=pytest.FILE_DUR,
     )
     yield
-    shutil.rmtree(pytest.DB_ROOT)
+    audeer.rmdir(pytest.DB_ROOT)
+    # Clean up docstring generated files
```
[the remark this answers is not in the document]
frankenjoe: If we want we could also generate the audio file there to simply the docstring example.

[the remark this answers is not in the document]
hagenw: Good idea, I changed accordingly.

audformat/core/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+import audiofile
+
+
+@pytest.fixture(scope='session', autouse=True)
+def prepare_docstring_tests():
+
+    # Prepare WAV files needed in doctests
+    #
+    # audformat.utils.to_filewise_index()
+    audiofile.write('f.wav', np.ones((1, 8000)), 8000)
```
frankenjoe, line 16: Maybe use a temporary directory as in:

https://github.com/audeering/audbackend/blob/dev/audbackend/core/conftest.py#L10

?

frankenjoe: Then you don't have to care about removing files after `yield`.

hagenw: I tried, see https://github.com/audeering/audformat/pull/370/commits/a69365a2740de6d656c5d5c0f04002cf9078683e, but I don't know why it is not working.

hagenw: Also the example seems slightly easier to me without the `tmp` folder even though it is a much cleaner solution.

hagenw: I found the error, I forgot to indent `yield` as well.

frankenjoe: Yes, otherwise the audio file is removed before the example is executed :)

frankenjoe: That is strange, why does it now include the absolute path and not before?

frankenjoe: Maybe a bug in `to_filewise_index()`?

hagenw: No, it depends what you provide as `output_folder`:

image

frankenjoe: But your output folder was relative, wasn't it?

hagenw: No, I had to write `audeer.path(tmp, 'split')` to make sure the resulting files are stored in the tmp folder and automatically cleaned.
Without using the tmp folder, the path is relative.

frankenjoe: But the current working directory should be set to `tmp` in the docstring example.

hagenw: No, this is not automatically happening.
I can enforce it with `os.chdir(tmp)`, but then other tests fail (even when I change the directory back after `yield`).

hagenw: I will try to change those other failing tests, and see if it makes sense.

hagenw: Currently blocked by https://github.com/audeering/audformat/issues/371

hagenw: #371 is merged and I was now able to use a tmp folder to store any files.

We have to change folders during tests, but it doesn't seem to have any undesired consequences.

audformat/core/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+
+    # Prepare files and tmp folder needed in doctests
+    with tempfile.TemporaryDirectory() as tmp:
+        doctest_namespace['tmp'] = tmp
+
+        # audformat.utils.to_filewise_index()
+        audiofile.write(audeer.path(tmp, 'f.wav'), np.ones((1, 8000)), 8000)
+
+        current_dir = os.getcwd()
+        os.chdir(tmp)
```
frankenjoe, line 22: Could be simplified to:

```suggestion        
        current_dir = os.getcwd()
        os.chdir(tmp)

        # audformat.utils.to_filewise_index()
        audiofile.write('f.wav', np.ones((1, 8000)), 8000)
```

audformat/core/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+import tempfile
+
+import numpy as np
+import pytest
+
+import audeer
+import audiofile
+
+
+@pytest.fixture(scope='session', autouse=True)
```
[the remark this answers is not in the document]
frankenjoe: Mhh, but it seems to have the disadvantage that it runs it for every docstring example. I still wonder if it's possible to set it up so that it's only executed during the docstring examples or alternatively for specific docstring examples.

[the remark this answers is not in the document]
frankenjoe: Ah, setting it to `'package'` seems to do the trick.

audformat/core/utils.py
```diff
@@ -1399,6 +1400,16 @@ def to_filewise_index(
         ValueError: if ``output_folder`` contained in path to files of
             original data
 
+    Examples:
+        >>> index = segmented_index(
+        ...     files=['f.wav', 'f.wav'],
+        ...     starts=[0, 0.5],
+        ...     ends=[0.5, 1],
+        ... )
+        >>> # Split into single files
```
frankenjoe, line 1409: do we really need this comment?

hagenw: I thought as a reminder that this function does not only provide an index, but does also create files.

But I'm also fine with removing it.

frankenjoe: I think it's clear enough

audformat/core/conftest.py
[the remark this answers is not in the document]
hagenw: I changed to package, but for whatever reason the output of `audeer.path('a')` did not contain the tmp folder for `'session'` nor for `'package'`. So it seems in both cases we are not inside the tmp folder. Which is strange because before tests were failing exactly because of the switch to the tmp folder.

audformat/core/utils.py
hagenw: I removed it.

audformat/core/conftest.py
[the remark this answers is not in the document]
frankenjoe: Sorry, I'm lost. What do you mean by "output of audeer.path('a')"?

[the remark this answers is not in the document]
hagenw: Inside a test (`test_expand_file_path`) I added:

```python
print(audeer.path('a'))
assert False
```

in order to force it printing out in which folder the test is executed and to my surprise it wasn't in the tmp folder, but inside `/home/audeering.local/hwierstorf/git/audeering/audformat/a` in both cases.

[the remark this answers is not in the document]
hagenw: But it's anyway the behavior we would like to have, so maybe we don't need to worry any further here.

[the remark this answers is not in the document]
hagenw: When I add `audeer.path('a')` in the docstring I get something like `/tmp/tmpacgu3lml/a` so the tmp folder seems indeed only be used there.

PR #372, opened 2023-04-14
Support relative path in expand_file_path()

### LINKED ISSUE #371: audformat.utils.expand_file_path() might expand the provided root argument
In `audformat.utils.expand_file_path()` we state:

image

And provide the following example:

```python
>>> index = audformat.filewise_index(['f1', 'f2'])
>>> audformat.utils.expand_file_path(index, '/some/where')
Index(['/some/where/f1', '/some/where/f2'], dtype='string', name='file')
```

What is not documented is the behavior if `root` is not ab absolute dir.
I would expect that it would raise an error stating that `root` is not an absolute dir or bahve in the following way:
```python
>>> audformat.utils.expand_file_path(index, 'some/where')
Index(['some/where/f1', 'some/where/f2'], dtype='string', name='file')
```

What it does instead is:
```python
>>> audformat.utils.expand_file_path(index, 'some/where')
Index(['/home/audeering.local/hwierstorf/git/audeering/audformat/some/where/f1', '/home/audeering.local/hwierstorf/git/audeering/audformat/some/where/f2'], dtype='string', name='file')
```

I would consider this a bug.

hagenw:
Closes #371 

This extends `audformat.expand_file_path()` to support relative and absolute path in both arguments (`root` and `index`).

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1,6 +1,7 @@
 import collections
 import errno
 import os
+import platform
 import re
 import sys
 import typing as typing
@@ -23,6 +24,14 @@ from audformat.core.index import (
     segmented_index,
 )
 from audformat.core.scheme import Scheme
+
+
+# Exclude examples that return a path containing `\`
+# from doctest on Windows
+if platform.system() in ['Windows']:  # pragma: no cover
+    __doctest_skip__ = [
+        'expand_file_path',
+    ]
 
 
 def concat(
@@ -477,12 +486,14 @@ def expand_file_path(
     r"""Expand relative path in index with root.
 
     Args:
-        index: index with relative file path conform to
+        index: index with absolute or relative file path
+            conform to
             :ref:`table specifications <data-tables:Tables>`
-        root: directory added in front of the relative file path
+        root: relative or absolute path added in front
+            of the index file path
 
     Returns:
-        index with absolute file path
+        index with root added to file path
 
     Raises:
         ValueError: if index is not conform to
@@ -492,14 +503,19 @@ def expand_file_path(
         >>> index = filewise_index(['f1', 'f2'])
         >>> index
         Index(['f1', 'f2'], dtype='string', name='file')
-        >>> expand_file_path(index, '/some/where')  # doctest: +SKIP
+        >>> expand_file_path(index, '/some/where')
         Index(['/some/where/f1', '/some/where/f2'], dtype='string', name='file')
+        >>> index = filewise_index(['/b/f1', '/b/f2'])
+        >>> index
+        Index(['/b/f1', '/b/f2'], dtype='string', name='file')
+        >>> expand_file_path(index, './a')
+        Index(['a/b/f1', 'a/b/f2'], dtype='string', name='file')
 
     """  # noqa: E501
     if len(index) == 0:
         return index
 
-    root = audeer.path(root) + os.path.sep
+    root = os.path.normpath(root) + os.path.sep
 
     if is_segmented_index(index):
         index = index.set_levels(root + index.levels[0], level=0)
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -491,8 +491,7 @@ def expand_file_path(
     and puts it in front of the file path in the index.
 
     Args:
-        index: index with absolute or relative file path
-            conform to
+        index: index conform to
             :ref:`table specifications <data-tables:Tables>`
         root: relative or absolute path added in front
             of the index file path
@@ -507,8 +506,8 @@ def expand_file_path(
     Examples:
         >>> expand_file_path(filewise_index(['f1', 'f2']), '/a')
         Index(['/a/f1', '/a/f2'], dtype='string', name='file')
-        >>> expand_file_path(filewise_index(['/f1', '/f2']), './a')
-        Index(['a//f1', 'a//f2'], dtype='string', name='file')
+        >>> expand_file_path(filewise_index(['f1', 'f2']), './a')
+        Index(['a/f1', 'a/f2'], dtype='string', name='file')
 
     """  # noqa: E501
     if len(index) == 0:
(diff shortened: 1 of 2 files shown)


hagenw:
The tests currently fail as we have the problem that for speed reasons we do `index = root + index` for a filewise index,
but if we have an absolute path in `index` it will result in `//`, e.g. `root='a/'` and `index=audformat.filewise_index(['/b'])` would result in `'a//b'`.

@jwagner do you see a way we can fix this without relying on `pandas.Index.map()` which is slower?

hagenw:
> The tests currently fail as we have the problem that for speed reasons we do `index = root + index` for a filewise index, but if we have an absolute path in `index` it will result in `//`, e.g. `root='a/'` and `index=audformat.filewise_index(['/b'])` would result in `'a//b'`.
> 
> @jwagner do you see a way we can fix this without relying on `pandas.Index.map()` which is slower?

I now decided to simply stay with the two `//` we are getting when adding to a absolute path.

audformat/core/utils.py
```diff
@@ -474,32 +483,38 @@ def expand_file_path(
...
+    r"""Expand path in index with root.
+
+    It applies :func:`os.path.normpath`
+    to the provided path ``root``,
+    adds a file separator at its end
+    and puts it in front of the file path in the index.
 
     Args:
-        index: index with relative file path conform to
+        index: index with absolute or relative file path
```
frankenjoe, line 494: Feels a bit strange to even highlight that the file path can be absolute. I mean yes, it can be, but expanding in that case doesn't really make sense. So I think I would either stick to what we had before or just say `index conform to ...`

audformat/core/utils.py
```diff
@@ -474,32 +483,38 @@ def expand_file_path(
         index: pd.Index,
         root: str,
 ) -> pd.Index:
-    r"""Expand relative path in index with root.
+    r"""Expand path in index with root.
+
+    It applies :func:`os.path.normpath`
```
frankenjoe, line 488: Since the result of the operation can include invalid paths, maybe we should not apply any processing to `root` but simply append whatever the user provides?

audformat/core/utils.py
```diff
@@ -474,32 +483,38 @@ def expand_file_path(
...
 
     Examples:
-        >>> index = filewise_index(['f1', 'f2'])
-        >>> index
-        Index(['f1', 'f2'], dtype='string', name='file')
-        >>> expand_file_path(index, '/some/where')  # doctest: +SKIP
-        Index(['/some/where/f1', '/some/where/f2'], dtype='string', name='file')
+        >>> expand_file_path(filewise_index(['f1', 'f2']), '/a')
+        Index(['/a/f1', '/a/f2'], dtype='string', name='file')
+        >>> expand_file_path(filewise_index(['/f1', '/f2']), './a')
```
frankenjoe, line 510: I am not sure if we should include an example that results in an invalid path. Maybe instead add a note in the docstring that the result may contain invalid paths and we do not check for it.

frankenjoe:
> @jwagner do you see a way we can fix this without relying on pandas.Index.map() which is slower?

We could check if there are absolute paths in the index and in that case raise an error. Otherwise, as I already pointed out in my comments I would mention in the docstring I simply add `root` without further processing and state in the docstring that the function simply appends whatever string is provided.

audformat/core/utils.py
hagenw: I don't think a path containing `//` is invalid. But I agree that it is maybe not a good idea to include such an example.

audformat/core/utils.py
hagenw: One problem here is that this mean the user provides `root='./a'` and the index contains `f1` and we end up with
`./af1` which is not what we want. So I think doing some processing with `root` is fine as it also adds no extra processing costs.

audformat/core/utils.py
hagenw: I switched to `index conform to ...`

frankenjoe:
> The problem is that this would break current behavior and might result in path entries that contain `.` or `..` which we also not want.

But isn't it already a breaking change that we do no longer call `audeer.path(root)`?

audformat/core/utils.py
frankenjoe: You are right we should at least ensure root ends on `/`, so we anyway have to process it. So calling `normpath()` should be fine, too.

hagenw:
> > The problem is that this would break current behavior and might result in path entries that contain `.` or `..` which we also not want.
> 
> But isn't it already a breaking change that we do no longer call `audeer.path(root)`?

Only when `root` is a relative path. For an absolute path we are not introducing a breaking change here at the moment, but we would when not removing `.` and `..`.

frankenjoe:
> might result in path entries that contain . or ..

I can contain `../`, e.g.:

```python
index = audformat.filewise_index(['f1'])
audformat.utils.expand_file_path(index, '../sub')
```
```
Index(['../sub/f1'], dtype='string', name='file')
```

but I guess that's a wanted behavior if we want to support adding a relative path.

audformat/core/utils.py
hagenw: One problem with `normpath` might be:

> On Windows, it converts forward slashes to backward slashes.

I'm not sure if `audeer.path()` that we used before also is doing this?

hagenw: I think it is the same behavior as in `audeer.path()` we are calling `abspath`:

image

hagenw: Sorry, I was wrong here :(

We are calling `os.path.realpath(os.path.expanduser(path))` in `audeer.path()`.
So switching to `normpath()` does indeed change the behavior on Windows.

 Do we want this to happen?

hagenw: Or maybe not, honestly I don't have a clue how these functions behave under Windows.

frankenjoe: Me neither and I already merged.

hagenw: Don't worry I checked at https://github.com/audeering/audeer/pull/97
and `audeer.path()` does indeed change `/` to `\`: 

image

So we don't added any breaking change here.

PR #373, opened 2023-04-14
TST: Use temporary working directory

frankenjoe:
Applies the same technique we used in #370 to set a temporary working directory when executing tests under `/tests`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,11 @@ import pytest
 import audformat.testing
 
 
+pytest.DB = audformat.testing.create_db()
+pytest.DB_ROOT = 'db'
+pytest.FILE_DUR = pd.to_timedelta('1s')
+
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1,27 +1,39 @@
...
-        file_duration=pytest.FILE_DUR,
-    )
-    yield
-    shutil.rmtree(pytest.DB_ROOT)
+        os.chdir(current_dir)
+
+
+pytest.DB = audformat.testing.create_db()
+pytest.DB_ROOT = 'db'
+pytest.FILE_DUR = pd.to_timedelta('1s')
```
hagenw, line 39: I find it easier to read if the variables are defined before the fixtures.

frankenjoe: ok, swapped

PR #374, opened 2023-04-25
Release 1.0.0

hagenw:
As the handling of the attachments in `audb` is now finished and working (https://github.com/audeering/audb/milestone/1), we should be able to make the 1.0.0 release for `audformat`.

image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
-Version 1.0.0 (2023-04-17)
+Version 1.0.0 (2023-04-25)
 --------------------------
 
 * Added: ``audformat.Scheme.labels_as_list`` property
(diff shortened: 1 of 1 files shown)


frankenjoe:
There's still one PR in `audb` which seems to be related to attachments, see https://github.com/audeering/audb/pull/290. Should we still wait until it is merged and a version of `audb` was published or should we rather target 1.0.0 here and then directly use it with the next `audb` release?

hagenw:
The remaining pull request in `audb` does not require any updates to `audformat`.

I would propose we first make the new `audfromat` release. Afterwards we can test that `audb` still works and also make a new `audb` release. But it is not required to do it this way as I don't see a need to require a newer version of `audformat` in `audb` than 0.16.1.

frankenjoe:
> I don't see a need to require a newer version of audformat in audb than 0.16.1.

Yes, that's why I was wondering if it is really urgent to go 1.0.0 here or first release a new version of `audb`.

PR #377, opened 2023-05-04
Fix precision of start and end in segmented_index()

hagenw:
Relates to https://github.com/audeering/audinterface/issues/113

This ensures that the precision of the start and end values of a segmented index always end after 6 digits which should be sufficient for all possible audio sampling rates used in the near future.

It turned out that the solution proposed in https://github.com/audeering/audinterface/issues/113#issuecomment-1504741227 does not always work and I had to add some manual rounding as well.
I also added a test that was failing for the current master.

In addition to start and end values, it also fixes the precision for the stored duration of a database. If you prefer we can also move this to another pull request.

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -28,6 +28,7 @@ from audformat.core.errors import (
     BadIdError,
     TableExistsError,
 )
+from audformat.core.index import to_timedelta
 from audformat.core.media import Media
 from audformat.core.rater import Rater
 from audformat.core.scheme import Scheme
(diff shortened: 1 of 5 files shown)


frankenjoe:
Is this a breaking change?

hagenw:
> Is this a breaking change?

I don't think so. But I have to admit I'm not sure ;) I checked a few existing databases and I was not able to find those strange times as reported in https://github.com/audeering/audinterface/issues/113. On the other hand it was easy to reproduce them locally, so maybe it depends also on the `pandas` version?

hagenw:
So maybe we should check what happens if you load a CSV file which contains index start and end values that use more than 6 digits. I think it would be risky if we would convert those to different values.

hagenw:
Closing in favor of #380.

PR #380, opened 2023-05-05
Ensure to_segmented_index() uses high precision

### LINKED ISSUE #379: Get filewise table as segmented has lower precision
I expect that this issue also causes https://github.com/audeering/audinterface/issues/113

```python
import audb
import audformat


name = 'emodb'
version = '1.3.0'
media = 'wav/16b10Wb.wav'

deps = audb.dependencies(name, version=version)
db = audb.load(name, version=version, media=media, verbose=False)
```

Then we have
```python
>>> deps.duration(media)
2.5225
>>> df = db['emotion'].get(as_segmented=True, allow_nat=False)
>>> df.index.get_level_values('end')[-1]
Timedelta('0 days 00:00:02.522499')
>>> index = audformat.segmented_index(['f'], [0], [2.5225])
>>> index.get_level_values('end')[-1]
Timedelta('0 days 00:00:02.522499999')
```

The reason why `get(as_segmented=True, allow_nat=False)` returns a different result is caused by https://github.com/audeering/audformat/blob/6b0beb16cdcfdbe46a230da650bdc9688b8dd97c/audformat/core/utils.py#L1625
as can be seen by
```python
>>> index = audformat.segmented_index(['f', 'f'], [0, 0], [2.5225, np.NaN])
>>> ends = index.get_level_values('end')
>>> ends
TimedeltaIndex(['0 days 00:00:02.522499999', NaT], dtype='timedelta64[ns]', name='end', freq=None)
>>> ends[1] = pd.to_timedelta(2.5225, unit='s')
>>> ends
TimedeltaIndex(['0 days 00:00:02.522499999', '0 days 00:00:02.522499'], dtype='timedelta64[ns]', name='end', freq=None)
```

It could be solved by relying on `map()` instead:
```python
>>> def replace_nan(x):
...     if pd.isna(x):
...         x = pd.to_timedelta(2.5225, unit='s')
...     return x
>>> index.get_level_values('end').map(replace_nan)
TimedeltaIndex(['0 days 00:00:02.522499999', '0 days 00:00:02.522499999'], dtype='timedelta64[ns]', name='end', freq=None)
```

/cc @schruefer

hagenw:
Closes #379 

This implements a test that fails for the problem described in #379 with:

```
>       pd.testing.assert_index_equal(result, expected)                                                                                                                                        
E       AssertionError: MultiIndex level [2] are different                                                                                                                                     
E                                                                                                                                                                                              
E       MultiIndex level [2] values are different (100.0 %)    
E       [left]:  TimedeltaIndex(['0 days 00:00:02.522499'], dtype='timedelta64[ns]', name='end', freq=None)
E       [right]: TimedeltaIndex(['0 days 00:00:02.522499999'], dtype='timedelta64[ns]', name='end', freq=None)
```

and fixes the test by updating the implementation as outlined in #379.

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1630,7 +1630,7 @@ def to_segmented_index(
             # by the collected duration values
 
             # Reverse durs list as pop()
-            # will return last enrty first
+            # will return last entry first
             durs.reverse()
 
             # Add an additional entry to the list
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1589,7 +1589,7 @@ def to_segmented_index(
 
         if any(has_nat):
 
-            # First gather duration values
+            # Gather duration values
             # for all NaT end entries
 
             idx_nat = np.where(has_nat)[0]
@@ -1626,7 +1626,7 @@ def to_segmented_index(
                 task_description='Read duration',
             )
 
-            # Now replace all NaT entries in end
+            # Replace all NaT entries in end
             # by the collected duration values.
             # We have to convert ends to a series first
             # in order to preserve precision of duration values
@@ -1635,8 +1635,6 @@ def to_segmented_index(
             ends[idx_nat] = durs
 
             # Create a new index
-            # as index.set_levels() does not work
-            # if the level contains only NaT entries
             index = segmented_index(files, starts, ends)
 
     if isinstance(obj, pd.Index):
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1647,6 +1647,9 @@ def to_segmented_index(
                     x = pd.to_timedelta(durs.pop(), unit='s')
                 return x
 
+            # Create a new index
+            # as index.set_levels() doe snot work
+            # if the level contains only NaT entries
             index = segmented_index(files, starts, ends.map(replace_nan))
 
     if isinstance(obj, pd.Index):
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1648,7 +1648,7 @@ def to_segmented_index(
                 return x
 
             # Create a new index
-            # as index.set_levels() doe snot work
+            # as index.set_levels() does not work
             # if the level contains only NaT entries
             index = segmented_index(files, starts, ends.map(replace_nan))
 
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1627,7 +1627,9 @@ def to_segmented_index(
             )
 
             # Now replace all NaT entries in end
-            # by the collected duration values
+            # by the collected duration values.
+            # We have to convert ends to a series first
+            # in order to preserve precision of duration values
 
             ends = ends.to_series()
             ends[idx_nat] = durs
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1589,6 +1589,9 @@ def to_segmented_index(
 
         if any(has_nat):
 
+            # First gather duration values
+            # for all NaT end entries
+
             idx_nat = np.where(has_nat)[0]
             files = index.get_level_values(define.IndexField.FILE)
             starts = index.get_level_values(define.IndexField.START)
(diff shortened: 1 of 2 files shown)


audformat/core/utils.py
```diff
@@ -1622,9 +1625,29 @@ def job(file: str) -> pd.Timedelta:
...
+            durs.reverse()
+
+            # Add an additional entry to the list
+            # as the very first entry
+            # passed by map() is the whole index
+            # which will also trigger a call
+            # to durs.pop()
+            # if ends contains only a single entry
+            if len(ends) == 1:
+                durs.append(0)
```
hagenw, line 1643: This one is spooky and I don't know why it is needed, it's just empirical result.

In general I'm not very happy that we use `durs.pop()`, but I didn't found a different solution. The problem is that we somehow first collect all durations and then need to update the entries using `map()`. If `map()` would not only see the `end` values, but also the `file` values we could use a dictionary for the mapping to the duration values, but I'm not aware of how to make the `file` or the current index (like enumerate) visible inside `map()`.

hagenw:
This implementation also solves the problem described in https://github.com/audeering/audinterface/issues/113.
When executing the reported code it will return:
```
                                                                                      emotion  emotion.confidence
file                                              start  end                                                     
/home/audeering.local/hwierstorf/audb/emodb/1.... 0 days 0 days 00:00:01.898250     happiness                0.90
                                                         0 days 00:00:01.611250       neutral                1.00
                                                         0 days 00:00:01.877812500      anger                0.95
                                                         0 days 00:00:02.522499999      anger                1.00
                                                                                        mean       std
file                                              start  end                                          
/home/audeering.local/hwierstorf/audb/emodb/1.... 0 days 0 days 00:00:01.898250    -0.000311  0.082317
                                                         0 days 00:00:01.611250    -0.000312  0.125304
                                                         0 days 00:00:01.877812500 -0.000296  0.127394
                                                         0 days 00:00:02.522499999 -0.000464  0.095558
```

audformat/core/utils.py
```diff
@@ -1622,9 +1625,32 @@ def job(file: str) -> pd.Timedelta:
                 progress_bar=verbose,
                 task_description='Read duration',
             )
-            ends.values[idx_nat] = durs
```
frankenjoe, line 1625: Since the proposed solution looks slighly hacky I wonder if we should simply do instead:

```python
            ends = ends.to_series()
            ends[idx_nat] = durs
```

frankenjoe:
Cool, great we now figured out what caused the precision mismatch and found and easy fix for it.

So if I got it right this not only solves #379, but also https://github.com/audeering/audinterface/issues/113 and makes https://github.com/audeering/audinterface/pull/114 obsolete?

hagenw:
> Cool, great we now figured out what caused the precision mismatch and found and easy fix.
> 
> So if I got it right this not only solves #379, but also [audeering/audinterface#113](https://github.com/audeering/audinterface/issues/113) and makes [audeering/audinterface#114](https://github.com/audeering/audinterface/pull/114) obsolete?

Yes, exactly.
I would propose that we make a new release of `audformat` and afterwards we try to add a test to `audinterface` that fails for https://github.com/audeering/audinterface/issues/113, and is fixed when relying on the new `audformat` version.

audformat/core/utils.py
```diff
@@ -1622,8 +1625,18 @@ def job(file: str) -> pd.Timedelta:
...
+            # by the collected duration values.
+            # We have to convert ends to a series first
+            # in order to preserve precision of duration values
+
+            ends = ends.to_series()
+            ends[idx_nat] = durs
+
+            # Create a new index
+            # as index.set_levels() does not work
+            # if the level contains only NaT entries
```
frankenjoe, line 1639: When can this happen? I thought we just replaced all NaTs.

hagenw: You are right, this can no longer happen.
I tried to replace creating a new index with
```python
index = index.set_levels(ends, level=2)
```
but this fails for examples with files having the same duration with
```
ValueError: Level values must be unique: [Timedelta('0 days 00:00:01'), Timedelta('0 days 00:00:01')] on level 2
```

Should I stay with the current implementation, but just update the comment?

frankenjoe: As far as I know, with `set_levels()` you can replace the existing values, but you cannot have more values afterwards. But that's what we do here as we replace a single `NaT` entry with multiple different values. I guess simply removing the comment should be fine.

hagenw: Thanks, I also figured out why `set_levels()` does not work in the mean time, and as there is no `set_level_values()` that we could use instead I decided to stay with the current implementation and removed the comment

hagenw: The same problem was discussed at https://github.com/pandas-dev/pandas/issues/28294#issuecomment-528456052 and there it was also recommended to create a new index.

PR #381, opened 2023-05-08
Release 1.0.1

hagenw:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,12 +10,10 @@ and this project adheres to `Semantic Versioning`_.
 Version 1.0.1 (2023-05-08)
 --------------------------
 
-* Fixed: ensure the number of digits in ``end``
-  of a segmented index
-  is identical
-  when creating the index with ``audformat.segmented_index()``
-  or with the ``as_segmented``
-  argument of ``audformat.Table.get()``
+* Fixed: ensure ``audformat.Table.get()``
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,17 @@ The format is based on `Keep a Changelog`_,
...
 
+Version 1.0.1 (2023-05-08)
+--------------------------
+
+* Fixed: ensure the number of digits in ``end``
+  of a segmented index
+  is identical
+  when creating the index with ``audformat.segmented_index()``
+  or with the ``as_segmented``
+  argument of ``audformat.Table.get()``
```
frankenjoe, line 18:
```suggestion
* Fixed: ensure ``audformat.segmented_index()``
  uses same precision for ``end`` values
  as ``audformat.segmented_index()``
  and ``audformat.Table.get()`` with ``as_segmented=True``
```

hagenw: Changed to:

```
* Fixed: ensure ``audformat.Table.get()``
  with ``as_segmented=True``
  uses same precision for ``end`` values
  as ``audformat.segmented_index()``
```

frankenjoe: I guess we don't have to mention `utils.to_segmented_index()` here, right? Cause that's what I actually wanted to write, but then typed ``audformat.segmented_index()`` twice :)

PR #395, opened 2023-10-09
Add support for Python 3.12

hagenw:
Add tests and support for Python 3.12.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,8 @@ jobs:
           - os: ubuntu-latest
             python-version: '3.11'
           - os: ubuntu-latest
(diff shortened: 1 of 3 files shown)


hagenw:
The test for Python 3.12 worked, but it took 18x times as long as the other tests. I will restart the tests and see if this always happens.

hagenw:
It takes that long for Python 3.12 as it needs to build the wheel for `pandas` which takes a very long time.

We should only continue here as soon as `pandas` wheels for Python 3.12 are available.

hagenw:
We cannot support Python 3.12 until it is supported in `audb` (meaning `audbackend`) as we use this in docstring example

Issue #382, opened 2023-05-09
Switch to pyarrow engine when reading CSV files


hagenw:
As discussed in https://datapythonista.me/blog/pandas-20-and-the-arrow-revolution-part-i `pyarrow` can also be used to handle data types inside `pandas` and read CSV files.

I did a small test with a 304 MB CSV file containing 1,461,090 rows:

* reading with `pandas`: 3.99 s
* reading with `pyarrow`: 0.64 s

The most obvious difference is then that the resulting data types are called `int64[pyarrow]`, `string[pyarrow]`, and so on. There might also be other differences as it was stated in the article that not all operations are yet supported by the `pyarrow` data types, but maybe we are lucky and can use it already for the cases we have in `audformat`.

It might also well align with https://github.com/audeering/audformat/issues/321 and https://github.com/audeering/audformat/issues/376

<details><summary>Benchmark code</summary>

```python
import pandas as pd
import time

path = 'db.csv'
start = time.time()
df = pd.read_csv(path)
end = time.time()
print(f'{end - start:.2f} s')
start = time.time()
df = pd.read_csv(path, engine='pyarrow', dtype_backend='pyarrow')
end = time.time()
print(f'{end - start:.2f} s')
```

</details>

frankenjoe:
Impressive speed-up, so yes, we should definitely give it a try!

hagenw:
The problem is that it will be not trivial as the main speedup comes not from `engine='pyarrow'`, but from `dtype_backend='pyarrow'` which requires some under the hood work and might also have unwanted consequences for users as the returned dataframes might behave different than before.

PR #399, opened 2023-10-19
Add audformat.Database.get()

### LINKED ISSUE #398: Introduce Database.get() to access all datapoints for a scheme
When working with multiple databases you are often faced with the task of getting labels for a particular scheme, e.g. `gender`. But as those can be stored very differently it is very hard to write code that works for all databases.

I would propose to add `audformat.Database.get(schemes)`, that you can provide a single scheme or list of schemes and it will return all relevant data points in a series.

An example implementation is given by:

```python
import typing

import pandas as pd

import audb
import audeer
import audformat


def get(
        db: audformat.Database,
        schemes: typing.Union[str, typing.Sequence],
) -> pd.Series:
    r"""Get all data points for the requested scheme(s)."""

    requested_schemes = audeer.to_list(schemes)

    # Check if requested schemes
    # are stored as labels in other schemes
    scheme_mappings = []
    for scheme_id, scheme in db.schemes.items():

        if scheme.uses_table and scheme_id in db.misc_tables:
            # Labels stored as misc table
            for column_id, column in db[scheme_id].columns.items():
                for requested_scheme in requested_schemes:
                    if (
                            column.scheme_id is not None
                            and requested_scheme in column.scheme_id
                    ):
                        scheme_mappings.append((scheme_id, requested_scheme))
                        break
        else:
            # Labels stored in scheme
            for requested_scheme in requested_schemes:
                if scheme_id in scheme.labels_as_list:
                    scheme_mappings.append((scheme_id, requested_scheme))
                    break

        # Get data points for requested schemes
        ys = []
        for table_id, table in db.tables.items():
            for column_id, column in table.columns.items():
                if any_scheme_in_column(requested_schemes, column, column_id):
                    y = db[table_id][column_id].get()
                    ys.append(clean_y(y))
                else:
                    for (scheme_id, mapping) in scheme_mappings:
                        if scheme_id in column_id:
                            y = db[table_id][column_id].get(map=mapping)
                            ys.append(clean_y(y))

        index = audformat.utils.union([y.index for y in ys])
        y = audformat.utils.concat(ys).loc[index]
        y.name = ', '.join(requested_schemes)

    return y


def any_scheme_in_column(schemes, column, column_id):
    r"""Check if any of the schemes is attached to the column.

    At the moment we also check for the column name,
    as some databases might forget to attach the scheme.

    """
    return any(
        [
            scheme in column_id
            or (column.scheme_id is not None and scheme in column.scheme_id)
            for scheme in schemes
        ]
    )


def clean_y(y: pd.Series) -> pd.Series:
    r"""Remove NaN and normalize dtype of series."""
    # TODO: at the moment we simply convert to string
    # to avoid errors for different categorical data types.
    # In real implementation we need to adjust those dtypes.
    return y.dropna().astype('string')
```

We can test with:

```python
db = audb.load(
    'emodb',
    version='1.4.1',
    only_metadata=True,
    full_path=False,
    verbose=False,
)
y = get(db, ['gender', 'sex'])
print(y.head())
```

which returns

```
file
wav/03a01Fa.wav    male
wav/03a01Nc.wav    male
wav/03a01Wa.wav    male
wav/03a02Fc.wav    male
wav/03a02Nc.wav    male
Name: gender, sex, dtype: string
```

and

```python
y = get(db, 'age')
print(y.head())
```

which returns

```
file
wav/03a01Fa.wav    31
wav/03a01Nc.wav    31
wav/03a01Wa.wav    31
wav/03a02Fc.wav    31
wav/03a02Nc.wav    31
Name: age, dtype: string
```

What is not yet solved is the handling of multi-channel databases. E.g. if you have a two speaker conversation and gender labels for both speakers, `audformat.utils.concat(ys).loc[index]` will fail as you have the same index entry with different labels (or you have the same labels, then it would not fail, but you will have missing data).

We might also think about providing an argument to `get()` to restrict it to a given table type, or limiting it to certain samples by providing an `index` argument.

/cc @ChristianGeng @frankenjoe @felixbur @audeerington

hagenw:
Closes #398 

Adds `audformat.Database.get()` to request a data frame containing columns with  labels based on a scheme, limited by selected tables and/or splits and additional schemes drawn from the whole database.

To also support databases that did not used schemes, forgot to assign it to a column or use dictionaries to provide scheme labels, I added the `strict=False` argument.

For schemes that have a simple mapping,  e.g. transcriptions with `{'a0': 'some text'}` the label is expanded unless the user explicitly disables it with `map=False`.

For example:

```python
import audb
import audformat

db = audb.load('emodb', version='1.4.1', only_metadata=True, full_path=False)
db.get('emotion', ['gender'])
```
returns
```
                   emotion  gender
file                              
wav/03a01Fa.wav  happiness    male
wav/03a01Nc.wav    neutral    male
wav/03a01Wa.wav      anger    male
wav/03a02Fc.wav  happiness    male
wav/03a02Nc.wav    neutral    male
...                    ...     ...
wav/16b10Lb.wav    boredom  female
wav/16b10Tb.wav    sadness  female
wav/16b10Td.wav    sadness  female
wav/16b10Wa.wav      anger  female
wav/16b10Wb.wav      anger  female

[535 rows x 2 columns]

```

image

---

I also updated the docstring example for the `audformat.Database` class:

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -519,6 +519,18 @@ class Database(HeaderBase):
         Returns:
             data frame with values
 
+        Raises:
+            ValueError: if different labels are found
+                for a requested scheme under the same index entry
+            ValueError: if ``original_column_names`` is ``True``
+                and two columns in the returned data frame
+                have the same name
+                and cannot be joined
+                due to overlapping data
+                or different data type
+            TypeError: if labels of different data type are found
+                for a requested scheme
+
         Examples:
             Return all labels that match a requested scheme.
 
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -702,9 +702,9 @@ class Database(HeaderBase):
         additional_schemes = audeer.to_list(additional_schemes)
 
         if tables is None:
-            selected_tables = list(self.tables)
+            tables = list(self.tables)
         else:
-            selected_tables = audeer.to_list(tables)
+            tables = audeer.to_list(tables)
         if splits is not None:
             splits = audeer.to_list(splits)
 
@@ -733,7 +733,7 @@ class Database(HeaderBase):
 
         # --- Get data for requested schemes
         ys_requested_scheme = []
-        for table_id in selected_tables:
+        for table_id in tables:
 
             # Handle non-existing tables
             if table_id not in self.tables:
@@ -832,42 +832,29 @@ class Database(HeaderBase):
             obj = obj.to_frame()
 
         # Append additional schemes
-        if len(additional_schemes) > 0:
-
-            # Limit available index in database
-            if (
-                    (tables is not None or splits is not None)
-                    and len(obj) > 0
-            ):
-                for _, table in self.tables.items():
-                    table._df = table.get(
-                        utils.intersect([table.index, index]),
-                        copy=False,
-                    )
-
-            objs = [obj]
-            for scheme in additional_schemes:
-                if len(obj) == 0:
-                    obj = pd.DataFrame(
-                        {
-                            scheme: [],
-                        },
-                        index=filewise_index(),
-                        dtype='object',
-                    )
-                else:
-                    obj = self.get(
-                        scheme,
-                        strict=strict,
-                        original_column_names=original_column_names,
-                        aggregate_function=aggregate_function,
-                    )
-                objs.append(obj)
-            if len(objs) > 1:
-                obj = utils.concat(objs)
-                obj = obj.loc[index]
-                if len(obj) == 0:
-                    obj.index = filewise_index()
+        objs = [obj]
+        for scheme in additional_schemes:
+            if len(obj) == 0:
+                obj = pd.DataFrame(
+                    {
+                        scheme: [],
+                    },
+                    index=filewise_index(),
+                    dtype='object',
+                )
+            else:
+                obj = self.get(
+                    scheme,
+                    strict=strict,
+                    original_column_names=original_column_names,
+                    aggregate_function=aggregate_function,
+                )
+            objs.append(obj)
+        if len(objs) > 1:
+            obj = utils.concat(objs)
+            obj = obj.loc[index]
+            if len(obj) == 0:
+                obj.index = filewise_index()
 
         return obj
 
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -442,7 +442,8 @@ class Database(HeaderBase):
             *,
             tables: typing.Union[str, typing.Sequence] = None,
             splits: typing.Union[str, typing.Sequence] = None,
-            aggregate_function: typing.Callable[
+            aggregate_function: typing.Callable[pd.Series, typing.Any] = None,
+            modify_function: typing.Callable[
                 [pd.Series, 'Database', str, str],
                 pd.Series
             ] = None,
@@ -460,7 +461,7 @@ class Database(HeaderBase):
         If at least one returned label belongs to a segmented table,
         the returned data frame has a segmented index.
 
-        A custom ``aggregate_function`` can be provided
+        A custom ``modify_function`` can be provided
         that modifies how the labels are stored.
         The first argument of the function
         is a series ``y``
@@ -478,7 +479,7 @@ class Database(HeaderBase):
         stored in tables with the IDs
         ``'channel0'``
         and ``'channel1'``
-        one can provide an ``aggregate_function``
+        one can provide an ``modify_function``
         that adds the table name to the scheme name::
 
             def add_channel_name(y, db, table_id, column_id):
@@ -501,7 +502,17 @@ class Database(HeaderBase):
             schemes: scheme or sequence of scheme
             tables: limit returned samples to selected tables
             splits: limit returned samples to selected splits
-            aggregate_function: callable to handle collection
+            aggregate_function: function to aggregate overlapping values.
+                The function gets a :class:`pandas.Series`
+                with overlapping values
+                as input.
+                E.g. set to
+                ``lambda y: y.mean()``
+                to average the values
+                or to
+                ``tuple``
+                to return them as a tuple
+            modify_function: callable to handle collection
                 of scheme for certain tables and columns,
                 see discussion above
 
@@ -515,11 +526,11 @@ class Database(HeaderBase):
                 y,
                 table_id,
                 column_id,
-                aggregate_function,
+                modify_function,
         ):
             if y is not None:
-                if aggregate_function is not None:
-                    y = aggregate_function(y, self, table_id, column_id)
+                if modify_function is not None:
+                    y = modify_function(y, self, table_id, column_id)
                 ys.append(y)
 
         def scheme_in_column(scheme_id, column, column_id):
@@ -576,7 +587,7 @@ class Database(HeaderBase):
                             y,
                             table_id,
                             column_id,
-                            aggregate_function,
+                            modify_function,
                         )
                     # Get series based on label of scheme
                     else:
@@ -588,7 +599,7 @@ class Database(HeaderBase):
                                     y,
                                     table_id,
                                     column_id,
-                                    aggregate_function,
+                                    modify_function,
                                 )
 
             # Ensure we have a common dtype for requested scheme
@@ -623,7 +634,10 @@ class Database(HeaderBase):
             ys += ys_requested_scheme
 
         index = utils.union([y.index for y in ys])
-        obj = utils.concat(ys).loc[index]
+        obj = utils.concat(
+            ys,
+            aggregate_function=aggregate_function,
+        ).loc[index]
 
         if len(obj) == 0:
             obj = pd.DataFrame()
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -650,13 +650,19 @@ class Database(HeaderBase):
             wav/03a04Wc.wav      anger  boredom
             wav/03a05Aa.wav       fear  sadness
             ...
-            >>> db.get('emotion', aggregate_function=lambda y: y.mode())
-            ...
+            >>> db.get('emotion', aggregate_function=lambda y: y.mode()[0]).head()
+                               emotion
+            file
+            wav/03a01Fa.wav  happiness
+            wav/03a01Nc.wav    neutral
+            wav/03a01Wa.wav      anger
+            wav/03a02Fc.wav  happiness
+            wav/03a02Nc.wav    neutral
 
             Alternatively,
             use ``original_column_names`` to return column IDs.
 
-            >>> db.get('emotion', original_column_names=True)
+            >>> db.get('emotion', original_column_names=True).head()
                                emotion     random
             file
             wav/03a01Fa.wav  happiness  happiness
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -449,6 +449,7 @@ class Database(HeaderBase):
             tables: typing.Union[str, typing.Sequence] = None,
             splits: typing.Union[str, typing.Sequence] = None,
             strict: bool = False,
+            map: bool = True,
             original_column_names: bool = False,
             aggregate_function: typing.Callable[
                 [pd.Series],
@@ -498,6 +499,10 @@ class Database(HeaderBase):
                 the search is extended to columns
                 that match the name of the scheme
                 or the name of a label in the scheme
+            map: if ``True``
+                and a requested scheme
+                has labels with mappings,
+                those will be returned
             original_column_names: if ``True``
                 keep the original column names
                 (possibly results in multiple columns).
@@ -558,6 +563,22 @@ class Database(HeaderBase):
             wav/03a01Wa.wav      anger   male
             wav/03a02Fc.wav  happiness   male
             wav/03a02Nc.wav    neutral   male
+            >>> db.get('transcription').head()
+                                                    transcription
+            file
+            wav/03a01Fa.wav  Der Lappen liegt auf dem Eisschrank.
+            wav/03a01Nc.wav  Der Lappen liegt auf dem Eisschrank.
+            wav/03a01Wa.wav  Der Lappen liegt auf dem Eisschrank.
+            wav/03a02Fc.wav     Das will sie am Mittwoch abgeben.
+            wav/03a02Nc.wav     Das will sie am Mittwoch abgeben.
+            >>> db.get('transcription', map=False).head()
+                            transcription
+            file
+            wav/03a01Fa.wav           a01
+            wav/03a01Nc.wav           a01
+            wav/03a01Wa.wav           a01
+            wav/03a02Fc.wav           a02
+            wav/03a02Nc.wav           a02
 
             Non-existent schemes are ignored.
 
@@ -764,7 +785,20 @@ class Database(HeaderBase):
                 y = None
                 # Scheme directly stored in column
                 if scheme_in_column(requested_scheme, column, column_id):
-                    y = self[table_id][column_id].get()
+                    if requested_scheme in self.schemes:
+                        labels = self.schemes[requested_scheme].labels
+                    if (
+                            map is True
+                            and requested_scheme in self.schemes
+                            and isinstance(labels, dict)
+                            and not isinstance(list(labels.values())[0], dict)
+                            and self[table_id][column_id].scheme_id == requested_scheme
+                    ):
+                        # {'a': 'Full a', 'b': 'Full b'} is mapped
+                        # {'a': {'d': 1}, 'b': {'d': 2}} is not mapped
+                        y = self[table_id][column_id].get(map=requested_scheme)
+                    else:
+                        y = self[table_id][column_id].get()
                     y.name = requested_scheme
                     append_series(ys, y, column_id)
                 # Get series based on label of scheme
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -793,7 +793,9 @@ class Database(HeaderBase):
             for n, y in enumerate(ys_requested_scheme):
                 if not isinstance(y.dtype, pd.CategoricalDtype):
                     ys_requested_scheme[n] = y.astype(
-                        pd.CategoricalDtype(set(y.array.astype(dtype)))
+                        pd.CategoricalDtype(
+                            y.array.dropna().unique().astype(dtype)
+                        )
                     )
             # Find union of categorical data
             data = [y.array for y in ys_requested_scheme]
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -765,8 +765,11 @@ class Database(HeaderBase):
             ]
             if len(categorical_dtypes) > 0:
                 if len(set(dtypes_of_categories)) > 1:
-                    raise ValueError(
-                        'All categorical data must have the same dtype.'
+                    # Don't know if this can ever be triggered,
+                    # but make sure we raise the same error as
+                    # pd.api.types.union_categoricals
+                    raise TypeError(  # pragma: nocover
+                        'dtype of categories must be the same'
                     )
                 dtype = dtypes_of_categories[0]
                 # Convert everything to categorical data
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -454,6 +454,7 @@ class Database(HeaderBase):
                 [pd.Series],
                 typing.Any
             ] = None,
+            num_workers: int = 1,
     ) -> pd.DataFrame:
         r"""Get labels by scheme.
 
@@ -515,6 +516,10 @@ class Database(HeaderBase):
                 or to
                 ``tuple``
                 to return them as a tuple
+            num_workers: number of parallel jobs,
+                when retrieving ``additional_schemes``.
+                If ``None`` will be set to the number of processors
+                on the machine multiplied by 5
 
         Returns:
             data frame with values
@@ -673,18 +678,26 @@ class Database(HeaderBase):
 
         """  # noqa: E501
 
-        def append_series(
-                ys,
-                y,
-                table_id,
-                column_id,
-                label_id,
-        ):
+        def append_series(ys, y, column_id):
             if y is not None:
                 if original_column_names:
                     y.name = column_id
                 y = y[~y.index.duplicated(keep='first')]
                 ys.append(y)
+
+        def empty_frame(name):
+            return pd.DataFrame(
+                {name: []},
+                index=filewise_index(),
+                dtype='object',
+            )
+
+        def empty_series(name):
+            return pd.Series(
+                index=filewise_index(),
+                dtype='object',
+                name=name,
+            )
 
         def scheme_in_column(scheme_id, column, column_id):
             # Check if scheme_id
@@ -732,7 +745,7 @@ class Database(HeaderBase):
                     scheme_mappings.append((scheme_id, requested_scheme))
 
         # --- Get data for requested schemes
-        ys_requested_scheme = []
+        ys = []
         for table_id in tables:
 
             # Handle non-existing tables
@@ -751,36 +764,21 @@ class Database(HeaderBase):
                 if scheme_in_column(requested_scheme, column, column_id):
                     y = self[table_id][column_id].get()
                     y.name = requested_scheme
-                    append_series(
-                        ys_requested_scheme,
-                        y,
-                        table_id,
-                        column_id,
-                        None,
-                    )
+                    append_series(ys, y, column_id)
                 # Get series based on label of scheme
                 else:
                     for (scheme_id, mapping) in scheme_mappings:
                         if scheme_in_column(scheme_id, column, column_id):
                             if column.scheme_id is None:
-                                y = pd.Series(
-                                    index=filewise_index(),
-                                    dtype='object',
-                                )
+                                y = empty_series(requested_scheme)
                             else:
                                 y = self[table_id][column_id].get(map=mapping)
-                            y.name = requested_scheme
-                            append_series(
-                                ys_requested_scheme,
-                                y,
-                                table_id,
-                                column_id,
-                                mapping,
-                            )
+                                y.name = requested_scheme
+                            append_series(ys, y, column_id)
 
-        # Ensure we have a common dtype for requested scheme
+        # --- Ensure we have a common dtype for requested scheme
         categorical_dtypes = [
-            y.dtype for y in ys_requested_scheme
+            y.dtype for y in ys
             if isinstance(y.dtype, pd.CategoricalDtype)
         ]
         dtypes_of_categories = [
@@ -797,59 +795,47 @@ class Database(HeaderBase):
                 )
             dtype = dtypes_of_categories[0]
             # Convert everything to categorical data
-            for n, y in enumerate(ys_requested_scheme):
+            for n, y in enumerate(ys):
                 if not isinstance(y.dtype, pd.CategoricalDtype):
-                    ys_requested_scheme[n] = y.astype(
+                    ys[n] = y.astype(
                         pd.CategoricalDtype(
                             y.array.dropna().unique().astype(dtype)
                         )
                     )
             # Find union of categorical data
-            data = [y.array for y in ys_requested_scheme]
+            data = [y.array for y in ys]
             data = pd.api.types.union_categoricals(data)
-            ys_requested_scheme = [
-                y.astype(data.dtype)
-                for y in ys_requested_scheme
-            ]
+            ys = [y.astype(data.dtype) for y in ys]
 
-        index = utils.union([y.index for y in ys_requested_scheme])
+        # --- Combine all labels
+        index = utils.union([y.index for y in ys])
 
-        obj = utils.concat(
-            ys_requested_scheme,
-            aggregate_function=aggregate_function,
-        )
+        obj = utils.concat(ys, aggregate_function=aggregate_function)
         obj = obj.loc[index]
 
         if len(obj) == 0:
-            obj = pd.DataFrame(
-                {
-                    requested_scheme: [],
-                },
-                index=filewise_index(),
-                dtype='object',
-            )
+            obj = empty_frame(requested_scheme)
         elif isinstance(obj, pd.Series):
             obj = obj.to_frame()
 
-        # Append additional schemes
-        objs = [obj]
-        for scheme in additional_schemes:
-            if len(obj) == 0:
-                obj = pd.DataFrame(
-                    {
-                        scheme: [],
-                    },
-                    index=filewise_index(),
-                    dtype='object',
-                )
-            else:
-                obj = self.get(
-                    scheme,
-                    strict=strict,
-                    original_column_names=original_column_names,
-                    aggregate_function=aggregate_function,
-                )
-            objs.append(obj)
+        # --- Append additional schemes
+        def add_additional_scheme(scheme):
+            return self.get(
+                scheme,
+                strict=strict,
+                original_column_names=original_column_names,
+                aggregate_function=aggregate_function,
+            )
+
+        if len(obj) == 0:
+            objs = [empty_frame(scheme) for scheme in additional_schemes]
+        else:
+            objs = audeer.run_tasks(
+                add_additional_scheme,
+                params=[([scheme], {}) for scheme in additional_schemes],
+                num_workers=num_workers,
+            )
+        objs = [obj] + [obj for obj in objs if obj is not None]
         if len(objs) > 1:
             obj = utils.concat(objs)
             obj = obj.loc[index]
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -534,91 +534,91 @@ class Database(HeaderBase):
         Examples:
             Return all labels that match a requested scheme.
 
-            >>> db = Database('mydb')
-            >>> db.schemes['rating'] = Scheme('float')
-            >>> db.schemes['confidence'] = Scheme('float')
-            >>> db.splits['train'] = Split('train')
-            >>> db.splits['test'] = Split('test')
-            >>> db['run1'] = Table(filewise_index(['f1', 'f2']), split_id='train')
-            >>> db['run1']['rating'] = Column(scheme_id='rating')
-            >>> db['run1']['rating'].set([0.0, 0.9])
-            >>> db['run1']['confidence'] = Column(scheme_id='confidence')
-            >>> db['run1']['confidence'].set([1, 1])
-            >>> db['run2'] = Table(filewise_index(['f3']), split_id='test')
-            >>> db['run2']['rating'] = Column(scheme_id='rating')
-            >>> db['run2']['rating'].set([0.7])
-            >>> db['run2']['confidence'] = Column(scheme_id='confidence')
-            >>> db['run2']['confidence'].set([1])
-            >>> db.get('rating')
-                 rating
+            >>> import audb
+            >>> db = audb.load(
+            ...     'emodb',
+            ...     version='1.4.1',
+            ...     only_metadata=True,
+            ...     full_path=False,
+            ...     verbose=False
+            ... )
+            >>> db.get('emotion').head()
+                               emotion
             file
-            f1       0.0
-            f2       0.9
-            f3       0.7
-            >>> db.get('rating', ['confidence'])
-                 rating confidence
+            wav/03a01Fa.wav  happiness
+            wav/03a01Nc.wav    neutral
+            wav/03a01Wa.wav      anger
+            wav/03a02Fc.wav  happiness
+            wav/03a02Nc.wav    neutral
+            >>> db.get('emotion', ['gender']).head()
+                               emotion gender
             file
-            f1       0.0        1.0
-            f2       0.9        1.0
-            f3       0.7        1.0
+            wav/03a01Fa.wav  happiness   male
+            wav/03a01Nc.wav    neutral   male
+            wav/03a01Wa.wav      anger   male
+            wav/03a02Fc.wav  happiness   male
+            wav/03a02Nc.wav    neutral   male
 
             Non-existent schemes are ignored.
 
-            >>> db.get('rating', ['rater'])
-                 rating  rater
+            >>> db.get('emotion', ['non-existing']).head()
+                   emotion non-existing
             file
-            f1       0.0   NaN
-            f2       0.9   NaN
-            f3       0.7   NaN
+            wav/03a01Fa.wav  happiness          NaN
+            wav/03a01Nc.wav    neutral          NaN
+            wav/03a01Wa.wav      anger          NaN
+            wav/03a02Fc.wav  happiness          NaN
+            wav/03a02Nc.wav    neutral          NaN
 
             Limit to a particular table or split.
 
-            >>> db.get('rating', tables=['run1'])
-                 rating
+            >>> db.get('emotion', tables=['emotion.categories.train.gold_standard']).head()
+                               emotion
             file
-            f1       0.0
-            f2       0.9
-            >>> db.get('rating', splits=['test'])
-                 rating
+            wav/03a01Fa.wav  happiness
+            wav/03a01Nc.wav    neutral
+            wav/03a01Wa.wav      anger
+            wav/03a02Fc.wav  happiness
+            wav/03a02Nc.wav    neutral
+            >>> db.get('emotion', splits=['test']).head()
+                               emotion
             file
-            f3       0.7
+            wav/12a01Fb.wav  happiness
+            wav/12a01Lb.wav    boredom
+            wav/12a01Nb.wav    neutral
+            wav/12a01Wc.wav      anger
+            wav/12a02Ac.wav       fear
 
-            Return scheme name if column ID is different from scheme.
+            Return requested scheme name independent of column ID.
 
-            >>> db = Database('mydb')
-            >>> db.schemes['rating'] = Scheme('float')
-            >>> db['run1'] = Table(filewise_index(['f1', 'f2']))
-            >>> db['run1']['rater1'] = Column(scheme_id='rating')
-            >>> db['run1']['rater1'].set([0.0, 0.9])
-            >>> db['run2'] = Table(filewise_index(['f3']))
-            >>> db['run2']['rater1'] = Column(scheme_id='rating')
-            >>> db['run2']['rater1'].set([0.7])
-            >>> db.get('rating')
-                 rating
+            >>> db['emotion'].columns
+            emotion:
+              {scheme_id: emotion, rater_id: gold}
+            emotion.confidence:
+              {scheme_id: confidence, rater_id: gold}
+            >>> db.get('confidence').head()
+                             confidence
             file
-            f1       0.0
-            f2       0.9
-            f3       0.7
+            wav/03a01Fa.wav        0.90
+            wav/03a01Nc.wav        1.00
+            wav/03a01Wa.wav        0.95
+            wav/03a02Fc.wav        0.85
+            wav/03a02Nc.wav        1.00
 
             If ``strict`` is ``True``
             only values that have an attached scheme are returned.
 
-            >>> db = Database('mydb')
-            >>> speakers = {'s1': {'age': 23}, 's2': {'age': 25}}
-            >>> db.schemes['speaker'] = Scheme(labels=speakers)
-            >>> db['files'] = Table(filewise_index(['f1', 'f2']))
-            >>> db['files']['speaker'] = Column(scheme_id='speaker')
-            >>> db['files']['speaker'].set(['s1', 's2'])
-            >>> db['files']['rating'] = Column()
-            >>> db['files']['rating'].set([0.2, 0.7])
-            >>> db.get('rating', ['age'])
-                 rating age
+            >>> db.get('emotion.confidence').head()
+                             emotion.confidence
             file
-            f1       0.2 23
-            f2       0.7 25
-            >>> db.get('rating', ['age'], strict=True)
+            wav/03a01Fa.wav                0.90
+            wav/03a01Nc.wav                1.00
+            wav/03a01Wa.wav                0.95
+            wav/03a02Fc.wav                0.85
+            wav/03a02Nc.wav                1.00
+            >>> db.get('emotion.confidence', strict=True).head()
             Empty DataFrame
-            Columns: [rating, age]
+            Columns: [emotion.confidence]
             Index: []
 
             If more then one value exists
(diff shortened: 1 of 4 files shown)


frankenjoe:
> Handling of different data types

To me it sounds fine if we raise an error if we get different data types for the same scheme. Maybe we can explain in the error message to the user how she should fix the database. Or would you say it's a valid use-case to store gender values in one place as `['female', 'male']`, in another as `['f', 'm']` and somewhere else as `[0, 1]`? To me it sounds just fair to blame the database in that case.

> Loose definition of a scheme

Again, I am not sure if we should encourage users to store values that are semantically related in one place as a plain column (e.g. a column named 'gender' without scheme) and in another place using a scheme (e.g. in a speaker column that connects to a gender scheme). As before, I would say it's fine to either find values in an according column or through a scheme, but raise an error if we find both.

> Handling of different labels for same data point

This is indeed tricky. The current solution of spreading into several columns is indeed not too elegant. Once for the ordering you mentioned, but it also makes the result hard to predict as we do not know in advance how many columns will be returned. If we go for this solution I would at least use a multi-index for the column that at least `result['scheme']` will return all columns for `scheme` and not just the first. Or alternatively, combine the values into tuples.

---

A completely different approach would be to return an iterator so that the search results are returned one after another. This would leave it to the user how to deal with those issues :)

---

One general issue I see with the current solution is that it returns values from all splits. But in many use-cases you want the values of a specific split.

hagenw:
> One general issue I see with the current solution is that it returns values from all splits. But in many use-cases you want the values of a specific split.

There are the `tables` and/or `splits` arguments which you can use to get only a certain spit.
What might be suboptimal at the moment is that the data is first limited to a given split, which means if you want to to request a label from the test set + gender information, it will most likely not find any gender information.
So maybe we should change and limit only the returned results to the requested split.

hagenw:
> A completely different approach would be to return an iterator so that the search results are returned one after another. This would leave it to the user how to deal with those issues :)

I see the advantage of this as well, but it also has a big downside.
The main motivation to introduce `Database.get()` was for me that I can write code that gets a particular label
without having to know in which table it is stored, e.g. I know which test set I should use and just wanted to add gender information as well. On the other hand we also still have the problem that you get any number of columns back.
If we cannot solve this we might indeed switch to not return a data frame, but it wouldn't be my preferred solution for now.

> As before, I would say it's fine to either find values in an according column or through a scheme, but raise an error if we find both.

I think we need both solutions:
* One that never raises an error as you might want to parse a huge number of databases that are created by different authors and that you cannot easily fix. For the problem with different datatypes we might see if there is a way to automatically find a data type that can handle all the data (or in the worse case, simply convert everything to `object`).
* One that raises errors and recommends fixes. We might limit this to return schemes only, or we add another independent argument for it.

To summaries: add 1-2 arguments to handle how strict we with parsing.

hagenw:
## Handling of different labels for same data point

I think we cannot simply return a tuple inside a column that has more as one label as the information were the label comes from is lost, but most likely required for further processing.

One solution might be to use multi-index columns, but I'm not convinced by their usability in our case.

At the moment I would propose to expand the index of the data frame:
* always return a segmented index
* add `table` and `column` levels that store where the label are coming from

Example:

```python
import audformat

db = audformat.Database('db')
db.schemes['rating'] = audformat.Scheme('int')
index = audformat.filewise_index(['f1', 'f2'])
db['session1'] = audformat.Table(index)
db['session1']['rating'] = audformat.Column(scheme_id='rating')
db['session1']['rating'].set([0, 1])
db['session2'] = audformat.Table(index)
db['session2']['rating'] = audformat.Column(scheme_id='rating')
db['session2']['rating'].set([1, 1])
```

we will get

```python
>>> db.get('rating')
                                 rating
file start  end table    column        
f1   0 days NaT session1 rating       0
f2   0 days NaT session1 rating       1
f1   0 days NaT session2 rating       1
f2   0 days NaT session2 rating       1

```

The user would have then the possibility to count the number of available labels by using `len(df)`, and could check if there are duplicates with `any(df.index.droplevel([3, 4]).duplicated())` and if there are any duplicates the user could raise an error or provide code that can handle it, e.g. use only datapoints from the first session:

```python
df = df.reset_index([3, 4])
df = df[df.table == 'session1']
```

If there are no duplicates the index can be used as it is and should work with `audinterface`.

---

@frankenjoe I have not yet updated the tests and first wanted to get some feedback on this solution.

hagenw:
> But there seems indeed be a problem with the way it is currently implemented, e.g.:

I have now updated the handling of the `tables` and `splits` arguments and filter only at the very end:

```python
>>> db = audb.load('emodb', version='1.4.1', only_metadata=True, full_path=False)
>>> db.get(['emotion', 'gender'], splits=['test'])
                   emotion  gender
file                              
wav/12a01Fb.wav  happiness    male
wav/12a01Lb.wav    boredom    male
wav/12a01Nb.wav    neutral    male
wav/12a01Wc.wav      anger    male
wav/12a02Ac.wav       fear    male
...                    ...     ...
wav/16b10Lb.wav    boredom  female
wav/16b10Tb.wav    sadness  female
wav/16b10Td.wav    sadness  female
wav/16b10Wa.wav      anger  female
wav/16b10Wb.wav      anger  female

[231 rows x 2 columns]
```

frankenjoe:
> At the moment I would propose to expand the index of the data frame

Mhh, probably better than expanding into the columns. But it has the downside that a user will probably have to do some post-processing before it can be used e.g. to train a model. How about introducing an argument where the user can specify an aggregate function that is applied if we get more than one result for a file / segment?

hagenw:
> > At the moment I would propose to expand the index of the data frame
> 
> Mhh, probably better than expanding into the columns. But it has the downside that a user will probably have to do some post-processing before it can be used e.g. to train a model. How about introducing an argument where the user can specify an aggregate function that is applied if we get more than one result for a file / segment?

As a user would indeed need to add post-processing to handle the multiple labels anyway, it seems indeed reasonable to solve the issue by providing the custom code already as an aggregate function.
Thanks for the suggestion, I will try to implement it.

hagenw:
I implemented now a first version of the `aggregate_function` argument and it seems to be very helpful.
You can solve the issue of multiple labels with it,
but also the case where you have non-matching dtypes.
I have added several examples to the tests and to the docstring (see updated screenshot).

Here an easy example, that converts all labels to uppercase and adjusts the dtype accordingly:

```python
def upper(y, db, table_id, column_id):
    data = [v.upper() for v in y.values]
    dtype = pd.CategoricalDtype(categories=['FEMALE', 'MALE'], ordered=False)
    return pd.Series(data, index=y.index, name=y.name, dtype=dtype)

db = audb.load('emodb', version='1.4.1', only_metadata=True, full_path=False)
db.get('gender', aggregate_function=upper)
```

which returns

```
                 gender
file                   
wav/03a01Fa.wav    MALE
wav/03a01Nc.wav    MALE
wav/03a01Wa.wav    MALE
wav/03a02Fc.wav    MALE
wav/03a02Nc.wav    MALE
...                 ...
wav/16b10Lb.wav  FEMALE
wav/16b10Tb.wav  FEMALE
wav/16b10Td.wav  FEMALE
wav/16b10Wa.wav  FEMALE
wav/16b10Wb.wav  FEMALE

[535 rows x 1 columns]
```

@frankenjoe: if you agree with its implementation, I would update some of the error messages we raise and point to defining such a function, and maybe look for a shorter name of the argument?

frankenjoe:
> Here an easy example, that converts all labels to uppercase and adjusts the dtype accordingly:

Could you also provide a simple example where we have multiple values for the same file or segment? As far as I see, it now calls the aggregate function with the values from a specific column, I was expecting a function that gets called with a list of values that it found for a specific segment or file. But maybe that is already covered with your solution and I just don't see it :)

hagenw:
Sure, let's start with this database:

```python
import audformat

db = audformat.Database('db')
db.schemes['rating'] = audformat.Scheme('int')
index = audformat.filewise_index(['f1', 'f2'])
db['session1'] = audformat.Table(index)
db['session1']['rating'] = audformat.Column(scheme_id='rating')
db['session1']['rating'].set([0, 1])
db['session2'] = audformat.Table(index)
db['session2']['rating'] = audformat.Column(scheme_id='rating')
db['session2']['rating'].set([1, 1])
```

Without aggregate function you will now get an error (I would update the error message and point to the `aggregate_function` argument): 

```python
>>> db.get('rating')
...
ValueError: Found overlapping data in column 'rating':
      left  right
file             
f1       0      1
```

There are several possible solutions for this database:
1. select a particular session
2. add the table name to the column
3. calculate mean over the values

### 1.  select a particular session

```python
def select_session1(y, db, table_id, column_id):
    if table_id != 'session1':
        index = audformat.filewise_index()
        y = pd.Series(index=index, dtype=y.dtype, name=y.name)
    return y
```

```python
>>> db.get('rating', aggregate_function=select_session1)
      rating
file        
f1         0
f2         1
```

### 2. add the table name to the column

```python
def add_table_name(y, db, table_id, column_id):
    y.name = f'{y.name}-{table_id}'
    return y
```

```python
>>> db.get('rating', aggregate_function=add_table_name)
      rating-session1  rating-session2
file                                  
f1                  0                1
f2                  1                1
```

### 3. calculate mean over the values

This is indeed not so nicely supported.
You can do the following,
which of cause requires knowledge of the table names:

```python
def average_sessions(y, db, table_id, column_id):
    if table_id in ['session1', 'session2']:
        name = y.name
        if table_id == 'session1':
            y2 = db['session2'][column_id].get()
        else:
            y2 = db['session1'][column_id].get()
        y = pd.concat([y, y2], axis=1).mean(axis=1)
        y.name = name
    return y
```

```python
>>> db.get('rating', aggregate_function=average_sessions)
      rating
file        
f1       0.5
f2       1.0
```

---

The case 3, would be much easier with your suggested approach, but I find it also nice to have the possibility to change all labels as this also allows you to combine stuff (see `['gender', 'sex']` example in the tests), or to fix dtypes.

Don't know if there is a way to combine both approaches?

frankenjoe:
> Don't know if there is a way to combine both approaches?

I would use the aggregate function only to combine values that belong to the same file / segment. As you said case 3. is rather complicated and this was actually the main motivation to have an aggregate function.

To replace the aggregate function that you propose we could have an argument that `column` and `table` IDs are returned as additional columns in the result. Example 1 could then be solved by doing `df[df['column_id' == 'session1'])` and  Example 2 with `df['name'] += '-' + df['column_id']`.

hagenw:
>  Example 1 could then be solved by doing df[df['column_id' == 'session1'])

This will unfortunately not work as it cannot create `df` before as it will raise already an error as we have different labels for the same index. If we want to select a rating based on the table ID, this has to happen already inside the aggregate function. This was one of the big hurdles why I did not chose your suggested implementation, but just did a short cut that work on single columns.

But I will try if I manage to create an aggregate function, that has also access to all the table and column names of the label values.

frankenjoe:
> This will unfortunately not work as it cannot create df before as it will raise already an error as we have different labels for the same index

Wait a moment, but that is the error solved by the aggregation function, no?

hagenw:
> > This will unfortunately not work as it cannot create df before as it will raise already an error as we have different labels for the same index
> 
> Wait a moment, but that is the error solved by the aggregation function, no?

Yes, but if we aggregate the values first, e.g. by taking the mean, we cannot later on select the value for the first session.

frankenjoe:
Hehe, I think it's getting too complicated for me :)

hagenw:
One solution might be to combine both approaches:

* adding `aggregate_function` as argument to `audformat.utils.concat()` and `audformat.Database.get()` that is not aware of where the values come from, but just combine them to a single value
* adding `modifier_function` as argument to `audformat.Database.get()` which can modify values written from a column and has access to the column ID, table ID and database object

Maybe I will start and create another pull request independent of this one adding `aggregate_function` to `audformat.utils.concat()`?

hagenw:
I created https://github.com/audeering/audformat/pull/401 to first add `aggregate_function` to `audformat.utils.concat()`.

hagenw:
I now added the following three arguments:

* `aggregate_function` that is simply passed on to `audformat.utils.concat()` to combine several values (e.g. by averaging them)
* `modify_function` that provides a way to select only certain parts of the matching values and/or to modifiy them, e.g. based on table ID, column ID and other entries from the database object.
* `strict` to disable the matching of column names and scheme label dictionary keys

I also added some examples to the docstring show casing what you can do it with it.

---

There is another question: with the current implementation we return the name of the column, or several columns if they have different names for a matching scheme:

```python
>>> db = Database('mydb')
>>> db.schemes['rating'] = Scheme('float')
>>> db['run1'] = Table(filewise_index(['f1', 'f2']))
>>> db['run1']['rater1'] = Column(scheme_id='rating')
>>> db['run1']['rater1'].set([0.0, 0.9])
>>> db['run2'] = Table(filewise_index(['f3']))
>>> db['run2']['rater1'] = Column(scheme_id='rating')
>>> db['run2']['rater1'].set([0.7])
>>> db.get('rating')
     rater1
file
f1       0.0
f2       0.9
f3       0.7

>>> db = Database('mydb')
>>> db.schemes['rating'] = Scheme('float')
>>> db['run1'] = Table(filewise_index(['f1', 'f2']))
>>> db['run1']['rater1'] = Column(scheme_id='rating')
>>> db['run1']['rater1'].set([0.0, 0.9])
>>> db['run1']['rater2'] = Column(scheme_id='rating')
>>> db['run1']['rater2'].set([0.2, 0.7])
>>> db.get('rating')
     rater1   rater2
file
f1       0.0      0.2
f2       0.9      0.7
```

This might be not a good idea as a user doesn't know how many columns will be returned and how they are named.
So maybe it is better as default to name the returned columns by the scheme and require an `aggregate_function` or `modify_function` to handle the case with different columns?

E.g. to re-enable the current behavior, the user could write a simple modify function:
```python
def return_column_names(y, db, table_id, column_id):
    y.name = column_id
    return y
```

frankenjoe:
> This might be not a good idea as a user doesn't know how many columns will be returned and how they are named.
So maybe it is better as default to name the returned columns by the scheme and require an aggregate_function or modify_function to handle the case with different columns?

Yes, I also find it problematic that you can request `rating` but it returns a column `raters1` like in your first example or even two columns as in your second example. In the second example my expectation was rather you the user has to provide a aggregate function. And in both cases, the name of the column should be `rating` I think.

frankenjoe:
I am also not too convinced about the `modify_function` yet. It seems that this function only makes sense if you already have knowledge about the table and column names in the database. But to me the main use-case of the `get()` function is that you don't need to know these details. Maybe we should split up into two functions? One function that implements the original idea of `get()` as a shortcut for quickly accessing values of a specific split. And a second one that works more like an iterator for matching tables / columns with the option to modify the returned values?

hagenw:
> > This might be not a good idea as a user doesn't know how many columns will be returned and how they are named.
> > So maybe it is better as default to name the returned columns by the scheme and require an aggregate_function or modify_function to handle the case with different columns?
> 
> Yes, I also find it problematic that you can request `rating` but it returns a column `raters1` like in your first example or even two columns as in your second example. In the second example my expectation was rather you the user has to provide a aggregate function. And in both cases, the name of the column should be `rating` I think.

I changed the behavior and it now always returns the requested schemes as column names.
Two cases are still missing and on my TODO:
* a returned empty data frame should also contain the column names of the requested schemes
* if some of the schemes is missing it should be included as an empty column in the data frame and not dropped.

hagenw:
> I wonder if we should think more about the actual use-case for the function. At the moment it seems very powerful, but also hard to predict what the result will be.
> 
> I guess one of the main use-cases will be to retrieve gold standard labels for a specific split. E.g.
> 
> ```python
> import audformat
> 
> db = audformat.Database('mydb')
> db.schemes['emotion'] = audformat.Scheme('float')
> db.splits['test'] = audformat.Split('test')
> db['test.gold'] = audformat.Table(
>     audformat.filewise_index(['f1', 'f2']),
>     split_id='test',
> )
> db['test.gold']['emotion'] = audformat.Column(scheme_id='emotion')
> db['test.gold']['emotion'].set([1.0, 1.0])
> db.get('emotion', splits='test')
> ```
> 
> ```
>       emotion
> file         
> f1        1.0
> f2        1.0
> ```
> 
> This is really nice, since `get()` allows us to retrieve the values without exact knowledge of the table and column names.
> 
> But let's assume we have also stored self report values somewhere in the database:
> 
> ```python
> db['self-report'] = audformat.Table(audformat.filewise_index(['f1', 'f2']))
> db['self-report']['emotion'] = audformat.Column(scheme_id='emotion')
> db['self-report']['emotion'].set([0.0, 0.0])
> db.get('emotion', splits='test', aggregate_function=np.mean)
> ```
> 
> ```
>       emotion
> file         
> f1        0.5
> f2        0.5
> ```
> 
> We can see that those mess up the result, even though the table with self-report values is not even assigned to the test split. So I think at the moment it's hard for a user to predict what the function will return.

Yes, there are several cases where we might have problems. E.g. database contains several test splits or different values reported under the same scheme (as in your example).
I don't think there is an easy solution for this problem, besides introducing restrictions during creation of the database.

At the moment, I would say that you should use `aggregate_function` only in the case you know exactly what will happen. E.g. if I'm only interested in counting the available test set files with emotion labels, your call to the function is still valid. If you want to use the labels for testing, there should be no need to average them.

hagenw:
> I am also not too convinced about the `modify_function` yet. It seems that this function only makes sense if you already have knowledge about the table and column names in the database. But to me the main use-case of the `get()` function is that you don't need to know these details. Maybe we should split up into two functions? One function that implements the original idea of `get()` as a shortcut for quickly accessing values of a specific split. 

It's true that you need to have knowledge about the databases and that the `modify_function` is also very complex.
I think there might still be use cases for it, e.g. if you have a couple of databases that all share a common structure, but still have slightly different table or column names.

> And a second one that works more like an iterator for matching tables / columns with the option to modify the returned values?

As long as we think we will need a `modify_function`, `Database.get()` seems still to be the best place to add it too instead of replicating all its behavior again.

hagenw:
As far as I see it there are only two use cases where `modify_function` is needed at the moment:
* to have the option to return a data frame containing the original column (and mapping) names instead of aggregating the values
* to have the possibility to fix non matching data types

For both cases there is an example in the tests (without considering mapping), e.g.

```python
def column_names(y, db, table_id, column_id, label_id):
    y.name = column_id
    return y
```

```python
def rename_sex_to_gender(y, *args):
    if y.name == 'sex':
        y.name = 'gender'
        # Make sure it uses the correct dtype as well
        dtype = pd.CategoricalDtype(
            categories=['female', '', 'male'],
            ordered=False,
        )
        y = y.astype(dtype)
    return y
```

All other modify functions in the tests could be replaced to some post-processing after calling `get()`.
So maybe we can replace those two cases with two arguments instead?

frankenjoe:
> I changed the behavior and it now always returns the requested schemes as column names.

Ok, I think that makes a lot of sense. 

There is still the issue that values from a split table can be messed up by values from a non-split table:

```python
import audformat

db = audformat.Database('mydb')
db.schemes['emotion'] = audformat.Scheme('float')
db.splits['test'] = audformat.Split('test')
db['rating'] = audformat.Table(
    split_id='test',
)
db['rating']['rater1'] = audformat.Column(scheme_id='emotion')
db['rating']['rater1'].set([1.0, 1.0])
db['rating']['rater2'] = audformat.Column(scheme_id='emotion')
db['rating']['rater2'].set([1.0, 1.0])
db.get('emotion', splits='test', aggregate_function=np.mean)
```
```
      emotion
file         
f1        1.0
f2        1.0
```

but:


```python
db['self-report'] = audformat.Table(audformat.filewise_index(['f1', 'f2']))
db['self-report']['emotion'] = audformat.Column(scheme_id='emotion')
db['self-report']['emotion'].set([0.0, 0.0])

db.get('emotion', splits='test', aggregate_function=np.mean)
```
```
       emotion
file          
f1    0.666667
f2    0.666667
```

In this example the problem is that you have to provide an `aggegrate_function`. I guess you would suggest now that the user should write a `modify_function` in addition. But somehow it feels weird since the user explicitly asked for values from a specific split, but now has to manually exclude non-split tables. I guess the problem is that at some point you had to change the implementation to first retrieve all values and afterwards filter by split. I wonder if there is still a possibility to resolve the issue.

hagenw:
> I guess the problem is that at some point you had to change the implementation to first retrieve all values and afterwards filter by split. I wonder if there is still a possibility to resolve the issue.

Yes, because otherwise you cannot add values from other tables, e.g. `gender` stored in a non-split table, but you want `db.get(['rating', 'gender'], splits='test)`.

hagenw:
One solution might be to limit again at the beginning and using the following workflow to combine different schemes:

```python
import audformat

db = audformat.Database('mydb')
db.schemes['emotion'] = audformat.Scheme('float')
db.schemes['gender'] = audformat.Scheme('str')
db.splits['test'] = audformat.Split('test')
db['rating'] = audformat.Table(index[2:], split_id='test')
db['rating']['rater1'] = audformat.Column(scheme_id='emotion')
db['rating']['rater1'].set([1.0])
db['rating']['rater2'] = audformat.Column(scheme_id='emotion')
db['rating']['rater2'].set([1.0])
db['files'] = audformat.Table(index)
db['files']['gender'] = audformat.Column(scheme_id='gender')
db['files']['gender'].set(['female', 'male', 'female'])
```
And then
```python
df = db.get('emotion', splits='test')
df_gender = db.get('gender')
audformat.utils.concat([df, df_gender.loc[df.index]])
```
which returns
```
      emotion  gender
file                 
f3        1.0  female
```

---

The downside is that we had this behavior before, and the first thing you did was reported an error as some labels were not showing up. So I guess a user might be confused that the gender label cannot be requested for the test scheme here.

hagenw:
Here an example, for which a user might also not get the labels she might expect:

```python
import audformat

db = audformat.Database('mydb')
db.schemes['emotion'] = audformat.Scheme('float')
db.splits['test'] = audformat.Split('test')
db['test'] = audformat.Table(index[1:], split_id='test')
db['test']['emotion'] = audformat.Column(scheme_id='emotion')
db['test']['emotion'].set([0.6, 0.7])
db['test-unbalanced'] = audformat.Table(index, split_id='test')
db['test-unbalanced']['emotion'] = audformat.Column(scheme_id='emotion')
db['test-unbalanced']['emotion'].set([0.4, 0.6, 0.7])
```

Requesting the `test` split will not result in an error, but return the unbalanced test set:

```python
>>> db.get('emotion', splits='test')
      emotion
file         
f2        0.6
f3        0.7
f1        0.4
```

frankenjoe:
Maybe we should split the `schemes` argument into `scheme` and `additional_schemes`? For `scheme` we only look in the specified splits and the other ones we then grab from all tables.

The example above would then change to `db.get('rating', additonal_schemes=['gender'], splits='test)`

But I am not sure if this makes any sense :)

hagenw:
> The example above would then change to db.get('rating', additonal_schemes=['gender'], splits='test)
>
> But I am not sure if this makes any sense :)

I thought about the same and it might indeed be a better solution as it would allow to use `tables` and `splits` to first limit the search for `rating` and then add whatever other metadata a user might want to add.

hagenw:
> Maybe we should split the `schemes` argument into `scheme` and `additional_schemes`? For `scheme` we only look in the specified splits and the other ones we then grab from all tables.
> 
> The example above would then change to `db.get('rating', additonal_schemes=['gender'], splits='test)`
> 
> But I am not sure if this makes any sense :)

If have now implemented this by restricting the first argument `scheme` to requesting a single scheme and adding the `additonal_schemes` argument. The search for the labels requested with `scheme` can be limited using the `tables` and `splits` arguments, which will have no influence on the labels searched for with `additional_schemes`.

Coming back to your example from above:

```python
import audformat
import numpy as np

db = audformat.Database('mydb')
db.schemes['emotion'] = audformat.Scheme('float')
db.splits['test'] = audformat.Split('test')
db['rating'] = audformat.Table(
    split_id='test',
)
db['rating']['rater1'] = audformat.Column(scheme_id='emotion')
db['rating']['rater1'].set([1.0, 1.0])
db['rating']['rater2'] = audformat.Column(scheme_id='emotion')
db['rating']['rater2'].set([1.0, 1.0])
db['self-report'] = audformat.Table(audformat.filewise_index(['f1', 'f2']))
db['self-report']['emotion'] = audformat.Column(scheme_id='emotion')
db['self-report']['emotion'].set([0.0, 0.0])
```
Here we no get:

```python
>>> db.get('emotion', splits='test', aggregate_function=np.mean)
      emotion
file         
f1        1.0
f2        1.0
```

hagenw:
I have further removed the `modify_function` argument as it is no longer needed. In addition to `aggregate_function` I just added `original_column_names` to provide the option to return the original columns instead:

```python
import audformat
db = audformat.Database('mydb')
db.schemes['rating'] = audformat.Scheme('float')
db['run1'] = audformat.Table(audformat.filewise_index(['f1', 'f2']))
db['run1']['rater1'] = audformat.Column(scheme_id='rating')
db['run1']['rater1'].set([0.0, 0.9])
db['run1']['rater2'] = audformat.Column(scheme_id='rating')
db['run1']['rater2'].set([0.2, 0.7])
db.get('rating', original_column_names=True)
```

```
      rater1  rater2
file                
f1       0.0     0.2
f2       0.9     0.7
```

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+        schemes of the database,
+        but also columns with the same name
+        or labels of a scheme
+        with the requested name as key.
+        If at least one returned label belongs to a segmented table,
+        the returned data frame has a segmented index.
+        An ``aggregate_function`` can be provided
+        that specifies how values are combined
+        if more than one value is returned
+        per index entry.
```
frankenjoe, line 478:
```suggestion
        if more than one value is found
        for the same file or segment.
```

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+            aggregate_function: typing.Callable[
+                [pd.Series],
+                typing.Any
+            ] = None,
+    ) -> pd.DataFrame:
+        r"""Get labels by scheme.
+
+        Return all labels
+        that match the requested ``scheme``.
+        The request can be limited to selected ``tables``
```
frankenjoe, line 462:
```suggestion
        The request can be limited to specific ``tables``
```

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+        and/or ``splits``
+        and annotated by ``additional_schemes``
+        drawn from the whole database.
```
frankenjoe, line 465:
```suggestion
        and/or ``splits``.
        By providing ``additional_schemes``
        the result can be enriched
        with labels from other schemes
        (searched in all tables).
```

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+            original_column_names: bool = False,
+            aggregate_function: typing.Callable[
+                [pd.Series],
+                typing.Any
+            ] = None,
+    ) -> pd.DataFrame:
+        r"""Get labels by scheme.
+
+        Return all labels
+        that match the requested ``scheme``.
```
frankenjoe, line 461:
```suggestion
        Return all labels
        from columns assigned to a 
        :class:`audformat.Scheme`
        with name ``scheme``.
```

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+            aggregate_function: callable to aggregate overlapping values.
+                The function gets a :class:`pandas.Series`
+                with overlapping values
+                as input.
+                E.g. set to
+                ``lambda y: y.mean()``
+                to average the values
+                or to
+                ``tuple``
+                to return them as a tuple
```
frankenjoe, line 514: We could also add a reference to `audformat.utils.concat()` instead.

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+        An ``aggregate_function`` can be provided
+        that specifies how values are combined
+        if more than one value is returned
+        per index entry.
+
+        Args:
+            scheme: scheme ID for which labels should be returned.
+                The search for the labels
+                can be limited
+                by the ``tables`` and ``splits`` arguments
```
frankenjoe, line 484:
```suggestion
            scheme: scheme ID for which labels should be returned.
                The search can be restricted to specific tables and splits
                by the ``tables`` and ``splits`` arguments.
                Or extended to columns with that same name
                or the name of a label in the scheme
                using the `strict` argument
```

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+                ``tuple``
+                to return them as a tuple
+
+        Returns:
+            data frame with values
+
+        Examples:
+            Return all labels that match a requested scheme.
+
+            >>> db = Database('mydb')
```
frankenjoe, line 522: The following examples are quite difficult to grasp and seem a bit lengthy for the Examples section. I wonder if we should replace it with a simple `emodb` example and instead move them to a dedicated Usage section that provides some more information?

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+                by the ``tables`` and ``splits`` arguments
+            tables: limit search for ``scheme`` to selected tables
+            splits: limit search for ``scheme`` to selected splits
+            strict: if ``False``
+                not only values with the associated scheme(s)
+                are returned,
+                but also values
+                from columns with a column ID matchting the scheme
+                and values from scheme labels,
+                for matching dictionary keys
```
frankenjoe, line 497:
```suggestion
                the search is extended to columns
                that match the name of the scheme
                or the name of a label in the scheme
```

audformat/core/database.py
```diff
@@ -433,6 +441,385 @@ def duration(file: str) -> pd.Timedelta:
...
+            original_column_names: if ``True``
+                return the requested labels
+                under their original column names.
+                For mapped schemes,
+                the column name before mapping is returned,
+                e.g. when requesting ``'gender'``
+                it might return a column named ``'speaker'``
```
frankenjoe, line 504:
```suggestion
                keep the original column names
                (possibly results in multiple columns).
                For mapped schemes,
                the column name before mapping is returned,
                e.g. when requesting ``'gender'``
                it might return a column named ``'speaker'``
```

audformat/core/database.py
hagenw: I switched to `emodb` for using the examples, but one of the examples results in an error, that I do not understand yet:

```python
import audb
import audformat

db = audb.load('emodb', version='1.4.1', only_metadata=True, full_path=False)
db['emotion']['random'] = audformat.Column(scheme_id='emotion')
db['emotion']['random'].set(db['emotion']['emotion'].get().sample(frac=1, random_state=1))
db.get('emotion', aggregate_function=lambda y: y.mode())
```
results in
```
ValueError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 db.get('emotion', aggregate_function=lambda y: y.mode())

File ~/git/audeering/audformat/audformat/core/database.py:805, in Database.get(self, scheme, additional_schemes, tables, splits, strict, original_column_names, aggregate_function)
    798     ys_requested_scheme = [
    799         y.astype(data.dtype)
    800         for y in ys_requested_scheme
    801     ]
    803 index = utils.union([y.index for y in ys_requested_scheme])
--> 805 obj = utils.concat(
    806     ys_requested_scheme,
    807     aggregate_function=aggregate_function,
    808 )
    809 obj = obj.loc[index]
    811 if len(obj) == 0:

File ~/git/audeering/audformat/audformat/core/utils.py:361, in concat(objs, overwrite, aggregate_function)
    359         print(f'{y=}')
    360         print(f'{y.index=}')
--> 361         columns_reindex[column].loc[y.index] = y
    363 # Use `None` to force `{}` return the correct index, see
    364 # https://github.com/pandas-dev/pandas/issues/52404
    365 df = pd.DataFrame(columns_reindex or None, index=index)

File ~/.envs/analysis/lib/python3.8/site-packages/pandas/core/indexing.py:818, in _LocationIndexer.__setitem__(self, key, value)
    815 self._has_valid_setitem_indexer(key)
    817 iloc = self if self.name == "iloc" else self.obj.iloc
--> 818 iloc._setitem_with_indexer(indexer, value, self.name)

File ~/.envs/analysis/lib/python3.8/site-packages/pandas/core/indexing.py:1797, in _iLocIndexer._setitem_with_indexer(self, indexer, value, name)
   1795     self._setitem_with_indexer_split_path(indexer, value, name)
   1796 else:
-> 1797     self._setitem_single_block(indexer, value, name)

File ~/.envs/analysis/lib/python3.8/site-packages/pandas/core/indexing.py:2072, in _iLocIndexer._setitem_single_block(self, indexer, value, name)
   2069     value = self._align_series(indexer, Series(value))
   2071 elif isinstance(value, ABCDataFrame) and name != "iloc":
-> 2072     value = self._align_frame(indexer, value)
   2074 # check for chained assignment
   2075 self.obj._check_is_chained_assignment_possible()

File ~/.envs/analysis/lib/python3.8/site-packages/pandas/core/indexing.py:2358, in _iLocIndexer._align_frame(self, indexer, df)
   2355         val = df.reindex(index=ax)._values
   2356     return val
-> 2358 raise ValueError("Incompatible indexer with DataFrame")

ValueError: Incompatible indexer with DataFrame
```

Don't know yet if this is related to the `pandas` issue reported at https://github.com/pandas-dev/pandas/issues/47229 or if there is something we need to fix inside `audformat.utils.concat()`.

hagenw: Sorry, was my fault as `lambda y: y.mode()` was not returning the correct structure.

I updated now all examples to use `emodb`, see also the updated screenshots in the description.

hagenw:
The following raises an error as well:

```python
import audformat
import numpy as np

db = audformat.Database('db')
db.schemes['speaker'] = audformat.Scheme(
    'int',
    labels={
        0: {'gender': 'female'},
        1: {'gedner': 'male'},  # added a typo here
    },
)
index = audformat.filewise_index(['f1.wav', 'f2.wav'])
db['files'] = audformat.Table(index)
db['files']['speaker'] = audformat.Column(scheme_id='speaker')
db['files']['speaker'].set([0, 1])

db.get('gender')
```

results in

```
...
ValueError: Cannot map 'speaker' to 'gender'. Expected one of ['gedner'].
```

The error is raised already when using `audformat.Column.get(map='gender')`, but I think here it would make more sense to catch that error and return `['female', np.NaN]` as labels.

hagenw:
> The following raises an error as well:
> 
> ```python
> import audformat
> import numpy as np
> 
> db = audformat.Database('db')
> db.schemes['speaker'] = audformat.Scheme(
>     'int',
>     labels={
>         0: {'gender': 'female'},
>         1: {'gedner': 'male'},  # added a typo here
>     },
> )
> index = audformat.filewise_index(['f1.wav', 'f2.wav'])
> db['files'] = audformat.Table(index)
> db['files']['speaker'] = audformat.Column(scheme_id='speaker')
> db['files']['speaker'].set([0, 1])
> 
> db.get('gender')
> ```
> 
> results in
> 
> ```
> ...
> ValueError: Cannot map 'speaker' to 'gender'. Expected one of ['gedner'].
> ```
> 
> The error is raised already when using `audformat.Column.get(map='gender')`, but I think here it would make more sense to catch that error and return `['female', np.NaN]` as labels.

I created https://github.com/audeering/audformat/pull/403/ to fix the behavior in `audformat.Column.get()`.

hagenw:
> > The question is if we should simply return an empty dataframe in this case, use the column name to guess where the scheme belongs to (when strict=False) and/or raise an error instead of returning an empty dataframe (when strict=True)?
> 
> To me it feels natural to return an empty column in this case, just like if we do if we do not find the scheme.

I changed it accordingly and added tests for it.

hagenw:
> Found another problematic case, we need to fix in the code that tries to automatically convert to the correct dtype:
> 
> ```python
> import audformat
> import numpy as np
> 
> db = audformat.Database('db')
> db.schemes['speaker'] = audformat.Scheme('int', labels={0: {'gender': 'female'}})
> db.schemes['gender'] = audformat.Scheme('str', labels=['female', 'male'])
> db['files'] = audformat.Table(audformat.filewise_index(['f1', 'f2']))
> db['files']['speaker'] = audformat.Column(scheme_id='speaker')
> db['files']['speaker'].set([0, np.NaN])
> db['files']['gender'] = audformat.Column(scheme_id='gender')
> db['files']['gender'].set(['female', np.NaN])
> 
> db.get('gender')
> ```
> 
> returns
> 
> ```
> ...
> ValueError: Categorical categories cannot be null
> ```
> 
> This error does not happen, if we have only the `speaker` or only the `gender` scheme + column.
> 
> I will add a test for it and fix it inside the dtype conversion/detection code.

Added a test for it and fixed it. Now we get:

```
      gender
file        
f1    female
f2       NaN
```

hagenw:
My updates here lead to the following error on one of our internal databases:

```
ValueError: cannot handle a non-unique multi-index!
```

need to investigate this.

hagenw:
> My updates here lead to the following error on one of our internal databases:
> 
> ```
> ValueError: cannot handle a non-unique multi-index!
> ```
> 
> need to investigate this.

I looked into this, the problem is that one of the columns with the requested labels contains duplicated index entries.
I created a test for it and a fix in the code.

But I'm still not completely sure if we should support it as a database should not contain a duplicated index in the first place and it will add a little bit of overhead here.
Execution time is anyway a topic we might need to revisit. So maybe, we first add my suggested fix here and then in another pull request we try to speed up the function and see what we might remove/change then?

hagenw:
When specifying an `aggregate_function` it slows down the processing, even when it is not needed. We should see if we can improve on this (the underlying code is in `audformat.utils.concat()` and was added in https://github.com/audeering/audformat/pull/401

frankenjoe:
> But I'm still not completely sure if we should support it as a database should not contain a duplicated index in the first place and it will add a little bit of overhead here.

I agree, we don't have to deal with this case since duplicates in the index are now allowed.

I have lost a bit the overview here - is this ready for review again?

hagenw:
I have further looked at execution time of this using the `imda-nsc-read-speech-balanced` database which has 2,269,064 files.

Here are the current execution times for specific requests:

| Command | Execution time |
| --- | --- |
| `db.get('sex')` | 17 s |
| `db.get('sex', ['speaker', 'age'])` | 34 s |
| `db.get('sex', aggregate_function=lambda y: y[0])` | 222 s |
| `db.get('sex', ['speaker', 'age'], aggregate_function=lambda y: y[0])` | 1165 s |

To mitigate this I tried several approaches:
* speedup `audformat.utils.union()`
* use multi-threading in `audformat.utils.concat()`
* use multi-threading in `audformat.Database.get()`

None of the approaches was successful, so I would propose to stay with the current implementation for now.

hagenw:
> > But I'm still not completely sure if we should support it as a database should not contain a duplicated index in the first place and it will add a little bit of overhead here.
> 
> I agree, we don't have to deal with this case since duplicates in the index are not allowed.

But somehow the popup when dealing with a published database. I'm not completely sure yet if this is an error inside the database already or if it happens when collecting the data.
To be on the safe side, I simply drop duplicated entries now.

> I have lost a bit the overview here - is this ready for review again?

I tried to speed up the implementation, but did not succeed.
So, yes, this is now ready for review.

frankenjoe:
> Here are the current execution times for specific requests:

It might be useful to also include the execution time that it takes to get the `sex` values by directly calling `Column.get()`.

frankenjoe:
> I'm not completely sure yet if this is an error inside the database already or if it happens when collecting the data.

Could we find out? I think we have installed checks in `audb` that forbid publishing duplicated indices at some point. So it might be useful to find out if this happens for databases we published before the check or if we have a bug in the check for duplicated indices.

hagenw:
> > Here are the current execution times for specific requests:
> 
> It might be useful to also include the execution time that it takes to get the `sex` values by directly calling `Column.get()`.

Good idea, I added it. It's much faster as the database has 4 tables that have around 700,000 of overlapping entries.

frankenjoe:
Has the `strict` argument an influence in your example? Cause if so, we might consider to set it to `True` by default (or maybe change it to `loose` with `False` as default). Actually, independent of speed issues it might be safer to restrict to schemes by default.

hagenw:
> > I'm not completely sure yet if this is an error inside the database already or if it happens when collecting the data.
> 
> Could we find out? I think we have installed checks in `audb` that forbid publishing duplicated indices at some point. So it might be useful to know this can only happen with databases we published before the check was introduced or if that check is buggy.

I looked at the database, there are four segmented tables that have two duplicated index entries.
The database was originally published with Gradle and is very old, so it might be that we did not have the checks for duplicated indices included then.

frankenjoe:
> The database was originally published with Gradle and is very old, so it might be that we did not have the checks for duplicated indices included then.

Ok, good. Maybe it makes sense to publish a new version of the database then and redo the analysis with that new version.

hagenw:
> Has the `strict` argument an influence in your example? Cause if so, we might consider to set it to `True` by default (or maybe change it to `loose` with `False` as default). Actually, independent of speed issues it might be safer to restrict to schemes by default.

In this case it has a big influence and reduces `df = db.get('sex', strict=True)` to 0.001 s, but the reason is that `sex` is stored inside the labels of the `speaker` scheme as a dictionary, which is not considered with `strict=True` and an empty data frame is returned. As we have a lot of those databases or databases with missing schemes, I would be in favor of staying with `strict=False` as default.

hagenw:
> > The database was originally published with Gradle and is very old, so it might be that we did not have the checks for duplicated indices included then.
> 
> Ok, good. Maybe it makes sense to publish a new version of the database then and redo the analysis with that new version.

I created an issue for the corresponding database and removed the code from this pull request.
The times in the benchmarks above are now around 7% smaller.

frankenjoe:
> In this case it has a big influence and reduces `df = db.get('sex', strict=True)` to 0.001 s, but the reason is that `sex` is stored inside the labels of the `speaker` scheme as a dictionary, which is not considered with `strict=True` and an empty data frame is returned. As we have a lot of those databases or databases with missing schemes, I would be in favor of staying with `strict=False` as default.

That's what I was expecting. So don't you see the risk a user waits 17s because she does not set `strict=True`?

frankenjoe:
> That's what I was expecting. So don't you see the risk a user waits 17s because she does not set `strict=True`?

A sorry, it's fast because nothing is returned :)

> sex is stored inside the labels of the speaker scheme as a dictionary, which is not considered with strict=True

Maybe it should be considered also when `strict=True`?

frankenjoe:
> No, because the user want's to get the labels:

Yes, sorry, I realized that and made another comment above.

hagenw:
>> sex is stored inside the labels of the speaker scheme as a dictionary, which is not considered with strict=True
>
> Maybe it should be considered also when strict=True?

I'm not so sure about this. Semantically, it is on the same level as providing data in a column without a scheme. You provide values, but you don't specify the corresponding data type and if there should be some categories. In both cases we have to extract those information from the data, which motivated me to exclude both cases from the `strict=True` case.

For the benchmark results above it would also make no difference in results as the `speaker` scheme with the labels is assigned to a column in all four tables. Which means if you set `strict=True` and return mapped labels as well it will not be faster then the current solution with `strict=False`.

hagenw:
Yes, it's not very nice.

There might be one different take on how to collect the labels:
1. First find all tables that contain columns with the scheme
2. Order tables by size
3. Start with the largest one and get the labels from matching column(s) and store them with the index of the table
4. For the next bigger table extract the labels only for the difference of its index compared to the already stored values

This would lead to more than one column and a need for an aggregate function only in cases where the scheme is stored in different columns in the same table. My guess is that it would be much faster for most databases.

The problem is that it also has some undesired consequences: let's say you have a scheme `rating` and you have 5 tables each containing ratings from a single rater with the same index. If you request the `rating` scheme you will get labels only from one of those tables and don't be aware that there are more labels stored in other tables.

audformat/core/database.py
```diff
@@ -433,6 +441,398 @@ def duration(file: str) -> pd.Timedelta:
...
+            for n, y in enumerate(ys):
+                if not isinstance(y.dtype, pd.CategoricalDtype):
+                    ys[n] = y.astype(
+                        pd.CategoricalDtype(
+                            y.array.dropna().unique().astype(dtype)
+                        )
+                    )
+            # Find union of categorical data
+            data = [y.array for y in ys]
+            data = pd.api.types.union_categoricals(data)
```
ChristianGeng, line 801: I am encountering the data type error already discussed when the same schemes returns mixed types. 

The comment was this:

>> To me it sounds fine if we raise an error if we get different data types for the same scheme. Maybe we can explain in the error message to the user how she should fix the database. 


The steps for me to reproduce were this:

```python 

    db = audb.load(
        'mozillacommonvoice',
        version='10.3.0',
        only_metadata=True,
        verbose=False,
        full_path=False,
    )

data = db.get('age') 

```

Schemes conflicting  were 

```
$ys[2].array                                                                                                                                                                                                                                                    
[55, 55, 25, 25, 45, ..., 25, 25, 25, 25, 25]                                                                                                                                                                                                                        
 Length: 85986                                                                                                                                                                                                                                                        
 Categories (8, int64): [55, 25, 45, 35, 15, 65, 75, 85]                                                                                                                                                                                                              
$ ys[3].array                                                                                                                                                                                                                                                    
['fourties', 'fourties', 'fourties', 'fourties', 'fourties', ..., 'thirties', 'thirties', 'thirties', 'thirties', 'thirties']                                                                                                                                        
Length: 8375                                                                                                                                                                                                                                                         
Categories (9, object): ['teens', 'twenties', 'thirties', 'fourties', ..., 'sixties', 'seventies',                                                                                                                                                                   
                         'eighties', 'nineties']                                                                                                                                                                                                                     

```

My question is: How could the suggestion for the user look like? 

The categories in the database are not easily reconciled (the int typed one does not really allow mapping to the object typed one).

hagenw:
In order to avoid the need for #405, I changed the code to first try to concat without the aggregate function and only pass on the argument if concat fails. This does not change any outcomes we check in the tests, but speeds up `get()` for all cases where `aggregate_function` is not needed to join the labels.

I rerun the benchmark from https://github.com/audeering/audformat/pull/399#issuecomment-1803540737:

```python
import audb
import time

db = audb.load('imda-nsc-read-speech-balanced', only_metadata=True, full_path=False)

t = time.time()
db['files']['speaker'].get(map='sex')
print(time.time() - t)

t = time.time()
db.get('sex')
print(time.time() - t)

t = time.time()
db.get('sex', ['speaker', 'age'])
print(time.time() - t)

t = time.time()
db.get('sex', aggregate_function=lambda y: y[0])
print(time.time() - t)

t = time.time()
db.get('sex', ['speaker', 'age'], aggregate_function=lambda y: y[0])
print(time.time() - t)
```

| Command | Execution time |
| --- | --- |
| `db['files']['speaker'].get(map='sex')` | 0.16 s |
| `db.get('sex')` | 15 s |
| `db.get('sex', ['speaker', 'age'])` | 31 s |
| `db.get('sex', aggregate_function=lambda y: y[0])` | 15 s |
| `db.get('sex', ['speaker', 'age'], aggregate_function=lambda y: y[0])` | 31 s |

audformat/core/database.py
hagenw: I have updated the error message and your example now results in:

```
...
TypeError: Cannot join labels for scheme 'age' with different data types: int64, object
```

frankenjoe:
Cool, thank you. So it looks like the aggregate function is no longer a bottleneck, at least when no duplicates are found. The jump in execution time from 0.16s to 15s is still quite large. Do you know what other bottlenecks we have here and do you see room for further improvements or is this something we have to live with? I'm just a bit afraid users will avoid using the function if it creates such an overhead.

hagenw:
At the moment I don't see any obvious solution to speed it further up. There is a different approach how to collect labels that would provide a speedup as described in https://github.com/audeering/audformat/pull/399#issuecomment-1805290804. But that algorithm is more an approximation of finding the correct labels and I would like to avoid it.

The jump from 0.16s to 15s can be explained by `imda-nsc-read-speech-balanced` being one of the worst databases regarding label collection. It has 2,269,064 samples and four tables containing the requested scheme that have an overlap of 756,354 samples which makes `audformat.utils.concat()` slow, I guess mainly `audformat.utils.union()` is the bottleneck.

E.g. when using `mozillacommonvoice`, which has 83 tables, collecting all gender labels manually would be cumbersome, but executing `df = db.get('gender')` takes 9 s. For `agender` it takes 0.3 s and 0.02 s for `timit`.

My hope would be that later on we find ways to speed it up (e.g. by changing the algorithm or by using a different engines to handle types (https://github.com/audeering/audformat/issues/382)) and the performance is fast enough for a first release.

hagenw:
There is one other interesting case we do not cover yet with `audformat.Database.get()`:

```python
>>> db = audb.load('emodb', version='1.4.1', only_metadata=True, full_path=False)
>>> db['files']['transcription'].get().head()
file
wav/03a01Fa.wav    a01
wav/03a01Nc.wav    a01
wav/03a01Wa.wav    a01
wav/03a02Fc.wav    a02
wav/03a02Nc.wav    a02
Name: transcription, dtype: category
Categories (10, object): ['a01', 'a02', 'a04', 'a05', ..., 'b02', 'b03', 'b09', 'b10']

>>> db['files']['transcription'].get(map='transcription').head()
file
wav/03a01Fa.wav    Der Lappen liegt auf dem Eisschrank.
wav/03a01Nc.wav    Der Lappen liegt auf dem Eisschrank.
wav/03a01Wa.wav    Der Lappen liegt auf dem Eisschrank.
wav/03a02Fc.wav       Das will sie am Mittwoch abgeben.
wav/03a02Nc.wav       Das will sie am Mittwoch abgeben.
Name: transcription, dtype: string
```

whereas we can only do:

```python
>>> db.get('transcription').head()
                transcription
file                         
wav/03a01Fa.wav           a01
wav/03a01Nc.wav           a01
wav/03a01Wa.wav           a01
wav/03a02Fc.wav           a02
wav/03a02Nc.wav           a02
```

Should we ignore this for now, or add something like a `map=False` to cover the current case as default and `map=True` to return?
```python
>>> db.get('transcription', map=True).head()
file
wav/03a01Fa.wav    Der Lappen liegt auf dem Eisschrank.
wav/03a01Nc.wav    Der Lappen liegt auf dem Eisschrank.
wav/03a01Wa.wav    Der Lappen liegt auf dem Eisschrank.
wav/03a02Fc.wav       Das will sie am Mittwoch abgeben.
wav/03a02Nc.wav       Das will sie am Mittwoch abgeben.
```

frankenjoe:
> Should we ignore this for now, or add something like a map=False to cover the current case as default and map=True to return?

Yes, I think that makes sense.

hagenw:
I added now `map` argument to handle simple mappings, but I decided to go with `True` as the default argument, as the mappings are usually the actual labels you are looking for.
So now we have:

```python
>>> db = audb.load('emodb', version='1.4.1', only_metadata=True, full_path=False)
>>> db.get('transcription').head()
                                        transcription
file                                                 
wav/03a01Fa.wav  Der Lappen liegt auf dem Eisschrank.
wav/03a01Nc.wav  Der Lappen liegt auf dem Eisschrank.
wav/03a01Wa.wav  Der Lappen liegt auf dem Eisschrank.
wav/03a02Fc.wav     Das will sie am Mittwoch abgeben.
wav/03a02Nc.wav     Das will sie am Mittwoch abgeben.
```

hagenw:
I integrated now the `aggregate_strategy` argument from https://github.com/audeering/audformat/pull/405 and checked that the `Database.get()` still works for all our databases and the benchmark times stayed the same.

Which means this should also be finally ready to review/merge.

frankenjoe:
Cool, another great feature in place!

PR #401, opened 2023-10-25
Add aggregate_function argument to utils.concat()

hagenw:
This adds the `aggregate_function` argument to `audformat.utils.concat()`.
If `overwrite=False` and `aggregate_function` is not `None` it will be used as a callable that combines values for all entries that have more than one data point.

image

**NOTE**: in `test_utils_concat.py` only the tests `test_concat_aggregate_function()` and `test_concat_overwrite_aggregate_function()` are new, the other part is moved from `test_util.py`.


audformat/core/utils.py
```diff
@@ -269,6 +312,25 @@ def concat(
...
+                axis=1,
+                ignore_index=True,
+            )
+            dtype = columns_reindex[column].dtype
+            columns_reindex[column] = df.apply(aggregate_function, axis=1)
+            # Restore the original dtype if possible
+            try:
+                columns_reindex[column] = columns_reindex[column].astype(dtype)
+            except (TypeError, ValueError):
+                pass
```
frankenjoe, line 332: Could it make sense to raise here to avoid the `dtype` silently changes?

audformat/core/utils.py
```diff
@@ -233,26 +261,41 @@ def concat(
...
+                        )
+                        column = column.loc[~column.index.isin(intersection)]
+
+                    else:
+                        combine = pd.DataFrame(
+                            {
+                                'left':
+                                columns_reindex[column.name][intersection],
+                                'right':
+                                column[intersection]
```
frankenjoe, line 282:
```suggestion
                                column[intersection],
```

audformat/core/utils.py
```diff
@@ -64,18 +65,35 @@ def concat(
...
+    that converts the overlapping values
+    into a single value.
 
     Args:
         objs: objects
         overwrite: overwrite values where indices overlap
+        aggregate_function: function to aggregate values
+            for all entries
+            that contain more than one value per index.
+            The function gets a dataframe row as input.
```
frankenjoe, line 80:
```suggestion
            The function gets a :class:`pd.DataFrame` row as input.
```

But maybe more explicit:

```
gets a :class:`pd.Series` with the overlapping values as input
```

?

audformat/core/utils.py
```diff
@@ -64,18 +65,35 @@ def concat(
...
+    with ``aggregate_function``
+    that converts the overlapping values
+    into a single value.
 
     Args:
         objs: objects
         overwrite: overwrite values where indices overlap
+        aggregate_function: function to aggregate values
+            for all entries
+            that contain more than one value per index.
```
frankenjoe, line 79: Maybe simplify to:

```suggestion
        aggregate_function: function to aggregate overlapping values.
```

frankenjoe:
```python
index = audformat.filewise_index(['f1', 'f2', 'f3', 'f4'])
df1 = pd.DataFrame(
    {
        'a': [1, 1, 1],
        'b': [1, 1, 1],
    },
    index=index[:3],
)
df2 = pd.DataFrame(
    {
        'a': [2, 2, 2],
        'b': [2, 2, 2],
    },
    index=index[1:],
)

audformat.utils.concat([df1, df2], aggregate_function=np.sum)
```
```
         a     b
file            
f1    <NA>  <NA>
f2       3     3
f3       3     3
f4       2     2
```

But I was expecting:

```
         a     b
file            
f1       1     1
f2       3     3
f3       3     3
f4       2     2
```

hagenw:
Thanks for spotting this, I fixed it (and a related issue) and now we get:

```
      a  b
file      
f1    1  1
f2    3  3
f3    3  3
f4    2  2
```

audformat/core/utils.py
hagenw: Not really. E.g. when we decide to aggregate the values inside a tuple, the dtype has to change.

Note, there was also an error and I had to rewrite this section as:

```python
            y = df.apply(aggregate_function, axis=1)
            # Restore the original dtype if possible
            try:
                y = y.astype(dtype)
            except (TypeError, ValueError):
                columns_reindex[column] = columns_reindex[column].astype(
                    y.dtype
                )
            columns_reindex[column].loc[y.index] = y
```

audformat/core/utils.py
hagenw: Good idea, I changed to your second suggestion.

audformat/core/utils.py
```diff
@@ -64,18 +65,35 @@ def concat(
...
         overwrite: overwrite values where indices overlap
+        aggregate_function: function to aggregate overlapping values.
+            The function gets a :class:`pandas.Series`
+            with overlapping values
+            as input.
+            E.g. set to
+            :func:`numpy.mean`
+            to average the values
+            or to
+            ``lambda row: tuple(row.to_list())``
```
frankenjoe, line 85: I think we can simply do:

```suggestion
            ``lambda row: tuple(row)``
```

and I think I would rewrite to:

```
lambda y: tuple(y)
```

audformat/core/utils.py
```diff
@@ -64,18 +65,35 @@ def concat(
...
+            to return all values as a tuple
```
frankenjoe, line 86:
```suggestion
            to return them as a tuple
```

audformat/core/utils.py
hagenw: Changed to your second suggestion

frankenjoe:
> Thanks for spotting this, I fixed it (and a related issue) and now we get:

I can confirm that example is now working with two frames, but when I add a third one I get:

```python
index = audformat.filewise_index(['f1', 'f2', 'f3', 'f4'])
df1 = pd.DataFrame(
    {
        'a': [1, 1, 1],
        'b': [1, 1, 1],
    },
    index=index[:3],
)
df2 = pd.DataFrame(
    {
        'a': [2, 2, 2],
        'b': [2, 2, 2],
    },
    index=index[1:],
)
df3 = pd.DataFrame(
    {
        'a': [3],
        'b': [3],
    },
    index[:1],
)

audformat.utils.concat([df1, df2, df3], aggregate_function=np.sum)
```
```
      a  b
file      
f1    1  1
f2    3  3
f3    3  3
f4    2  2
```

but what I would expect is:

```
      a  b
file      
f1    4  4
f2    3  3
f3    3  3
f4    2  2
```

audformat/core/utils.py
```diff
@@ -64,18 +65,35 @@ def concat(
...
         overwrite: overwrite values where indices overlap
+        aggregate_function: function to aggregate overlapping values.
+            The function gets a :class:`pandas.Series`
+            with overlapping values
+            as input.
+            E.g. set to
+            :func:`numpy.mean`
+            to average the values
+            or to
+            ``lambda y: tuple(y)``
```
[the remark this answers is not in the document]
frankenjoe: I guess it woudn't be too bad to have at least one example `lambda` example, but we could also do this in the `Examples` section.

hagenw:
Thanks for finding another example that didn't worked, there was indeed a bigger error in how the overlapping values were collected as for the very first column we need to collect all values that overlap with any of the other columns. Whereas for all other columns it is just fine to collect values that overlap with the first column. I changed the code accordingly, and added a few more tests (most likely still not enough ;) ).

Now we get:

```python
import audformat
import numpy as np
import pandas as pd

index = audformat.filewise_index(['f1', 'f2', 'f3', 'f4'])
df1 = pd.DataFrame(
    {
        'a': [1, 1, 1],
        'b': [1, 1, 1],
    },
    index=index[:3],
)
df2 = pd.DataFrame(
    {
        'a': [2, 2, 2],
        'b': [2, 2, 2],
    },
    index=index[1:],
)
df3 = pd.DataFrame(
    {
        'a': [3],
        'b': [3],
    },
    index[:1],
)

audformat.utils.concat([df1, df2, df3], aggregate_function=np.sum)
```

```
      a  b
file      
f1    4  4
f2    3  3
f3    3  3
f4    2  2
```

audformat/core/utils.py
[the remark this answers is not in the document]
hagenw: I now replaced `numpy.mean` by `lambda y: y.mean()` but stayed in the Examples section with `np.sum` to show both possibilities.

frankenjoe:
Looks good, can't find another failing example. Nice extension, strange we didn't come up with this aggregate solution earlier.

hagenw:
I guess there was no urgent need to have it earlier.

PR #407, opened 2023-11-30
CI: simplify changelog parsing

hagenw:
Before we were replacing new line and other characters https://github.com/actions/create-release/issues/25#issuecomment-562687270. But as set-output is now deprecated we are no longer using it, the current approach leads to malformed changelog lists on the Github releases page.


hagenw:
This solution is working see https://github.com/audeering/audformat/releases/tag/v1.1.0

PR #416, opened 2024-02-02
Support loading db with splits in misc tables

hagenw:
It was possible to assign splits to misc tables before and also to store those databases. But when trzing to load a database with a misc table that has a split assigned, loading failed as the splits objects were loaded after the misc tables were loaded.

This pull request adds a test for loading a database with a misc table that has a split assigned. This tests fails for the current master. The needed fix is then added to the code of `audformat.Database.load()`.


maxschmitt (approved):
It all looks good to me.

PR #419, opened 2024-03-20
Store tables as PARQUET files

### LINKED ISSUE #376: Support an additional file format for tables
One of the biggest limitations we are currently facing is due to our usage of the CSV format to store tables.

There are strong arguments in favour of CSV as it is widely supported and human readable. I would not propose to replace it, but provide the possibility that tables can be provided in a different format as well.

The additional format should support storing array data like embeddings in its columns and it should support reading only parts of it into memory to not have a limit on database sizes.

I don't see too much overhead in implementing it. We could add an argument to `audformat.Table`  to indicate when it should not be stored as CSV, and in addition we can also make it dependent on the used scheme, e. g. if somebody specifies an array scheme (which we will need to add) to a column the table will then automatically not stored as CSV.

This way we could for example add an embeddings table to a database containing embeddings from different models in its columns.

### LINKED ISSUE #382: Switch to pyarrow engine when reading CSV files

hagenw:
Closes #382
Closes #376 

Several improvement/speed ups to the table handling by using `pyarrow`:

* Support for storing tables as PARQUET files
* Reading tables from CSV files with `pyarrow`
* Require `pandas >=2.1.0`
* Require `pyarrow`
* Remove support for Python 3.8 (as we need `pandas >=2.1.0`)
* Add a new CI job "ubuntu-latest, 3.9, minimum", that tests with all Python packages installed at its minimum version

I decided to stay with `"csv"` as the default setting for the `storage_format` argument in `audformat.Database.save()` and `audformat.Table.save()`. This way, we could make a new release of `audformat` and can test storing tables as PARQUET file with some databases, without exposing it to other users as well. In addition, `audb` needs to be updated to support publishing PARQUET tables.

### Loading benchmark

We can benchmark the behavior with loading a dataset from a folder, that contains all tables with
`audformat.Database.load(db_root, load_data=True)`. Results are given as average over 10 runs.

| Dataset | # table rows | CSV (previous) | CSV (new) | PARQUET | PKL |
| --- | --- | --- | --- | --- | --- |
| mozillacommonvoice | 1,923,269 | 8.62 ± 0.3 s | 4.76 ± 0.6 s | 3.58 ± 0.7 s | 2.30 ± 0.8 s |
| voxceleb2 | 2,894,987 | 17.44 ± 0.3 s  | 6.05 ± 0.2 s | 1.33 ± 0.1 s | 1.01 ± 0.1 s |
| librispeech | 36,422,346 | 1036.1 ± 30 s  | 284.95 ± 17.3 s | 26.88 ± 0.7 s | 3.01 ± 1.0 s |
| imda-nsc-read-speech-balanced | 4,538,128 | 8.38 ± 0.2 s | 2.86 ± 0.4 s | 3.07 ± 0.3 s | 2.04 ± 0.3 s |
| emodb | 1,605 | 0.02 ± 0.00 s | 0.05 ± 0.01 s  | 0.03 ± 0.01 s | 0.03 ± 0.07 s |

<details><summary>Benchmark code</summary>

```python
import os
import time

import numpy as np

import audb
import audeer
import audformat


repetitions = 10

db_root = audeer.path("./db")
audeer.rmdir(db_root)
cache_root = audeer.path("./cache")
audeer.rmdir(cache_root)
audeer.mkdir(cache_root)

datasets = [
    ("mozillacommonvoice", "10.3.0"),
    ("voxceleb2", "2.6.0"),
    ("librispeech", "3.1.0"),
    ("imda-nsc-read-speech-balanced", "2.6.0"),
    ("emodb", "1.4.1"),
]
for name, version in datasets:
    audeer.rmdir(db_root)
    audb.load_to(
        db_root,
        name,
        version=version,
        only_metadata=True,
        cache_root=cache_root,
        num_workers=1,
    )

    # PKL files
    times = []
    for _ in range(repetitions):
        t0 = time.time()
        db = audformat.Database.load(db_root, load_data=True)
        t = time.time() - t0
        times.append(t)
    print(
        f"Execution time with PKL tables {name}: "
        f"{np.mean(times):.3f}±{np.std(times):.3f}s"
    )

    # CSV files
    pkl_files = audeer.list_file_names(db_root, filetype="pkl")
    for pkl_file in pkl_files:
        os.remove(pkl_file)
    times = []
    for _ in range(repetitions):
        t0 = time.time()
        db = audformat.Database.load(db_root, load_data=True)
        t = time.time() - t0
        times.append(t)
    print(
        f"Execution time with CSV tables {name}: "
        f"{np.mean(times):.3f}±{np.std(times):.3f}s"
    )

    # PARQUET files
    if audformat.__version__ > "1.1.4":
        # Replace CSV with PARQUET files
        # and repeat benchmark
        for table_id in list(db):
            path = audeer.path(db_root, f"db.{table_id}")
            db[table_id].save(path, storage_format="parquet")
            os.remove(f"{path}.csv")
        times = []
        for _ in range(repetitions):
            t0 = time.time()
            db = audformat.Database.load(db_root, load_data=True)
            t = time.time() - t0
            times.append(t)
        print(
            f"Execution time with PARQUET tables {name}: "
            f"{np.mean(times):.3f}±{np.std(times):.3f}s"
        )
```

</details>

The benchmark highlights two important results:

* The new implementation speeds up loading of CSV files by a factor of ~4.
* Reading from PARQUET is at least as fast as reading from CSV, and for some datasets much faster.

### Memory benchmark

I investigated memory consumption using [heaptrack](https://github.com/KDE/heaptrack), when loading the `phonetic-transcription.train-clean-360` table from the `librispeech` dataset from our internal server. Stored as CSV file the table has a size of 1,3 G, stored as PARQUET file 49 M.

| branch | peak heap memory | execution time |
| --- | --- | --- |
| CSV (main) | 1.36 G | 197.9 s |
| CSV (pull request) | 1.33 G | 43.9 s |
| PARQUET (pull request) | 1.11 G | 3.7 s |

<details><summary>Benchmark code</summary>

`csv-table-loading.py`
```python
import time

import audb
import audformat


# Measure execution time when loading a CSV table.
# Assumes the file `db.phonetic-transcription.train-clean-360.csv`
# or `db.phonetic-transcription.train-clean-360.parquet`
# exists in the same folder
# as this script.
# Note, that the CSV file is only loaded
# when the PARQUET file is not present.

db = audb.info.header("librispeech")
table_id = "phonetic-transcription.train-clean-360"
t0 = time.time()
db[table_id].load(f"db.{table_id}")
t = time.time() - t0
print(f"Execution time: {t:.1f} s")
```

The memory consumption is then profiled with:
```bash
$ heaptrack python csv-table-loading.py
```

The execution time was measured without `heaptrack`:
```bash
$ python csv-table-loading.py
```

</details>

### Why and when is reading from a CSV file slow?

By far the slowest part when reading a CSV file with `pyarrow` is the conversion to `pandas.Timedelta` values for columns, that specify a duration.
E.g. when reading the CSV from the memory benchmark, the reading with `pyarrow` and conversion to `pandas.DataFrame` takes **3 s**, whereas the conversion of the `start` and `end` column to `pandas.Timedelta` takes roughly **40 s**.
There is a dedicated dtype with [`pyarrow.duration`](https://arrow.apache.org/docs/python/generated/pyarrow.duration.html), but it does not have yet reading support for CSV files. When trying so, you get:

```
pyarrow.lib.ArrowNotImplementedError: CSV conversion to duration[ns] is not supported
```

When storing a table as a PARQUET file we use `duration[ns]` for time values, and converting those back to `pandas.Timedelta` seems to be much faster. Reading the PARQUET file needs **0.3 s**, converting to the dataframe then takes the remaining **3.4 s**.

The conversion can very likely be speed up when switching to use `pyarrow` based dtypes in the dataframe as we do for `audb.Dependencies._df`, but at the moment this is not fully supported in `pandas`, e.g. timedelta values are not implemented yet (https://github.com/pandas-dev/pandas/issues/52284).

### Hashing of PARQUET files

As we have experienced already at https://github.com/audeering/audb/pull/372, PARQUET files are not stored in a reproducible fashion and might return different MD5 hash sums, even though they store the same dataframe.
To overcome this problem, I calculate now a hash based on the content of the dataframe and store the resulting value inside the metadata of schema of the PARQUET file. Which means in `audb` we can access it by just loading the schema of a PARQUET file. The corresponding code to access the hash is:
```python
import pyarrow.parquet as parquet

hash = parquet.read_schema(path).metadata[b"hash"].decode()
```

This approach is faster than calculating the MD5 sum with `audeer.md5()`.
Execution time benchmarked as average over 100 repetitions:

| method | execution time |
| --- | --- |
| read hash | 0.0001 ± 0.0000 s |
| calculate md5 | 0.0716 ± 0.0007 s |

<details><summary>Benchmark code</summary>

```python
import time

import numpy as np
import pyarrow.parquet as parquet

import audeer

# Compare execution times
# for calculating the MD5 sum
# vs reading the hash from the PARQUET file.
# Assumes the file `db.phonetic-transcription.train-clean-360.parquet`
# exists in the same folder
# as this script.
table_id = "phonetic-transcription.train-clean-360"
repetitions = 100 

path = f"db.{table_id}.parquet"
times = []
for _ in range(repetitions):
    t0 = time.time()
    audeer.md5(path)
    t = time.time() - t0
    times.append(t)
print(f"Execution time MD5: {np.mean(times):.4f} ± {np.std(times):.4f} s")

times = []
for _ in range(repetitions):
    t0 = time.time()
    parquet.read_schema(path).metadata[b"hash"].decode()
    t = time.time() - t0
    times.append(t)
print(f"Execution time hash: {np.mean(times):.4f} ± {np.std(times):.4f} s")
```

</details>

### Writing benchmark

Comparison of saving the `phonetic-transcription.train-clean-360` table from `librispeech` in different formats.
Note, saving as a PARQUET file includes calculating of the hash of the underlying dataframe.

| format | execution time |
| --- | --- |
| CSV | 22.2 s |
| PARQUET | 5.1 s |
| PKL | 1.0 s |

<details><summary>Benchmark code</summary>

```python
import os
import time

import audb
import audformat


# Measure execution time when saving table.
# Assumes the file `db.phonetic-transcription.train-clean-360.csv`
# or `db.phonetic-transcription.train-clean-360.parquet`
# exists in the same folder
# as this script.
db = audb.info.header("librispeech")
table_id = "phonetic-transcription.train-clean-360"
db[table_id].load(f"db.{table_id}")

t0 = time.time()
db[table_id].save("table", storage_format="csv")
t = time.time() - t0
os.remove("table.csv")
print(f"Execution time CSV: {t:.1f} s")

t0 = time.time()
db[table_id].save("table", storage_format="parquet")
t = time.time() - t0
os.remove("table.parquet")
print(f"Execution time PARQUET: {t:.1f} s")

t0 = time.time()
db[table_id].save("table", storage_format="pkl")
t = time.time() - t0
os.remove("table.pkl")
print(f"Execution time PKL: {t:.1f} s")
```

</details>

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -819,6 +819,102 @@ class Base(HeaderBase):
 
         return self
 
+    def _convert_pyarrow_dtypes(
+        self,
+        df: pd.DataFrame,
+        *,
+        convert_all: bool = False,
+    ) -> pd.DataFrame:
+        r"""Convert dtypes that are not handled by pyarrow.
+
+        This adjusts dtypes in a dataframe,
+        that could not be set correctly
+        when converting to the dataframe
+        from pyarrow.
+
+        Args:
+            df: dataframe,
+            convert_all: if ``False``,
+                converts all columns with
+                ``"object"`` audformat dtype,
+                and all columns with a scheme with labels.
+                If ``"True"``,
+                it converts additionally all columns with
+                ``"bool"``, ``"int"``, and ``"time"`` audformat dtypes
+
+        Returns:
+            dataframe with converted dtypes
+
+        """
+        # Collect columns with dtypes,
+        # that cannot directly be converted
+        # from pyarrow to pandas
+        bool_columns = []
+        int_columns = []
+        time_columns = []
+        object_columns = []
+
+        # Collect columns
+        # with scheme labels
+        labeled_columns = []
+
+        # Collect columns,
+        # belonging to the index
+        index_columns = []
+
+        # --- Index ---
+        index_columns += list(self.levels.keys())
+        for level, dtype in self.levels.items():
+            if dtype == define.DataType.BOOL:
+                bool_columns.append(level)
+            elif dtype == define.DataType.INTEGER:
+                int_columns.append(level)
+            elif dtype == define.DataType.TIME:
+                time_columns.append(level)
+            elif dtype == define.DataType.OBJECT:
+                object_columns.append(level)
+
+        # --- Columns ---
+        for column_id, column in self.columns.items():
+            if column.scheme_id is not None:
+                scheme = self.db.schemes[column.scheme_id]
+                if scheme.labels is not None:
+                    labeled_columns.append(column_id)
+                elif scheme.dtype == define.DataType.BOOL:
+                    bool_columns.append(column_id)
+                elif scheme.dtype == define.DataType.INTEGER:
+                    int_columns.append(column_id)
+                elif scheme.dtype == define.DataType.TIME:
+                    time_columns.append(column_id)
+                elif scheme.dtype == define.DataType.OBJECT:
+                    object_columns.append(column_id)
+            else:
+                # No scheme defaults to `object` dtype
+                object_columns.append(column_id)
+
+        if convert_all:
+            for column in bool_columns:
+                df[column] = df[column].astype("boolean")
+            for column in int_columns:
+                df[column] = df[column].astype("Int64")
+            for column in time_columns:
+                df[column] = df[column].astype("timedelta64[ns]")
+        for column in object_columns:
+            df[column] = df[column].astype("object")
+            df[column] = df[column].replace(pd.NA, None)
+        for column in labeled_columns:
+            scheme = self.db.schemes[self.columns[column].scheme_id]
+            labels = scheme._labels_to_list()
+            if len(labels) > 0 and isinstance(labels[0], int):
+                # allow nullable
+                labels = pd.array(labels, dtype="int64")
+            dtype = pd.api.types.CategoricalDtype(
+                categories=labels,
+                ordered=False,
+            )
+            df[column] = df[column].astype(dtype)
+        return df
+
     def _get_by_index(
         self,
         index: pd.Index,
@@ -841,134 +937,43 @@ class Base(HeaderBase):
             path: path to table, including file extension
 
         """
-        schemes = self.db.schemes
-
-        # === Infer dtypes ===
-
         # Collect pyarrow dtypes
-        # of the CSV file,
-        # by inspecting the audformat schemes,
-        # and the index
-        # associated with the table.
-        # The dtypes are used to create
+        # of all columns,
+        # including index columns.
+        # The dtypes are stored as a tuple
+        # ``(column, dtype)``,
+        # and are used to create
         # the pyarrow.Schema
         # used when reading the CSV file
         pyarrow_dtypes = []
-
-        # Collect columns,
-        # that cannot directly be converted
-        # from pyarrow to pandas
-        timedelta_columns = []
-        boolean_columns = []
-        object_columns = []
-        integer_columns = []
-
-        # Collect columns,
-        # belonging to the index
-        index_columns = []
-
-        # --- Index ---
-        if hasattr(self, "type"):
-            levels = {}
-            # filewise or segmented table
-            levels[define.IndexField.FILE] = define.DataType.STRING
-            if self.type == define.IndexType.SEGMENTED:
-                # segmented table
-                for level in [define.IndexField.START, define.IndexField.END]:
-                    levels[level] = define.DataType.TIME
-        else:
-            # misc table
-            levels = self.levels
-        index_columns += list(levels.keys())
-        for name, dtype in levels.items():
-            pyarrow_dtype = to_pyarrow_dtype(dtype)
-            if pyarrow_dtype is not None:
-                pyarrow_dtypes.append((name, pyarrow_dtype))
-                if dtype == define.DataType.TIME:
-                    timedelta_columns.append(name)
-                elif dtype == define.DataType.INTEGER:
-                    integer_columns.append(name)
-                elif dtype == define.DataType.BOOL:
-                    boolean_columns.append(name)
-            else:
-                object_columns.append(name)
-
-        # --- Columns ---
-        categories = {}
-        columns = list(self.columns)
+        # Index
+        for level, dtype in self.levels.items():
+            if dtype != define.DataType.OBJECT:
+                pyarrow_dtypes.append((level, to_pyarrow_dtype(dtype)))
+        # Columns
         for column_id, column in self.columns.items():
             if column.scheme_id is not None:
-                scheme = schemes[column.scheme_id]
-                if scheme.labels is not None:
-                    categories[column_id] = scheme._labels_to_list()
-                pyarrow_dtype = to_pyarrow_dtype(scheme.dtype)
-                if pyarrow_dtype is not None:
-                    pyarrow_dtypes.append((column_id, pyarrow_dtype))
-                    if scheme.dtype == define.DataType.TIME:
-                        timedelta_columns.append(column_id)
-                    elif scheme.dtype == define.DataType.BOOL:
-                        boolean_columns.append(column_id)
-                    elif scheme.dtype == define.DataType.INTEGER:
-                        integer_columns.append(column_id)
-                else:
-                    object_columns.append(column_id)
-            else:
-                object_columns.append(column_id)
+                dtype = self.db.schemes[column.scheme_id].dtype
+                if dtype != define.DataType.OBJECT:
+                    pyarrow_dtypes.append((column_id, to_pyarrow_dtype(dtype)))
 
-        # === Read CSV ===
-        schema = pa.schema(pyarrow_dtypes)
+        # Read CSV file
         table = csv.read_csv(
             path,
             read_options=csv.ReadOptions(
-                column_names=index_columns + columns,
+                column_names=list(self.levels.keys()) + list(self.columns.keys()),
                 skip_rows=1,
             ),
             convert_options=csv.ConvertOptions(
-                column_types=schema,
+                column_types=pa.schema(pyarrow_dtypes),
                 strings_can_be_null=True,
             ),
         )
-        df = table.to_pandas(
-            deduplicate_objects=False,
-            types_mapper={
-                pa.string(): pd.StringDtype(),
-            }.get,  # we have to provide a callable, not a dict
-        )
+        df = self._pyarrow_table_to_dataframe(table)
 
-        # === Adjust dtypes ===
-
-        # Adjust dtypes, that cannot be handled by pyarrow
-        for column in timedelta_columns:
-            df[column] = df[column].astype("timedelta64[ns]")
-        for column in boolean_columns:
-            df[column] = df[column].astype("boolean")
-        for column in object_columns:
-            df[column] = df[column].astype("object")
-            df[column] = df[column].replace(pd.NA, None)
-        for column in integer_columns:
-            df[column] = df[column].astype("Int64")
-        for column, labels in categories.items():
-            if len(labels) > 0 and isinstance(labels[0], int):
-                # allow nullable
-                labels = pd.array(labels, dtype="int64")
-            dtype = pd.api.types.CategoricalDtype(
-                categories=labels,
-                ordered=False,
-            )
-            df[column] = df[column].astype(dtype)
-
-        # === Set index ===
-
-        # When assigning more than one column,
-        # a MultiIndex is assigned.
-        # Setting a MultiIndex does not always preserve pandas dtypes,
-        # so we need to set them manually.
-        #
-        if len(index_columns) > 1:
-            index_dtypes = {column: df[column].dtype for column in index_columns}
-        df.set_index(index_columns, inplace=True)
-        if len(index_columns) > 1:
-            df.index = utils.set_index_dtypes(df.index, index_dtypes)
+        # Adjust dtypes and set index
+        df = self._convert_pyarrow_dtypes(df, convert_all=True)
+        df = self._set_index(df, list(self.levels.keys()))
 
         self._df = df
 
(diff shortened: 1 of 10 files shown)


audformat/core/table.py
```diff
@@ -891,14 +928,226 @@ def _load_pickled(self, path: str):
...
         self._df = df
 
+    def _pyarrow_convert_dtypes(
+        self,
+        df: pd.DataFrame,
+        *,
+        convert_all: bool = False,
+    ) -> pd.DataFrame:
+        r"""Convert dtypes that are not handled by pyarrow.
+
```
ChristianGeng, line 938:
```suggestion
```

causes D202: "No blank lines allowed after function docstring"

hagenw: Now, I'm getting:

```
D205 1 blank line required between summary line and description
```

Seems like our settings are not compatible?

hagenw: >  "No blank lines allowed after function docstring"

The docstring was also not finished at this line, just the first line of it. Maybe you put your suggestion on the wrong line?

tests/test_table.py
```diff
@@ -1118,22 +1122,25 @@ def test_load(tmpdir):
...
+        table_loaded._db = table._db
+        table_loaded.load(path_no_ext)
+        pd.testing.assert_frame_equal(table.df, table_loaded.df)
 
-    # check if pickle file was recovered from CSV
-    df = pd.read_pickle(path_pkl)
-    pd.testing.assert_frame_equal(table.df, df)
+        # check if pickle file was recovered
+        df = pd.read_pickle(path_pkl)
+        pd.testing.assert_frame_equal(table.df, df)
```
ChristianGeng, line 1141: I remember that I had some strange asymmetry concerning the results of `pd.testing.assert_frame_equal`  and `pandas.DataFrame.equals`. Digging into the code seems to also suggests that the implementation in the testing module does not rely on the core data frame function.

hagenw: `pandas.DataFrame.equals` does not check for the same dtype:

image

`pandas.testing.assert_frame_equal` does:

image

So, if you really want to ensure that two dataframes are identical in a test, `pandas.testing.assert_frame_equal` should be the choice.

ChristianGeng (approved):
# Review

Store tables as PARQUET files #419 <https://github.com/audeering/audformat/pull/419>


## minimum dependencies

downgrade to minimum dependencies is a nice feature for having the earliest pandas version supported.


## Hashing

Hashing depends on the pandas utility in `pandas/pandas/core/util/hashing.py` - via `audformat.utils.hash`.

The pandas function has a few input arguments, most of its defaults are not worrying. I would e.g. not assume that utf-encoding would change anytime in the near future. I also would assume that the keeping the index to "True" makes sense as the function only targes series, frames and audformat-type indices which is stricter than only hashing values.

However what I find a bit worrying is that the line defining the `_default_hash_key` has been changed only two months ago:

```bash
git blame pandas/core/util/hashing.pyp  | grep "_default_hash_key ="
```

The hash key's value seems kind of generic but I wonder whether it only had been introduced recently, or whether it is changed sometimes?

So far it has been only changed in a single commit:

```bash
git log -G "_default_hash_key =" --format=%H -- pandas/core/util/hashing.py
```

    b1525c4a3788d161653b04a71a84e44847bedc1b

This is good. But will it be safely backwards compatible in the future? Or does it have to be future proof at all? I have searched the code of the pandas library. From what I have seen, the hashing function is never used in order to verify the identity of an object or artifact - it seems only to be used for algorithmic (speedup) purposes. So the pandas functioning in a future pandas release is unlikely to be hampered when the underlying hashing would be changed - our use as identifying objects would possibly.


## Status of `_pyarrow_convert_dtypes`

Lots of the new functionality is covered by retyping `table._pyarrow_convert_dtypes`. Currently it only is used within the `table` module. Also it is only tested within the context of tests targeting misc tables and tables, so this is a kind of integration test approach.

I am not saying that it needs to be unit tested, and I don't know whether it would ever be used outside of the context of the table module. In case that one would want to unit-test it at some later stage, or even only stylistically, one would probably look whether a `@staticmethod` decorator would make sense. However the method depends on `_levels_and_dtypes`, `columns` and `db.schemes` that describe the table in terms of indices, columns and their respective types - so one would have to pass around these, which would make it cumbersome. In other words, the cohesion of `_pyarrow_convert_dtypes` is quite high. Is there a way to decrease this? All of data passed around contain information about indices/columns and how these relate to schemes and dtypes.

I would approve for now as I do not know whether any of these comments will have any consequence codewise.

tests/test_table.py
```diff
@@ -1204,6 +1211,153 @@ def test_map(table, map):
     pd.testing.assert_frame_equal(result, expected)
 
 
+@pytest.mark.parametrize("storage_format", ["csv", "parquet"])
+def test_hash(tmpdir, storage_format):
```
ChristianGeng, line 1215: I wonder whether the tests of hashing would not be better refactored into a test class, with one dedicated test for each assertion?

I am asking because it is hard to read and understand whether the assertions are based on chained transformations of the database or whether they each rely on the saved database.

audformat/core/table.py
ChristianGeng: >> Maybe you put your suggestion on the wrong line?

Ah probably this is it. D205  you get is on blank lines inside the docstring. I had been targeting D202 which is on the blank line between docstring and first line of code of the function. Apologies.

I am using the project's `pyproject.toml`. Maybe the ruff version has diverged?

tests/test_table.py
ChristianGeng: So testing the typing is an addition that comes with asserts from `pd.testing`.
Then I will resolve as this is the stricter test, sth like `assert df1.equals(df2)` is not necessary.

audformat/core/table.py
hagenw: > Maybe the ruff version has diverged?

If you use `pre-commit` to run `ruff`, this shouldn't be the case as it handles installing the correct `ruff` version, e.g.

```bash
$ pre-commit install
$ pre-commit run --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed
codespell................................................................Passed
```

ChristianGeng: I know but sometimes this is impossible - in particular when `python3.10`  as configured in `.pre-commit-config.yaml` is locally not installable - as on my box at the moment.  I would then fall-back to a local ruff check.  I never had this problem, but in future I should probably match  the version that `pre-commit` prescribes.

hagenw: I guess you should ask for a new PC if you cannot install Python 3.10 ;)

hagenw:
> The hash key's value seems kind of generic but I wonder whether it only had been introduced recently, or whether it is changed sometimes?

This is indeed a good point. I created https://github.com/audeering/audformat/pull/436 to propose a better hash solution.

tests/test_table.py
hagenw: I'm not that familiar with using test classes in `pytest`. Is the following what you would have in mind?

```python
@pytest.mark.parametrize("storage_format", ["csv", "parquet"])
class TestHash:
    r"""Test if PARQUET file hash changes with table.

    We store a MD5 sum associated with the dataframe,
    that was used to create the file,
    in the metadata of the PARQUET file.
    Those MD5 sum is supposed to change,
    if any of the table rows, (index) columns changes,
    the data type of the entries changes,
    or the name of a column changes.

    Args:
        tmpdir: tmpdir fixture
        storage_format: storage format of table file

    """

    def db(self, tmpdir, storage_format):
        r"""Create minimal database with scheme and table."""
        self.db_root = audeer.path(tmpdir, "db")
        self.storage_format = storage_format
        self.table_file = audeer.path(self.db_root, f"db.table.{storage_format}")
        db = audformat.Database("mydb")
        db.schemes["int"] = audformat.Scheme("int")
        index = audformat.segmented_index(["f1", "f2"], [0, 1], [1, 2])
        db["table"] = audformat.Table(index)
        db["table"]["column"] = audformat.Column(scheme_id="int")
        db["table"]["column"].set([0, 1])
        db.save(self.db_root, storage_format=self.storage_format)
        return db

    def md5(self) -> str:
        r"""Get MD5 sum for table file."""
        ext = self.storage_format
        if ext == "csv":
            md5 = audeer.md5(self.table_file)
        elif ext == "parquet":
            md5 = parquet.read_schema(self.table_file).metadata[b"hash"].decode()
        return md5

    def test_change_index(self, tmpdir, storage_format):
        r"""Change table index."""
        db = self.db(tmpdir, storage_format)
        md5 = self.md5()
        index = audformat.segmented_index(["f1", "f1"], [0, 1], [1, 2])
        db["table"] = audformat.Table(index)
        db["table"]["column"] = audformat.Column(scheme_id="int")
        db["table"]["column"].set([0, 1])
        db.save(self.db_root, storage_format=self.storage_format)
        assert self.md5() != md5

    def test_change_column_name(self, tmpdir, storage_format):
        r"""Change table column name."""
        db = self.db(tmpdir, storage_format)
        md5 = self.md5()
        index = audformat.segmented_index(["f1", "f2"], [0, 1], [1, 2])
        db["table"] = audformat.Table(index)
        db["table"]["col"] = audformat.Column(scheme_id="int")
        db["table"]["col"].set([0, 1])
        db.save(self.db_root, storage_format=self.storage_format)
        assert self.md5() != md5

    def test_change_column_order(self, tmpdir, storage_format):
        r"""Change order of table columns."""
        db = self.db(tmpdir, storage_format)
        index = audformat.segmented_index(["f1", "f2"], [0, 1], [1, 2])
        db["table"] = audformat.Table(index)
        db["table"]["col1"] = audformat.Column(scheme_id="int")
        db["table"]["col1"].set([0, 1])
        db["table"]["col2"] = audformat.Column(scheme_id="int")
        db["table"]["col2"].set([0, 1])
        db.save(self.db_root, storage_format=self.storage_format)
        md5 = self.md5()
        db["table"] = audformat.Table(index)
        db["table"]["col2"] = audformat.Column(scheme_id="int")
        db["table"]["col2"].set([0, 1])
        db["table"]["col1"] = audformat.Column(scheme_id="int")
        db["table"]["col1"].set([0, 1])
        db.save(self.db_root, storage_format=self.storage_format)
        assert self.md5() != md5

    def test_change_row_order(self, tmpdir, storage_format):
        r"""Change order of table rows."""
        db = self.db(tmpdir, storage_format)
        md5 = self.md5()
        index = audformat.segmented_index(["f2", "f1"], [1, 0], [2, 1])
        db["table"] = audformat.Table(index)
        db["table"]["column"] = audformat.Column(scheme_id="int")
        db["table"]["column"].set([1, 0])
        db.save(self.db_root, storage_format=storage_format)
        assert self.md5() != md5

    def test_change_values(self, tmpdir, storage_format):
        r"""Change table values."""
        db = self.db(tmpdir, storage_format)
        md5 = self.md5()
        index = audformat.segmented_index(["f1", "f2"], [0, 1], [1, 2])
        db["table"] = audformat.Table(index)
        db["table"]["column"] = audformat.Column(scheme_id="int")
        db["table"]["column"].set([1, 0])
        db.save(self.db_root, storage_format=self.storage_format)
        assert self.md5() != md5

    def test_copy_table(self, tmpdir, storage_format):
        r"""Replace table with identical copy."""
        db = self.db(tmpdir, storage_format)
        md5 = self.md5()
        table = db["table"].copy()
        db["table"] = table
        db.save(self.db_root, storage_format=self.storage_format)
        assert self.md5() == md5
```

ChristianGeng: Yes,  test classes in pytest afaik do not do much more than improving readability through isolation. 
Now it becomes clear that each time `db.save`  is called a different tmpdir is setup and that the transactions do not depend on each other - so at least for me - it is much nicer to read: we can quickly see that it is robust or how it behaves against copying, column exchange and so on. 

Once you commit and run the tests the conversation can be resolved - I am anticipating by pressing the "reolve conversation" button right now.

hagenw:
Regarding the complexity of `audformat.Table._pyarrow_convert_dtypes()`. If you have correctly spotted, the conversion can only be done if the table is assigned to a database, and hence has information on the used schemes and columns.
Originally, I thought I might be able to simplify it by adding `index_names`, `index_dtypes`, `column_names`, `column_dtypes` as arguments, but the problem is that it also needs to inspect `self.db.schemes`.

So far, I have not found a better way to restructure the code, but I agree that we should make `audformat.Table._pyarrow_convert_dtypes()` a class method and add unit tests for it. I created https://github.com/audeering/audformat/issues/437 to track this.

tests/test_table.py
hagenw: OK, great, I have pushed my changes.

Issue #420, opened 2024-04-24
utils.read_csv doesn't recognize format


felixbur:
Hi, I want to read in the attached file 
[train.csv](https://github.com/audeering/audformat/files/15092129/train.csv)

with 
```
df = audformat.utils.read_csv(data_file)
```
but get back a Series, no error:
```
type(df)
<class 'pandas.core.series.Series'>
```
it works as expected with pandas.read_csv()

hagenw:
If only one column is present in the data, it returns a series. You can convert to a dataframe with:

```python
y = audformat.utils.read_csv("train.csv")
df = y.to_frame()
```

Or if it should be a dataframe independent of the number of columns:
```python
obj = audformat.utils.read_csv("train.csv")
if isinstance(obj, pd.Series):
    df = obj.to_frame()
else:
    df = obj
```

hagenw:
The current behavior is not written in the text of the corresponding docstring, but indicated by the return types:

![image](https://github.com/audeering/audformat/assets/173624/61660b7d-7d88-4ea8-976e-2684418e36fe)

PR #421, opened 2024-04-24
Add as_dataframe argument to read_csv()

hagenw:
Relates to #420 

This adds the `as_dataframe` argument to `audformat.utils.read_csv()`, to have the option to enforce that always a dataframe is returned. The pull request also enhances the documentation of the default behavior, which returns a series if the CSV file holds one column, and an index if the CSV file contains 0 data columns.

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1261,6 +1261,7 @@ def map_language(language: str) -> str:
 
 def read_csv(
     *args,
+    always_return_dataframe: bool = False,
     **kwargs,
 ) -> typing.Union[pd.Index, pd.Series, pd.DataFrame]:
     r"""Read object from CSV file.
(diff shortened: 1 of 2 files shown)


hagenw:
The name `always_return_dataframe` seems a little bit long to me, do you have some better suggestions we could use instead?

audeerington:
Cool, seems like a useful option to have!

> The name `always_return_dataframe` seems a little bit long to me, do you have some better suggestions we could use instead?

Maybe `return_dataframe` or `as_dataframe`?

felixbur:
i like "as_dataframe"

hagenw:
I changed the argument name to `as_dataframe` and updated the description of the pull request accordingly.

audeerington (approved):

PR #423, opened 2024-04-26
utils.read_csv() now handles start, end in seconds

### LINKED ISSUE #422: audformat.utils.read_csv() converts float time values to nanoseconds
In `audformat.segmented_index()` we convert float and integer values provided for `start` and `end` to `pandas.Timedelta` by treating them as seconds. When reading a CSV file, that contains time values as float/integers in seconds with `audformat.utils.read_csv()` they are converted to `pandas.Timedelta` by treating them as nanoseconds.

I would treat this as a but and change it to always assume float/integer time values represent seconds.

/cc @felixbur

hagenw:
Closes #422 

This adjusts `audformat.utils.read_csv()` to handle time values provided for `start` and `end` values as seconds when provided as float/integer and not as time string, e.g. `00:00:01`. The behavior is now identical to `audformat.segmented_index()`.

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -21,7 +21,6 @@ from audformat.core.index import filewise_index
 from audformat.core.index import is_filewise_index
 from audformat.core.index import is_segmented_index
 from audformat.core.index import segmented_index
-from audformat.core.index import to_timedelta
 from audformat.core.scheme import Scheme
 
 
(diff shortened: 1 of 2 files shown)


audformat/core/utils.py
```diff
@@ -1316,12 +1333,12 @@ def read_csv(
 
     starts = None
     if define.IndexField.START in frame.columns:
-        starts = pd.to_timedelta(frame[define.IndexField.START])
+        starts = frame[define.IndexField.START]
```
hagenw, line 1336: We can drop the call to `pd.to_timedelta()` as the `starts` values are handed to `audformat.segmented_index()`. The same holds for the `ends` values.

audeerington (approved):
Nice, that is also the behavior I would expect from `utils.read_csv()`, so good to have that changed.

Looks good to me!

PR #427, opened 2024-05-15
Fix Database.get() for non-existent schemes

### LINKED ISSUE #426: audformat.Database.get() is not always returning the requested scheme
For the following example, we need the `crema-d` dataset:

```python
import audb

db = audb.load("crema-d", version="1.3.0", only_metadata=True, full_path=False)
```

The dataset has a `sex` scheme, that we can request as additional scheme:

```python
>>> db.get("emotion", additional_schemes=["speaker", "sex"], tables="emotion.categories.test.gold_standard")
                          emotion speaker     sex
file                                             
1048/1048_IEO_NEU_XX.wav  neutral    1048    male
1048/1048_IEO_HAP_LO.wav  neutral    1048    male
1048/1048_IEO_HAP_MD.wav  neutral    1048    male
1048/1048_IEO_HAP_HI.wav  neutral    1048    male
1048/1048_IEO_SAD_LO.wav  neutral    1048    male
...                           ...     ...     ...
1091/1091_WSI_HAP_XX.wav  neutral    1091  female
1091/1091_WSI_SAD_XX.wav  neutral    1091  female
1091/1091_WSI_ANG_XX.wav    anger    1091  female
1091/1091_WSI_FEA_XX.wav  sadness    1091  female
1091/1091_WSI_DIS_XX.wav  neutral    1091  female

[1392 rows x 3 columns]
```

But when requesting the same scheme, but another in addition, we get nothing back:
```python
>>> db.get("emotion", additional_schemes=["speaker", "bla", "sex"], tables="emotion.categories.test.gold_standard")
                          emotion speaker  bla  sex
file                                               
1048/1048_IEO_NEU_XX.wav  neutral    1048  NaN  NaN
1048/1048_IEO_HAP_LO.wav  neutral    1048  NaN  NaN
1048/1048_IEO_HAP_MD.wav  neutral    1048  NaN  NaN
1048/1048_IEO_HAP_HI.wav  neutral    1048  NaN  NaN
1048/1048_IEO_SAD_LO.wav  neutral    1048  NaN  NaN
...                           ...     ...  ...  ...
1091/1091_WSI_HAP_XX.wav  neutral    1091  NaN  NaN
1091/1091_WSI_SAD_XX.wav  neutral    1091  NaN  NaN
1091/1091_WSI_ANG_XX.wav    anger    1091  NaN  NaN
1091/1091_WSI_FEA_XX.wav  sadness    1091  NaN  NaN
1091/1091_WSI_DIS_XX.wav  neutral    1091  NaN  NaN

[1392 rows x 4 columns]
```

hagenw:
Closes #426 

Add two tests that cover #246. The code itself is fixed by removing an `if` statement, that explicitly was setting an requested additional scheme to an empty frame, if the additional scheme requested before was also empty. I guess this was some left over code, that was no longer needed, but did also not fail the tests we had so far.

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -855,16 +855,13 @@ class Database(HeaderBase):
         # --- Append additional schemes
         objs = [obj]
         for scheme in additional_schemes:
-            if len(obj) == 0:
-                obj = empty_frame(scheme)
-            else:
-                obj = self.get(
-                    scheme,
-                    strict=strict,
-                    map=map,
-                    original_column_names=original_column_names,
-                    aggregate_function=aggregate_function,
-                )
+            obj = self.get(
+                scheme,
+                strict=strict,
+                map=map,
+                original_column_names=original_column_names,
+                aggregate_function=aggregate_function,
+            )
             objs.append(obj)
         if len(objs) > 1:
             obj = utils.concat(objs)
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -855,14 +855,19 @@ class Database(HeaderBase):
         # --- Append additional schemes
         objs = [obj]
         for scheme in additional_schemes:
-            obj = self.get(
-                scheme,
-                strict=strict,
-                map=map,
-                original_column_names=original_column_names,
-                aggregate_function=aggregate_function,
-            )
-            objs.append(obj)
+            if len(obj) == 0:
+                # Skip searching for additional schemes,
+                # if main scheme does return empty frame
+                additional_obj = empty_frame(scheme)
+            else:
+                additional_obj = self.get(
+                    scheme,
+                    strict=strict,
+                    map=map,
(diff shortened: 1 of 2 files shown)


audeerington (approved):
Thanks for fixing this!

I would  guess that the original intention of the line is that you don't need to load the data at all to check for additional schemes if you already know that the length of the dataframe containing the main scheme is 0. So maybe this would save some time. But since `obj` is overwritten in the loop, this leads to the bug.

It seems to me that this case wouldn't come up very often, so I think it is fine to just remove the check like this.

hagenw:
Good point.
I think, this is still of interest as it can save execution time. The problem before was, that I used the same variable name for the additional schemes. I updated the code now to use the structure from `main`, but use `additional_obj` as variable name for the additional schemes, instead of `obj`.

audeerington:
Great, then everything looks good to me.

PR #431, opened 2024-05-30
Ensure correct boolean dtype in misc table index

### LINKED ISSUE #430: Bool index in misc table might have wrong dtype
In `audformat` we use nullable pandas dtypes in dataframes representing a table, such as `Int64` and `boolean`.

The index assigned to a misc table can have all possible `pandas` dtypes, and is not associated with an `audformat` scheme. To make it easier for the user (and the developer) we adjust the dtype of the index during assignment to match the `pandas` dtype, that would be used for a table column with the corresponding values. E.g. this means an index that uses `int64` values is adjusted to `Int64`:

```python
>>> import audformat
>>>  import pandas as pd
>>> index = pd.Index([0, 1, 2], name="integers")
>>> index.dtype
dtype('int64')
>>> table = audformat.MiscTable(index)
>>> table.index
Index([0, 1, 2], dtype='Int64', name='integers')
```

In the current `main` branch, the adjustment was not working for the `bool` dtype:
```python
>>> index = pd.Index([True, False], name="bool")
>>> index.dtype
dtype('bool')
>>> table = audformat.MiscTable(index)
>>> table.index
Index([True, False], dtype='bool', name='bool')
```

hagenw:
Closes #430 

This pull request adds extra tests to cover #430 and fixes it by ensuring `bool` dtype is always converted to `"boolean"` dtype.
The fix is done by incorporating the needed conversion into `audformat.core.utils._maybe_convert_int_dtype()` and renaming the function to `audformat.core.utils._maybe_convert_pandas_dtype()`

The second example from #430 now produces the desired result:

```python
>>> import audformat
>>> import pandas as pd
>>> index = pd.Index([True, False], name="bool")
>>> index.dtype
dtype('bool')
>>> table = audformat.MiscTable(index)
>>> table.index
Index([True, False], dtype='boolean', name='bool')
```

Besides the actual changes, I also added or expanded docstrings.
Note: I also added tests for columns of a normal table, even though the issue was not occurring there before.

### DIFF
--- a/audformat/core/common.py
+++ b/audformat/core/common.py
@@ -360,7 +360,14 @@ def to_audformat_dtype(dtype: typing.Union[str, typing.Type]) -> str:
 
 
 def to_pandas_dtype(dtype: str) -> str:
-    r"""Convert audformat to pandas dtype."""
+    r"""Convert audformat to pandas dtype.
+
+    We use ``"Int64"`` instead of ``"int64"``,
+    and ``"boolean"`` instead of ``"bool"``
+    to allow for nullable entries,
+    e.g. ``[0, 2, <NA>]``.
+
+    """
     if dtype == define.DataType.BOOL:
         return "boolean"
     elif dtype == define.DataType.DATE:
--- a/audformat/core/common.py
+++ b/audformat/core/common.py
@@ -367,6 +367,12 @@ def to_pandas_dtype(dtype: str) -> str:
     to allow for nullable entries,
     e.g. ``[0, 2, <NA>]``.
 
+    Args:
+        dtype: audformat dtype
+
+    Returns:
+        pandas dtype
+
     """
     if dtype == define.DataType.BOOL:
         return "boolean"
--- a/audformat/core/common.py
+++ b/audformat/core/common.py
@@ -359,7 +359,7 @@ def to_audformat_dtype(dtype: typing.Union[str, typing.Type]) -> str:
         return define.DataType.OBJECT
 
 
-def to_pandas_dtype(dtype: str) -> str:
+def to_pandas_dtype(dtype: str) -> typing.Optional[str]:
     r"""Convert audformat to pandas dtype.
 
     We use ``"Int64"`` instead of ``"int64"``,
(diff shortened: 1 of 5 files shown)


ChristianGeng (approved):
# Review

The MR improves typing consistency using nullable dtypes not only for int, but consistently now also for bool.

I have run the tests, they all work fine for me. I will approve this, but I have a more general questions, that we had in different form about schemes in `audb.Dependencies` quite recently:

Often the code currently uses if clauses. e.g. in `common.to_pandas_dtype` or class attributes as in `define.DataType` to handle typing information. From what I see more often in other environments (like e.g. sqlalchemy) is that they use dict like structures to do lookup of typing info.

While [this post](<https://stackoverflow.com/questions/15923766/python-dictionary-vs-if-statement-speed>) shows that the disassembly would favor lookup for performance reaseons, this would worry me not so much. However a dict-like implementation appears to be more idiomatic and more readable to me. In addition, lookup apparoaches allow for easier testing.

A minor aspect for me is that there are several implementations with of `test_to_pandas_dtype`. They seem all to be using the underlying `audformat.core.common.to_pandas_dtype` but this has confused me a little.

However these are things that should take too much space here. I will therefore approve without further ado. I can confirm that all the tests pass.

hagenw:
> Often the code currently uses if clauses. e.g. in common.to_pandas_dtype or class attributes as in define.DataType to handle typing information. From what I see more often in other environments (like e.g. sqlalchemy) is that they use dict like structures to do lookup of typing info.

I created https://github.com/audeering/audb/pull/420 as a suggestion how to improve the storage of definitions in `audb` using a dictionary. Maybe you can have a look at it. After we have solved it there we could also update `audformat` in the same way, but I would do it independent of this pull request.

hagenw:
> A minor aspect for me is that there are several implementations with of `test_to_pandas_dtype`. They seem all to be using the underlying `audformat.core.common.to_pandas_dtype` but this has confused me a little.

Are you confused, because they are only testing `audformat.core.common.to_pandas_dtype()` and not `audformat.core.utils._maybe_convert_pandas_dtype()`, or are you refering to something else here?

PR #438, opened 2024-06-21
DOC: add PARQUET tables to data format definition

hagenw:
In https://github.com/audeering/audformat/pull/419 we added support for tables stored as PARQUET files.
Here, we update the data format definition in the documentation accordingly:

image

I checked the rest of the documentation as well, but it seems we only need to update this single paragraph + table.


ChristianGeng (approved):
# Review

This MR updates the documentation reflecting the integration of the parquet format, and can be quickly approved.

PR #441, opened 2024-06-24
DOC: mention explicitly text files as media files

### LINKED ISSUE #439: Document that media files don't have to be audio or video files
In https://github.com/audeering/audb/pull/392 we added support for handling TXT files in `audformat` databases in `audb`.
In `audformat` we didn't needed to change anything, as in theory we can support all file formats.

But the official data format specification does only mention audio and video files as supported media files, e.g.

image

and 

image

I think we should reword that, and explain that also text or other files can be media files.

hagenw:
Closes #439 

Rephrases the README and data format specifications to better indicate that media files are not restricted in its file type.

In addition, it explicitly names text files as potential media files. I also state that a segmented table can be used, if applicable, as some file types might not have a direct connection to a timeline. E.g. you can create a segmented table for text files, but it might be meaningless as no other application might be able to use that information.

**README**

image

**Format specifications -- Introduction**

image

**Format specification -- Database**

image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,7 +7,7 @@ audformat
 Specification and reference implementation of **audformat**.
 
 audformat stores media data,
-such as audio or video,
+such as audio, video, or text
 together with corresponding annotations
 in a pre-defined way.
 This makes it easy to combine or replace databases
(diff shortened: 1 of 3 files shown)


ChristianGeng (approved):
This updates the documentation to reflect that it generalizes "audio" data to  more general "media" data. 

The only concern that I have is whether even "media" is general enough, or whether even a more general term like "sensor data" should be used, as  future releases might need to expand to other sensing modalities with fixed sampling rate - or even rare events that are not sampled.

Still, the description is valid for now and therefore approval will be given.

hagenw:
Thanks for pointing out the concern regarding "Media file". I created https://github.com/audeering/audformat/issues/444 to track that, as I also think this would require more rewriting of the documentation, and should be handled independent of this pull request.

PR #443, opened 2024-06-24
DOC: utils.hash() is not influenced by names

### LINKED ISSUE #434: audformat.utils.hash() does not consider column names
```python
>>> import audformat
>>> import pandas as pd
>>> df = pd.DataFrame([1], columns=["a"])
>>> audformat.utils.hash(df)
'-3807690387551515836'
>>> df = pd.DataFrame([1], columns=["b"])
>>> audformat.utils.hash(df)
'-3807690387551515836'
```

The behavior is expected as the underlying `pd.util.hash_pandas_object()` does also not take column names into account: https://github.com/pandas-dev/pandas/issues/46705.
I'm not sure if we should change it, but at least we should update the documentation explaining that column names are not considered.

hagenw:
Closes #434 

`audformat.utils.hash()` does only consider values, not level/column names.
To make this more obvious, I added a sentence to the docs:

image

I also added tests for the expected behavior.

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -669,7 +669,8 @@ def hash(
 
     Objects with the same elements
     produce the same hash string
-    independent of the ordering of the elements.
+    independent of the ordering of the elements,
+    and potential level or column names.
 
     Args:
         obj: object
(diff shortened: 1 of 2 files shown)


ChristianGeng (approved):
# Review

This MR extends testing of the hashing approach for asserting invariance of the hash 

- when index name changes for series and data frames
- column names change for data frames.

The resulting hashes are hardcoded which will therefore also detect changes in the unlikely case that the underlying numpy hashing implementation is changed.

This can be counted as approved.

PR #445, opened 2024-06-25
Release 1.2.0

hagenw:
image


hagenw:
For me the the tests also work locally when running them with `pytest` or `pytest test`. If you would like to debug this further, you could create a pull request in which you change the command that calls the tests in the CI job and see if they fail as well.

PR #450, opened 2024-07-11
Load csv tables with pandas if pyarrow fails

### LINKED ISSUE #449: Parsing of CSV files with pyarrow.csv.read_csv() can fail
We identified already two databases, for which loading of the tables failed when parsing the CSV file with `pyarrow` can fail.
One is `era-dialog`, the other one `austrian-parliament`:

```python
>>> db = audb.load("austrian-parliament", version="1.1.0", only_metadata=True)
...
File ~/.envs/analysis/lib/python3.10/site-packages/audformat/core/table.py:518, in Base.load(self, path)
    516     self._load_parquet(parquet_file)
    517 else:
--> 518     self._load_csv(csv_file)

File ~/.envs/analysis/lib/python3.10/site-packages/audformat/core/table.py:868, in Base._load_csv(self, path)
    866 levels = list(self._levels_and_dtypes.keys())
    867 columns = list(self.columns.keys())
--> 868 table = csv.read_csv(
    869     path,
    870     read_options=csv.ReadOptions(
    871         column_names=levels + columns,
    872         skip_rows=1,
    873     ),
    874     convert_options=csv.ConvertOptions(
    875         column_types=self._pyarrow_csv_schema(),
    876         strings_can_be_null=True,
    877     ),
    878 )
    879 df = self._pyarrow_table_to_dataframe(table, from_csv=True)
    881 self._df = df

File ~/.envs/analysis/lib/python3.10/site-packages/pyarrow/_csv.pyx:1261, in pyarrow._csv.read_csv()

File ~/.envs/analysis/lib/python3.10/site-packages/pyarrow/_csv.pyx:1270, in pyarrow._csv.read_csv()

File ~/.envs/analysis/lib/python3.10/site-packages/pyarrow/error.pxi:154, in pyarrow.lib.pyarrow_internal_check_status()

File ~/.envs/analysis/lib/python3.10/site-packages/pyarrow/error.pxi:91, in pyarrow.lib.check_status()

ArrowInvalid: CSV parse error: Expected 5 columns, got 9: data/693dc911-3248-47ba-b654-b867688845b0.wav,0 days 00:02:04.120000,0 days 00:02:09.460000,Aloi ...
```

hagenw:
Closes #449 

Unfortunately, loading csv files with `pyarrow.csv.read_csv()` as introduced in https://github.com/audeering/audformat/pull/419 is not as tolerant to malformed csv files as `pandas.read_csv()`. I have identified so far three cases in which loading of a csv file might fail (two of them are listed in #449):
1. Loading a csv file can fail, if the csv file contains more columns, as mentioned in the header of a database
2. Loading a csv file can also fail, if it is very long and contains a lot of special characters, like `"`, `""`, `,`. I did not added a test for it, because it turns out that the syntax of the csv file is correct, and it works when splitting the file into smaller ones.
3. Loading of a csv file can fail, if it contain some offsets in there `date` values, e.g. `+00:00`, which is the case for some of our older datasets.

As `pyarrow.csv.read_csv()` cannot be easily extended to handle those cases, I use now a `try`-`except` statement, that falls back to loading the file with `pandas.read_csv()`. This is very unfortunate as it means when implementing a new feature (e.g. streaming) it needs to be implemented for both cases. But I don't know a better solution at the moment.

In principle, we could solve 1. and 3. by updating the databases, but you will still no longer be able to load old versions then, which is not acceptable. How we could solve 2. otherwise, I don't know.

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -1159,7 +1159,9 @@ def test_load_broken_csv(tmpdir):
     """
     build_dir = audeer.mkdir(tmpdir, "build")
 
-    # Create database with single table and column
+    # Create database with hidden columns
+    # that are stored in csv,
+    # but not in the header of the table.
     #
     # Ensure:
     #
@@ -1167,7 +1169,9 @@ def test_load_broken_csv(tmpdir):
     # * the columns use schemes with time and date data types
     # * at least one column has no scheme
     #
-    # as those cases needed special care with csv files
+    # as those cases needed special care with csv files,
+    # before switching to use pyarrow.csv.read_csv()
+    # in https://github.com/audeering/audformat/pull/419
     #
     db = audformat.Database("mydb")
     db.schemes["date"] = audformat.Scheme("date")
@@ -1181,7 +1185,6 @@ def test_load_broken_csv(tmpdir):
     db["table"]["no-scheme"].set(["label"])
     db["empty-table"] = audformat.Table(audformat.filewise_index())
     db["empty-table"]["column"] = audformat.Column()
-
     # Add a hidden column to the table dataframes,
     # without adding it to the table header
     db["table"].df["hidden"] = ["hidden"]
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -1144,9 +1144,9 @@ def test_load(tmpdir):
 
 
 def test_load_broken_csv(tmpdir):
-    r"""Test loading of malformed CSV files.
+    r"""Test loading of malformed csv files.
 
-    If csv files contain a lot of special character,
+    If csv files contain a lot of special characters,
     or a different number of columns,
     than specified in the database header,
     loading of them should not fail.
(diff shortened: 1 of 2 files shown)


tests/test_table.py
```diff
@@ -1143,6 +1143,58 @@ def test_load(tmpdir):
...
+    db["table"]["no-scheme"].set(["label"])
+    db["empty-table"] = audformat.Table(audformat.filewise_index())
+    db["empty-table"]["column"] = audformat.Column()
+
+    # Add a hidden column to the table dataframes,
+    # without adding it to the table header
+    db["table"].df["hidden"] = ["hidden"]
+    db["empty-table"].df["hidden"] = []
+
+    db.save(build_dir, storage_format="csv")
```
ChristianGeng, line 1190: The setup of the test data and putting it into testable state consumes most of the lines implemented and makes it a little harder to read - I mean that the purpose of the test is pretty obvious so maybe this is not crucial at all. 
 
Still I find it stilistically nicer to separate test data creation from asserting .

The pytest way of doing so would very likely to implement (yet another) fixture.
One could though use a different, less pytest-like idiom, test classes containing setUp methods from the built-iin `unittest` module. Would something like [this](https://hands-on.cloud/python-unit-tests/#mocking-classes-in-python-unit-tests:~:text=for%20each%20test.-,Here%20is%20an%20example%3A,-import%20unittest%0Aclass)  not read nicely in the current situation? I mean it looks quite imperative but still separates setup and destruction of test data - like all xunit frameworks do.

hagenw: What I can try to do, is to create a test class and a method to set up the data, and one that does the test.

I will try to do that.

hagenw: OK, sorry, I was too fast with my comment in https://github.com/audeering/audformat/pull/450#discussion_r1675799599 as you can have the data definition and the actual test as two methods of the same class, and hence they are defined together and are easy to read.

I adapted the test to now have the following structure:

```python
class TestLoadBrokenCsv:

    def database_with_hidden_columns(self) -> audformat.Database:
        # create database

    def test_load_broken_csv(self, tmpdir):
        db = self.database_with_hidden_columns()
        # run the test
```

ChristianGeng: >> The creation of the data is the most important part of the test, and the data is meaningless in another test.

>> When creating a fixture, the code for the table is positioned at another place, e.g. at the top of the file, and a user needs to switch between the two positions in order to read it.

I think you could just put it right before the test itself. 

>> But I have to admit, I was never a big fan of object based programming and unittest, so maybe I just don't understand how you would like to have it ;)

In principle what you do is very similar. You could rename the method that creates the data as a @classmethod decorated `setup`  method and you would be almost done (you could use a `teardown` method to delete `tmpdir`
 ). 

But in this particular case this really does not matter, the way you implement it is good for the readability due to data creation and asserting being separated into two methods.   Using a class rather than a function imo also improves readability - you know that the data setup is used in this particular class.  xunit users often would setup and teardown methods, and then go one step further and have each assertion in a single method `def`.  

But for me this is definitively good enough. Not though that due to temporary failures, I was able to run the tests, but I could not recap the data examples from the issue description 

Still I think this is ready to be merged and I will approve.

ChristianGeng: Resolving this thread also.

ChristianGeng: I again cannot find the location where approval is given. But you can go ahead.

PR #452, opened 2024-07-12
Store tables per default as parquet files

hagenw:
As we now support storing tables as parquet files in `audb>=1.8.0`, and have published the first databases, I would propose to switch to parquet for storing tables in the next `audformat` release.

This pull request changes the default values of the `storage_format` argument in `audformat.Database.save()` and `audformat.Table.save()` to `"parquet"`.

image

image

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -970,7 +970,7 @@ class Database(HeaderBase):
         *,
         name: str = "db",
         indent: int = 2,
-        storage_format: str = define.TableStorageFormat.CSV,
+        storage_format: str = define.TableStorageFormat.PARQUET,
         update_other_formats: bool = True,
         header_only: bool = False,
         num_workers: typing.Optional[int] = 1,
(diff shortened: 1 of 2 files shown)


hagenw:
Seems our tests were already well prepared for this step, as I didn't had to change anything.

ChristianGeng (approved):
This basically changes the derfault table storage format to the newer parquet in two locations:

- the datatabase level
- at the level of the table

I do not know where the `Table.save` is used other than by `Database.save` so I cannot judge whether this is functionally necessary: grepping the source code of this repo and the audb repo suggests that this is never done. 

So my understanding is that one could get away without defaulting to parquet in `Table` - still it is better at the level of API consistency.

I will approve this right away.

hagenw:
> So my understanding is that one could get away without defaulting to parquet in Table - still it is better at the level of API consistency.

Exactly, we don't have to change it in `audformat.Table.save()`, but I also thought that it would be more consistent this way.

PR #453, opened 2024-07-15
Add cgeng to authors

hagenw:
Add @ChristianGeng to authors in `LICENSE` and `pyproject.toml`, and align order of authors of both files.


ChristianGeng (approved):
Approving.

PR #454, opened 2024-07-15
Add strict argument to audformat.utils.hash()

### LINKED ISSUE #447: Provide table dataframe hashing in public API?
In https://github.com/audeering/audformat/pull/419 we introduced storing of tables as parquet files, and created a custom hashing method for it as `audformat.utils.hash()` does not consider column names, and does not provide consistent hashes across different `pandas` versions.

The steps of the hashing, that could be added to a public function, are:

https://github.com/audeering/audformat/blob/c1328079be136a1801759891d851ca03953f4dad/audformat/core/table.py#L1137-L1142

The only downside might be that we have already `audformat.utils.hash()`, which is meant for index, series, and dataframe at the moment. In our own tools, we only use the hashing of index. So the question arises how we should name the new hash function, and how it should be positioned with regards to `audformat.utils.hash()`.

hagenw:
Closes #447

Adds the `strict` argument to `audformat.utils.hash()`.
If ``True`` it will take the order of rows and the names of index levels and columns into account. The returned hash is then identical to the one attached to parquet table files.

A hashing variant that takes into account the order of rows, name of columns and levels, and the data types was needed for calculating a hash for parquet tables as introduced in https://github.com/audeering/audformat/pull/419. Before, this was handled by a private method, independent of `audformat.utils.hash()`.

image

### DIFF
--- a/audformat/core/table.py
+++ b/audformat/core/table.py
@@ -1,7 +1,6 @@
 from __future__ import annotations  # allow typing without string
 
 import copy
-import hashlib
 import os
 import pickle
 import typing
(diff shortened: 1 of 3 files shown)


ChristianGeng (approved):
Apiwise I  think that having only one `utils.hash` as before and distinguishing calculations based on the `include_order_and_names` kwarg is nice. 

I  cannot do a technical review of the md5 algorithm though, so I am not attempting.
The test coverage is fairly extensive. This will help a lot in the discovery of implementation changes on the pandas side, or even in the unlikely case that sth changes in `hashlib.md5`. 

I would not be passionate about the name of the keyword argument, be it "strict" or "include_order_and_names" as long as the motivation becomes clear.  The code and the issue description contain already links to the pandas issue where this was discussed.  I wonder whether links should even be extended to point to the audformat issues that led to the implementation as is (the pyarrow metdatata has, effectively clarifying why this was needed)?

 
Apart from that I think that this is good and can be approved.

hagenw:
I added a reference to #419 in the description and renamed the argument to `strict`.

PR #455, opened 2024-07-15
Release 1.3.0

hagenw:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ and this project adheres to `Semantic Versioning`_.
 Version 1.3.0 (2024-07-15)
 --------------------------
 
-* Added: ``include_order_and_names`` argument
+* Added: ``strict`` argument
   to ``audformat.utils.hash()``.
   If set to ``True``,
   the order of the data,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
-Version 1.3.0 (2024-07-17)
+Version 1.3.0 (2024-07-18)
 --------------------------
 
 * Added: ``strict`` argument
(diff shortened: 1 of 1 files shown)


hagenw:
This is a breaking change release as code that expects a table to be stored as a csv file will break. But as around 99% of use cases will not directly try to access the csv files, but use `audb` and `audformat` to access tables, I think it should be fine to go with version `1.3.0` instead of `2.0.0`.

CHANGELOG.rst
```diff
@@ -7,6 +7,25 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 1.3.0 (2024-07-17)
+--------------------------
+
+* Added: ``strict`` argument
+  to ``audformat.utils.hash()``.
```
ChristianGeng, line 14: This has a full stop at the end of the sentence. The others do not. What is convention that we are using here?

ChristianGeng (approved):
This looks good to me content-wise. 

I have one question about interpuntuation, but that is not a blocker for giving approval.

CHANGELOG.rst
hagenw: The convention is to not use a full stop at the last sentence in an entry. But for earlier ones we use a full stops:
```
* Added: ``strict`` argument
  to ``audformat.utils.hash()``.
  If set to ``True``,
  the order of the data,
  and its level/column names
  are taken into account
  when calculating the hash
```
As you can see the last `hash` has no full stop.

ChristianGeng:
> This is a breaking change release as code that expects a table to be stored as a csv file will break. But as around 99% of use cases will not directly try to access the csv files, but use `audb` and `audformat` to access tables, I think it should be fine to go with version `1.3.0` instead of `2.0.0`.

`audformat.Database.save` and `audformat.Table.save` are part of the public API and documented. But the gross usage pattern stays intact. So going without a major release is fine.

PR #458, opened 2024-09-16
Release 1.3.1

hagenw:
We had a nice contribution in https://github.com/audeering/audformat/pull/456, which changed the ISO639 package we are using to a better maintained one. As this also fixes the handling of `audformat` by `poetry` I would vote to do an immediate bug fix release.

image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 1.3.1 (2024-09-16)
+--------------------------
+
+* Changed: replace not maintained ``iso-639`` dependency
+  with ``iso639-lang``
+* Fixed: ensure ``poetry`` can manage ``audformat``
+
+
 Version 1.3.0 (2024-07-18)
 --------------------------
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 1.3.1 (2024-09-16)
+--------------------------
+
+* Changed: replace not maintained ``iso-639`` dependency
```
ChristianGeng, line 13:
```suggestion
* Changed: replace unmaintained ``iso-639`` dependency
```

ChristianGeng (approved):
The package used until now is  deprecated with the latest release 8 years old:

https://pypi.org/project/iso639/#history

My basic understanding is that the proposed package is a good choice for replacement. There would be another candidate `python-iso639`.

Interestingly both `python-iso639` and the one we are going for ` iso639-lang`  seem to be releasing  at almost the same dates:

https://pypi.org/project/iso639-lang/#history
https://pypi.org/project/python-iso639/#history

This to me suggests that both are tracking changes in the `iso-639` definitions quite closely and make for good replacements.

For a review I simply verfified that tests are passing. 


Therefore I am approving this.

PR #459, opened 2024-09-19
DOC: add link to ISE 639-3 definition

hagenw:
This expands the docstring of `audformat.utils.map_language()` to contain additional information on ISO 639-3, and provides a link to a list of all included language codes.

image

### DIFF
--- a/audformat/core/utils.py
+++ b/audformat/core/utils.py
@@ -1273,7 +1273,7 @@ def map_language(language: str) -> str:
 
     `ISO 639-3`_ is an international standard for language codes.
     It defines three-letter codes for identifying languages,
-    with an aim to cover all known natural languages.
+    with the aim to cover all known languages.
 
     .. _ISO 639-3: https://iso639-3.sil.org/code_tables/639/data/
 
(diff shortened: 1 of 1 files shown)


audformat/core/utils.py
```diff
@@ -1271,6 +1271,12 @@ def map_file_path(
 def map_language(language: str) -> str:
     r"""Map language to ISO 639-3.
 
+    `ISO 639-3`_ is an international standard for language codes.
+    It defines three-letter codes for identifying languages,
+    with an aim to cover all known natural languages.
```
ChristianGeng, line 1276:
```suggestion
    with the aim to cover all known languages.
```

ChristianGeng (approved):
I will approve this.

I extended the scope of the description to omit the "natural" languages. 
I queried the catalog for examples that do contradict the "natural" limitation.
For example I have found "constructed" languages [here](https://iso639-3.sil.org/code_tables/639).

I found for example epo/Esperanto as an example of a constructed language.
There are other interesting cases: for example Akan is a "macrolanguage" in the iso639
terminology.  When looking for Latin I found it too wtih "Historical"  as its "Language Type"

hagenw:
Interesting, thanks for checking. I just copied "natural" from the Wikipedia description ;)

PR #467, opened 2025-05-13
Fix Database.update() for misc tables

### LINKED ISSUE #466: MiscTable is not handled correctly by `audformat.Database.update()`
When combining to databases with a misc table, the misc table should be combined in the end as well.

Minimal example:

```python
import audformat
import pandas as pd


db1 = audformat.Database("mydb")
db1.schemes["answer1"] = audformat.Scheme("str")
db1["sessions"] = audformat.MiscTable(pd.Index(["a"], name="session"))
db1["sessions"]["prompt_1"] = audformat.Column(scheme_id="answer1")
db1["sessions"]["prompt_1"].set(["response1"])

db2 = audformat.Database("mydb")
db2.schemes["answer1"] = audformat.Scheme("str")
db2.schemes["answer2"] = audformat.Scheme("str")
db2["sessions"] = audformat.MiscTable(pd.Index(["b"], name="session"))
db2["sessions"]["prompt_1"] = audformat.Column(scheme_id="answer1")
db2["sessions"]["prompt_2"] = audformat.Column(scheme_id="answer2")
db2["sessions"]["prompt_1"].set(["response1"])
db2["sessions"]["prompt_2"].set(["response2"])

db1.update(db2)
```

Expected output:

```python
>>> db1["sessions"].df
          prompt_1   prompt_2
session                      
a        response1       <NA>
b        response1  response2
```

But instead we get:

```python
>>> db1["sessions"].df
          prompt_1   prompt_2
session                      
b        response1  response2
```

---

This does not happen for a filewise index:

```python
import audformat
import pandas as pd


db1 = audformat.Database("mydb")
db1.schemes["answer1"] = audformat.Scheme("str")
db1["sessions"] = audformat.Table(audformat.filewise_index(["a"]))
db1["sessions"]["prompt_1"] = audformat.Column(scheme_id="answer1")
db1["sessions"]["prompt_1"].set(["response1"])

db2 = audformat.Database("mydb")
db2.schemes["answer1"] = audformat.Scheme("str")
db2.schemes["answer2"] = audformat.Scheme("str")
db2["sessions"] = audformat.Table(audformat.filewise_index(["b"]))
db2["sessions"]["prompt_1"] = audformat.Column(scheme_id="answer1")
db2["sessions"]["prompt_2"] = audformat.Column(scheme_id="answer2")
db2["sessions"]["prompt_1"].set(["response1"])
db2["sessions"]["prompt_2"].set(["response2"])

db1.update(db2)
```

There we get

```python
>>> db1["sessions"].df
       prompt_1   prompt_2
file                      
a     response1       <NA>
b     response1  response2
```

hagenw:
Closes #466 

Fixes `audformat.Database.update()` for the case of misc tables. The error was not covered by our previous test. I'm still not completely sure why, but to save time I added a new test for the particular problem.

The error was due to us looking only at `self.tables` for existing tables in the database that should be updated. I changed it to `list(self)` to include misc tables as well.

## Summary by Sourcery

Fix misc table handling in Database.update by including them in the update loop and add a corresponding test.

Bug Fixes:
- Support merging miscellaneous tables with new columns in Database.update.

Tests:
- Add test_update_misc_table to verify merging of misc tables when new columns are introduced.

### DIFF
--- a/audformat/core/database.py
+++ b/audformat/core/database.py
@@ -1245,7 +1245,7 @@ class Database(HeaderBase):
         for other in others:
             for table_id in list(other.misc_tables) + list(other.tables):
                 table = other[table_id]
-                if table_id in self.tables:
+                if table_id in list(self):
                     self[table_id].update(table, overwrite=overwrite)
                 else:
                     self[table_id] = table.copy()
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -970,7 +970,8 @@ def test_update_misc_table():
 
     This addresses the particular case
     of adding a new column
-    to the table,
+    to a misc table
+    in a database having only a single table,
     which was failing as described in
     https://github.com/audeering/audformat/issues/466
 
@@ -992,10 +993,8 @@ def test_update_misc_table():
     db2["sessions"]["prompt_2"].set(["response2"])
     df2 = db2["sessions"].df.copy()
 
-    df_expected = pd.concat([df2, df1])
+    df_expected = pd.concat([df1, df2])
 
     db1.update(db2)
-    print(f'{db1["sessions"].df=}')
-    print(f'{db2["sessions"].df=}')
-    print(f"{df_expected=}")
     pd.testing.assert_frame_equal(db1["sessions"].df, df_expected)
+    pd.testing.assert_frame_equal(db2["sessions"].df, df2)
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -963,3 +963,39 @@ def test_update(tmpdir):
         db.update(other1, overwrite=True, copy_attachments=True)
     with pytest.raises(RuntimeError):
         db.update(other1, overwrite=True, copy_media=True)
+
+
+def test_update_misc_table():
+    """Tests updating databases with misc tables.
+
+    This addresses the particular case
+    of adding a new column
+    to the table,
+    which was failing as described in
+    https://github.com/audeering/audformat/issues/466
+
+    """
+    db1 = audformat.Database("mydb")
+    db1.schemes["answer1"] = audformat.Scheme("str")
+    db1["sessions"] = audformat.MiscTable(pd.Index(["a"], name="session"))
+    db1["sessions"]["prompt_1"] = audformat.Column(scheme_id="answer1")
+    db1["sessions"]["prompt_1"].set(["response1"])
+    df1 = db1["sessions"].df.copy()
+
+    db2 = audformat.Database("mydb")
+    db2.schemes["answer1"] = audformat.Scheme("str")
+    db2.schemes["answer2"] = audformat.Scheme("str")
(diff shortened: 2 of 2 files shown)


hagenw:
One thing I realized when debugging the existing test is that `update()` can also change the content of the table which is used for to update, which I thought should not be the case, e.g. [at the beginning](https://github.com/audeering/audformat/blob/2872a9a114ac7574ea3ef5d2c2a2d71d9be05a2f/tests/test_database.py#L824) we have:

```python
db["misc"].df=        float                                                                                                                                                   idx                                                                                                                                                                           
a    0.137095                                                                          
b    0.783854                                                                          
other1["misc"].df=        float labels                                                 
idx                                                                                    
b    0.281209      c                                                                   
c    0.875555      b
```

But [after the first failed attempt](https://github.com/audeering/audformat/blob/2872a9a114ac7574ea3ef5d2c2a2d71d9be05a2f/tests/test_database.py#L829) we have:

```
db["misc"].df=        float              
idx          
b    0.783854                                                                          
c         NaN                                                                                                                                                                 
a    0.137095                                                                          
other1["misc"].df=        float labels                                                                                                                                        
idx                                                                                                                                                                           
b    0.281209      c
c    0.875555      b                                                                   
a         NaN    NaN
```

which means `other1["misc"].df` did also change.

But this is most likely another issue and not related to what we try to fix here.

frankenjoe:
> One thing I realized when debugging the existing test is that `update()` can also change the content of the table which is used for to update, which I thought should not be the case

I had a look and I found this line [here](https://github.com/audeering/audformat/blob/2872a9a114ac7574ea3ef5d2c2a2d71d9be05a2f/audformat/core/database.py#L1211). Here we see that the index of the table that is used to updated is indeed extended. I agree that this is a somehow unexpected behavior. I created https://github.com/audeering/audformat/issues/469.

frankenjoe:
> The error was not covered by our previous test. I'm still not completely sure why, but to save time I added a new test for the particular problem.

The test function that currently covers `Database.update()` is already awfully long. So it's anyway a good idea to add a new specific test.

frankenjoe:
Since the suggested change fixes the issue I will merge now.

hagenw:
> > The error was not covered by our previous test. I'm still not completely sure why, but to save time I added a new test for the particular problem.
> 
> The test function that currently covers `Database.update()` is already awfully long. So it's anyway a good idea to add a new specific test.

Yes, during the last year I also figured out that there are better ways to write long tests that depend on a status of an object like a database. One way is to use a class which ensures that the status of the database is the same for each of the single tests. One example can be found at https://github.com/audeering/audbcards/blob/a6a5cc1b356633965c1e39a1200cb61a5aae7c0b/tests/test_dataset.py#L507-L588.
But it would be indeed some effort to rewrite our tests in `audformat` accordingly.

PR #488, opened 2026-01-22
Fix to_segmented_index() for pandas 3.0

### LINKED ISSUE #487: Pandas 3.0.0 breaking changes
`pandas` released version 3.0.0, which introduces some breaking changes we need to address here.

## Timedelta/datetime updates

`pandas` changed the [default unit of timedelta and datetime entries](https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#datetime-timedelta-resolution-inference) from nanoseconds to a resolution that matches the given input precision. We want to stay with nanosecond resolution in all cases.

## String updates

`pandas` introduces a [new default string type](https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#dedicated-string-data-type-by-default) (`str`/`<StringDtype(na_value=nan)>`), which replaces `object` as default.
Unfortunately, the new string type is different from `string`/`<StringDtype(na_value=<NA>)>` as it uses a different value to represent missing values.

We should update the code here to ensure we get the same results wit pandas <3.0 and pandas >=3.0 by using `string`/`<StringDtype(na_value=<NA>)>` and `object` for categorical string labels.

hagenw:
Closes #487 

This add tests and fixes for the new `pandas==3.0.0` `timdelta[s]` instead of `timedelta[ns]` default.

This required the following fixes:

| Location | Fix |
| --- | --- |
| `audformat.utils.to_segmented_index()` | Convert ends to `timedelta64[ns]` before iloc assignment |
| `audformat.utils.union()` | Normalize timedelta dtypes to `timedelta64[ns]` in all code paths |
| `audformat.utils.intersect()` | Normalize timedelta dtypes to `timedelta64[ns]` |
| `audformat.utils.set_index_dtypes()` | Add `.astype(dtype)` after `pd.to_timedelta()` for empty levels |
| `audformat.segmented_index()` | Call `set_index_dtypes` to ensure `timedelta64[ns]` |
| `audformat.testing.add_table()` | Remove unnecessary `pd.to_timedelta()` call |
| `audformat.utils.hash()` | Enforce `object` dtype for string columns to get same hash under Python 3.14 |

## Summary by Sourcery

Ensure segmented index duration handling preserves sub-second precision with pandas 3.0 and later.

Bug Fixes:
- Fix to_segmented_index to always use nanosecond timedelta precision so assigning high-precision durations no longer raises type or FutureWarning errors with pandas 3.0.

Tests:
- Add regression test verifying to_segmented_index correctly handles sub-second duration values when the index uses second-level timedelta precision and that no FutureWarning is raised.

### DIFF
--- a/tests/test_database_get.py
+++ b/tests/test_database_get.py
@@ -511,7 +511,7 @@ def wrong_scheme_labels_db(tmpdir):
                     ]
                 ),
                 dtype=pd.CategoricalDtype(
-                    ["w1", "w2", "w3"],
+                    pd.Index(["w1", "w2", "w3"], dtype="object"),
                     ordered=False,
                 ),
             ),
@@ -603,7 +603,7 @@ def wrong_scheme_labels_db(tmpdir):
                             [0.2, 0.2, 0.5, 0.7],
                         ),
                         dtype=pd.CategoricalDtype(
-                            ["s1", "s2", "s3"],
+                            pd.Index(["s1", "s2", "s3"], dtype="object"),
                             ordered=False,
                         ),
                         name="speaker",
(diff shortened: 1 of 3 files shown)


hagenw:
We have now too many changes here, and introduced a lot of new issues that were not present before, so we should maybe better target the updates step by step in several pull requests.

Issue #529, opened 2026-04-10
audformat.segmented_index() not working as expected for pandas>=3.0


audeerington:
Creating a segmented index does not always work as expected with `pandas>=3.0`:

```python
import pandas as pd

import audformat


x = pd.Series(
    index=audformat.segmented_index(["file.wav"] * 2, [1.0, 1.01], [2.0, 3.5]),
    data=["a", "b"],
)
print(x)

```

Results in 
```
file      start                      end                      
file.wav  0 days 00:00:00.000000001  0 days 00:00:00.000000002    a
          0 days 00:00:01.010000     0 days 00:00:03.500000       b
dtype: str
```

It seems that it is due to the behavior of `pd.to_timedelta()`, e.g.:

```python
starts = pd.to_timedelta([1.0, 1.01], unit="s")
print(starts)
```
```
TimedeltaIndex(['0 days 00:00:00.000000001', '0 days 00:00:01.010000'], dtype='timedelta64[ns]', freq=None)
```

Whereas using a numpy array rather than a list seems to work fine:

```python
import numpy as np
starts = pd.to_timedelta(np.array([1.0, 1.01]), unit="s")
print(starts)
```
```
TimedeltaIndex(['0 days 00:00:01', '0 days 00:00:01.010000'], dtype='timedelta64[ns]', freq=None)
```

@hagenw

hagenw:
This looks like a upstream bug to me, I created https://github.com/pandas-dev/pandas/issues/65150

hagenw:
@audeerington as a first step could you maybe create a pull request that adds a failing test for this.

Then we wait how they react at `pandas` and wait for a new `pandas` release, or fix it ourselves.

hagenw:
They started to work on a fix at https://github.com/pandas-dev/pandas/pull/65170 and the last release was only 2 weeks ago, so I would assume that the fix is part of `pandas` 3.0.3. Hence, I would propose as a solution to change the dependencies on `pandas` from

```
'pandas >=2.1.0',  # for pyarrow -> timedelta conversion
```

to

```
# >=2.1.0 for pyarrow -> timedelta conversion
# !=3.0.0, !=3.0.1, !=3.0.2 for pandas timedelta issue
# (https://github.com/audeering/audformat/issues/529)
'pandas >=2.1.0, !=3.0.0, !=3.0.1, !=3.0.2',
```

audeerington:
Sounds good to me!

PR #530, opened 2026-04-13
TST: check total seconds of segmented index

audeerington:
This adds a test that checks whether the total seconds of the start and end levels of an index created with `audformat.segmented_index()` are as expected. (See https://github.com/audeering/audformat/issues/529)

## Summary by Sourcery

Tests:
- Add parametrized tests ensuring segmented_index start and end levels match expected total seconds for various floating-point inputs.

### DIFF
--- a/tests/test_index.py
+++ b/tests/test_index.py
@@ -310,7 +310,10 @@ def test_create_segmented_index(files, starts, ends):
     ],
 )
 def test_segmented_index_dtype(files, starts, ends):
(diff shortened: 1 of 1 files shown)


hagenw:
Looks good.

I tested locally, and it passes when using `pandas<3.0.0`.

I would propose we merge this failing test first and afterwards implement a solution for it.

hagenw (approved):

PR #531, opened 2026-04-13
Exclude pandas 3.0.0-3.0.2 from dependencies

### LINKED ISSUE #529: audformat.segmented_index() not working as expected for pandas>=3.0

hagenw:
Closes https://github.com/audeering/audformat/issues/529

Exclude `pandas` 3.0.0, 3.0.1, and 3.0.2 from dependencies due to https://github.com/pandas-dev/pandas/issues/65150.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,10 @@ dependencies = [
     'iso639-lang',
     'iso3166',
     'oyaml',
-    'pandas >=2.1.0',  # for pyarrow -> timedelta conversion
+    # >=2.1.0 for pyarrow -> timedelta conversion
+    # !=3.0.0, !=3.0.1, !=3.0.2 for pandas timedelta issue
+    # (https://github.com/audeering/audformat/issues/529)
+    'pandas >=2.1.0, !=3.0.0, !=3.0.1, !=3.0.2',
     'pyarrow >=10.0.1',  # for pyarrow strings in pandas
     'pyyaml >=5.4.1',
 ]


hagenw:
The downside of the current solution is that we cannot use `pandas` 3.0 until 3.0.3 is released.

An alternative would be to add a workaround (e.g. converting `starts` and `ends` to `numpy.array`).

I'm not sure yet what is the better approach.

hagenw:
The downside of the `numpy` workaround is that there might be other places in the code of `audformat` that are affected by the `pandas` issue and would not be handled by it.

hagenw:
But having in mind how important timedelta values are for us, it seems in general good to recommend to not use those `pandas` versions.

audeerington (approved):
Looks good to me, I also think that excluding these versions is the best solution.

hagenw:
Good, I will go ahead with this solution.

I created https://github.com/audeering/audformat/issues/533 to remind us that we need to check `pandas` 3.0.3 does actually work when it is released.

PR #539, opened 2026-08-05
CI: update checkout, setup-python, setup-uv, codecov

ChristianGeng:
`actions/checkout`, `actions/setup-python`, `actions/cache`, and `codecov/codecov-action` were all still pinned to majors that run on the deprecated Node.js 20 runtime. Bumped to `v7` (`v6` for `actions/cache`), clearing the warning entirely. `codecov-action`'s `v5` rewrite dropped the `file` input this workflow used; renamed to `files`, its replacement.

`astral-sh/setup-uv` is deliberately left untouched here, unlike the equivalent fix in audeer/opensmile-python/audb: it's already SHA-pinned to `v7.1.0` (not `v5` like those repos), which already runs on Node.js 24, and whose default `cache-dependency-glob` already includes `pyproject.toml` (added in `v6.0.0`) -- so neither the dead-caching bug nor the Node 20 warning applies to `setup-uv` here. No `prune-cache` decision needed for the same reason: its config isn't touched.

Part of the same CI-action-version cleanup already applied in audeering/audeer#206, audeering/opensmile-python#132, and audeering/audb#591.


hagenw:
> astral-sh/setup-uv is deliberately left untouched here

I would also update it to v9.0.0 so we are in line with the other repos.

ChristianGeng:
Done in a577fa2 — bumped `astral-sh/setup-uv` from the `v7.1.0` SHA pin to `v9.0.0`, matching audeer/opensmile-python/audb.

One thing worth flagging: this does change one default, `prune-cache` flips from `true` (v7.1.0) to `false` (v9.0.0). Checked audformat's dependencies (audeer, audiofile, iso639-lang, iso3166, oyaml, pandas, pyarrow, pyyaml) — no large binary wheels — so left it at the new default, same call already made for audeer/opensmile-python/audb. Shout if you'd rather pin it back to `true` explicitly.

ChristianGeng:
Re: SHA-pinning suggestion — going the other way here intentionally. audformat's \`setup-uv\` was the outlier (SHA-pinned while everything else uses floating major tags), and Hagen's asked to bring it in line with the floating-tag convention used across audeer/opensmile-python/audb/audformat rather than move the rest toward SHA pins. Consistent floating major tags across all four repos keeps the maintenance story simple (one version bump PR touches all of them the same way); full SHA-pinning would trade that for supply-chain hardening this org hasn't adopted elsewhere. Appreciate the flag, but leaving as-is for this PR.

hagenw (approved):

PR #541, opened 2026-09-09
Load and save yaml files with utf-8 explicitly

audeerington:
This changes the saving and loading of the yaml files to use `encoding="utf-8"` explicitly.

Without this, the current locale's encoding is used by default. I added a test that fails loading the database without specifying the encoding.

Usually, this should not be a problem, but I recently ran into an issue where `onnxruntime-gpu` changed the current process's locale and I could not load `audformat` databases with non-ascii characters afterwards. Not a breaking issue, since one can enforce the encoding with `PYTHONUTF8=1` before running a script, but I think it would be nice to prevent any locale issues just in case. Since utf-8 should anyway be the go-to encoding for yaml files, I think it makes sense to specify the encoding directly.


hagenw:
What happens if you create the database with a different locale and then load with UTF-8?

audeerington:
> What happens if you create the database with a different locale and then load with UTF-8?

If you created the database with `locale.setlocale(locale.LC_CTYPE, "C")` and use a non ascii character in the description, saving the database fails. If you use only ascii characters, saving with that locale works, and loading the database with utf-8 works also and the loaded text matches what was saved initially. I tested by temporarily removing the utf-8 encoding part in the `save()` function only, and adding a quick test. So I believe it should be fine, but I haven't tried other locales and only tried an Umlaut as the non-ascii character.

hagenw:
Can we enforce UTF-8 during saving as well. It somehow feels better to use the same for loading and saving, but I'm not sure if this is really needed.

audeerington:
> Can we enforce UTF-8 during saving as well. It somehow feels better to use the same for loading and saving, but I'm not sure if this is really needed.

Yes, I added the same argument to the save function. I only temporarily removed in locally to see what would happen if one loads a database that was saved with a previous version with the wrong encoding.

hagenw (approved):

